PR #178, opened 2014-01-16
requireSemicolon implementation

lahmatiy:
Boolean rule that requires semicolon after:
- var declaration
- expression statement
- return
- throw
- break
- continue
- do-while

Valid:

``` js
var a = 1;
var fn = function(){ return; };
if (a > 0)
  b = a * 5;
```

Invalid:

``` js
var a = 1
var fn = function(){ return };
if (a > 0)
  b = a * 5
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,18 +93,12 @@ Values: `true`
 
 ```js
 var a = 1;
-var fn = function(){ return; };
-if (a > 0)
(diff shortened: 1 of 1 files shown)


lahmatiy:
Sorry for delay.
I've made a new PR #951

PR #579, opened 2014-08-20
Update requireMultipleVarDecl rule to support one var per declaration

kilemensi:
`requireMultipleVarDecl` only supported one `var` per function scope and
consecutive `var` checks. It was not possible to check that each declaration
statement had it's own `var` terminated by a semicolon such as in the Felix's Node.js Style Guide. [Example.](Seravo/js-winning-style/blob/master/README.md#variable-declarations)

This means the original `requireMultipleVarDecl` option of boolean true for
consectutive `var` check is now changed to a string type aptly named
`"nonconsecutive"`, and a new option of `"onedecl"` is introduced for the new
check.

With
```js
"requireMultipleVarDecl": "onedecl"
```
the variable declaration style of
```js
var x = 1;
var y = 2; var z = 3;
```
will pass the style checks while that of
```js
var x = 1,
    y = 2;
```
will fail.


kristianroebuck:
Any news on this ticket? I'd be hugely grateful for the `"onedecl"` option.

markelog:
Wow, that is an old one, @kilemensi are you still interested?

kilemensi:
Hi @markelog

I'm still using this in my project since the functionality provided by this pullrequest is stil missing even in the latest `jscs` for me, but, code-wise, I haven't kept up with recent `jscs` changes and may be a new pull request might be best at this point.

markelog:
Well here, i see only test changes, new commits in jscs should not affect new features too much

kilemensi:
I've just had a quick look but I think changes made by @hzoo in #749, especially the `strict` optional kind of supersedes this PR.

markelog:
Well, currently, when you changing only tests it is not, but if you do provide a pull for this, merge should not be difficult

kilemensi:
@markelog 

Thanks for the patience. I didn't notice there were some missing file(s) (I think I was stupid enough to create the PR on `master` instead of it's own branch).

I've recreated the PR (Using @coveralls [a7bd0ac](https://github.com/jscs-dev/node-jscs/commit/a7bd0ac63a25e0f0a8a943b1d8b6b7099e9590c9)) and I think it should be clear what I was trying to do and why #749 pretty much accomplishes the same thing

lib/rules/require-multiple-var-decl.js
```diff
@@ -99,11 +113,11 @@ module.exports = function() {};
 module.exports.prototype = {
     configure: function(options) {
         assert(
-            options === true || options === 'onevar',
-            this.getOptionName() + ' option requires a true value or `onevar`'
+            options === 'onevar' || options === 'nonconsecutive' || options === 'onedecl',
```
markelog: Although this change in most probability will be landed to 2.0, it is better to preserve `true` value.

markelog:
Could you also add docs for this?

markelog:
friendly ping

kilemensi:
@markelog 

The new `strict` option of `disallowMultipleVarDecl (new in a sense it was added later than this PR was first opened) does the same thing that this PR was trying to achieve. Since that one has already been accepted and pushed out, I think we should contrinue with it, unless I'm missing something.

I've updated my project's `.jscsrc` to use the `"disallowMultipleVarDecl": "strict"` and so far so good.

markelog:
Okay, great, thank you!

PR #604, opened 2014-08-28
#533: Update requireAlignedObjectValues rule to work with whitespace aligned via spaces and/or tabs

### LINKED ISSUE #533: requireAlignedObjectValues does not work with tabs
Hello,

requireAlignedObjectValues rule does not work with tabs but only with spaces. Would it be possible to work with spaces too?

Example:
```javascript
q = {
	qwe    : 1,
	asdfgh : 2
};
```
Output:
```
$ jscs test.js 
No code style errors found.
```

---

And if tabs are used (eq to 4 spaces)
```javascript
q = {
	qwe		: 1,
	asdfgh	: 2
};
```
```
$ jscs test.js 
Alignment required at test.js :
     1 |q = {
     2 | qwe  : 1,
--------------^
     3 | asdfgh : 2
     4 |};
```

Also you can notice that there is no error on line 3 because there is only a single whitespace character and jscs thinks it is aligned fine.

---

Also this seems to be fine too using lots of tabs. I guess this is because there are the same number of tabs as spaces would be.
```javascript
q = {
	qwe				: 1,
	asdfgh	: 2
};
```

Thanks!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3187440-requirealignedobjectvalues-does-not-work-with-tabs?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SgtPooki:
This fixes #533. 

I verified that this fix works in the following scenarios:

- Only tabs used for alignment
- Only spaces used for alignment
- tabs used for some object keys and spaces used for others
- tabs and spaces used for some object keys, spaces used for some, and tabs used for others.

It should work in all scenarios and give you an accurate reading as to whether your objects are aligned or not as long as you update the requireAlignedObjectValues rule to be set to something like the following:
```
{
  "requireAlignedObjectValues" : {
    "mode" : "any string that was valid before",
    "tabSize" : "the amount of whitespace that your tabs count as"
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1454,12 +1454,16 @@ var x = {
 
 Requires proper alignment in object literals.
 
-Type: `String`
+Type: `String` or `Object`
 
 Values:
- - `"all"` for strict mode,
- - `"ignoreFunction"` ignores objects if one of the property values is a function expression,
- - `"ignoreLineBreak"` ignores objects if there are line breaks between properties
+ - `String`:
+    - `"all"` for strict mode,
+    - `"ignoreFunction"` ignores objects if one of the property values is a function expression,
+    - `"ignoreLineBreak"` ignores objects if there are line breaks between properties
+ - `Object`:
+    - `mode`: (required) specify one of the string options above
+    - `tabSize`: (default: `1`) considered the tab character as number of specified spaces
 
 #### Example
 
--- a/README.md
+++ b/README.md
@@ -2341,7 +2341,7 @@ Values:
  - `Integer`: lines should be at most the number of characters specified
  - `Object`:
     - `value`: (required) lines should be at most the number of characters specified
-    - `tabSize`: (default: `1`) considered the tab character as number of specified spaces
+    - `tabSize`: (default: `1`) considers the tab character as the number of specified spaces
     - `allowComments`: (default: `false`) allows comments to break the rule
     - `allowUrlComments`: (default: `false`) allows comments with long urls to break the rule
     - `allowRegex`: (default: `false`) allows regular expression literals to break the rule
--- a/lib/rules/require-aligned-object-values.js
+++ b/lib/rules/require-aligned-object-values.js
@@ -1,3 +1,7 @@
+/* globals require, module */
+/* jshint -W097 */
+'use strict';
+
 var assert = require('assert');
 var tokenHelper = require('../token-helper');
 
@@ -39,7 +43,8 @@ module.exports.prototype = {
 
         assert(
             typeof mode === 'string' && modes[mode],
-            this.getOptionName() + (isProperty ? '.mode' : '') + ' requires one of the following values: ' + Object.keys(modes).join(', ')
+            this.getOptionName() + (isProperty ? '.mode' : '') +
+                ' requires one of the following values: ' + Object.keys(modes).join(', ')
         );
 
         this._mode = modes[mode];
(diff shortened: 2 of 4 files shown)


mikesherov:
@markelog reopened it so that when I abstract tabSize out of a specific ruleset and into config like dialect, I know which rules to modify, including this.

markelog:
@mikesherov still relevant?

lib/rules/require-aligned-object-values.js
```diff
@@ -12,10 +40,13 @@ module.exports.prototype = {
             'skipWithFunction': 'ignoreFunction',
             'skipWithLineBreak': 'ignoreLineBreak'
         };
+
```
zxqfox: @SgtPooki As a man who merging like thouthands line of code each 1-2 weeks I can answer you: it's just stupidly annoying when someone adding empty lines and another one removing them. Personally in my team we have non-verbal rules: 1. Don't add empty lines in places in the code unless you specially fixing code style; 2. Use separate issues and commits for code-style fixes; 3. Changes should be focused on issues.

I'm totally agree with @mikesherov that there are no reasons to toggle these empty lines in this issue.

mikesherov:
@markelog you can reclose.

lib/rules/require-aligned-object-values.js
SgtPooki: @zxqfox I totally agree with separating style commits from code changes. I did not realize that is what he was talking about, or that I had even done it. That is why I asked for clarification.

zxqfox: I'm checking what empty lines I've added with `git show`, `git show HEAD~2` or `git diff origin/master` (after rebasing of origin/master) and just removing random spaces I've accidentally added: to make commits focused on issues.

README.md
```diff
@@ -1454,12 +1454,16 @@ var x = {
 
 Requires proper alignment in object literals.
 
-Type: `String`
+Type: `String` or `Object`
```
zxqfox: TypeS.

README.md
```diff
@@ -1454,12 +1454,16 @@ var x = {
...
 
 Values:
- - `"all"` for strict mode,
- - `"ignoreFunction"` ignores objects if one of the property values is a function expression,
- - `"ignoreLineBreak"` ignores objects if there are line breaks between properties
+ - `String`:
```
zxqfox: Not sure about this change.

lib/rules/require-aligned-object-values.js
```diff
@@ -1,10 +1,38 @@
...
+            this.configureMode(config.mode, true);
+            this.configureTabSize(config.tabSize);
+
+        } else {
+            this.configureMode(config);
+        }
+    },
+
+    configureTabSize: function(tabSize) {
```
zxqfox: :+1: for moving them as local functions into `configure`

lib/rules/require-aligned-object-values.js
```diff
@@ -23,17 +50,33 @@ module.exports.prototype = {
...
+
+        return {
+            pos: colonPosition,
+            colon: colon
+        };
+    },
+
     check: function(file, errors) {
-        var tokens = file.getTokens();
+
```
zxqfox: Intentional empty line.

lib/rules/require-aligned-object-values.js
```diff
@@ -23,17 +54,33 @@ module.exports.prototype = {
         return 'requireAlignedObjectValues';
     },
 
+    _getColonPosition: function(file, property) {
```
zxqfox: :+1: You using it in `check` and no reason to left it in prototype.

Issue #92, opened 2013-12-09
Rules for spaces inside various parens


jzaefferer:
In the [jQuery JavaScript style guide](http://contribute.jquery.org/style-guide/js/), whitespace is added inside various parens, round and square. A few samples:

```js
array = [ a, b ];
foo( options, object[ property ] );
```

For function calls, there are exceptions to this rule:

```js
// Function with a callback, object, or array as the sole argument:
// No space on either side of the argument
foo({
    a: "alpha",
    b: "beta"
});
 
// Function with a callback, object, or array as the first argument:
// No space before the first argument
foo(function() {
    // Do stuff
}, options );
 
// Function with a callback, object, or array as the last argument:
// No space after after the last argument
foo( data, function() {
    // Do stuff
});
```

I have no idea how much effort it is to write rules for this. Before I look into that, I'd like to hear from someone if this is reasonable and would be merged. Maybe some hints on how to implement it?

mikesherov:
The saga of this issue has now been resolved by simplifying the jQuery style guide. I'm going to close this issue and file another one just for adding existing rules to the jQuery style guide.

PR #679, opened 2014-10-12
New rule: requireSpacesAroundFunctionArguments

markelog:
Another portion for #92

Opposite rule will follow.

Still have not idea how to catch grouping case :-(

### DIFF
--- a/README.md
+++ b/README.md
@@ -601,7 +601,95 @@ var a = b?c :d;
 var a = b?c: d;
 ```
 
+### requireSpacesAroundFunctionArguments
 
+Requires spaces before first and after last function arguments
+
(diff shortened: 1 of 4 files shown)


mikesherov:
Looks like this rule can be drastically simplified now that https://github.com/jquery/contribute.jquery.org/pull/104 has landed. ping @markelog

markelog:
> Looks like this rule can be drastically simplified 

Or not accepted, it seems `requireSpacesInsideParentheses` is now suitable for our needs

mikesherov:
Well, sorry for making you do this work... Let's close this then. Reopen if you disagree!

PR #781, opened 2014-11-20
Presets: Remove use of outdated validateJSDoc rule

### LINKED ISSUE #778: Invalid JSDoc param argument
```
~/D/node-precinct (sass ⚡) jscs -p google index.js
Invalid JSDoc @param argument name at index.js :
     8 |/**
     9 | * Finds the list of dependencies for the given file
    10 | * @param {String} file - path of the file whose dependencies to find
-----------^
    11 | * @param {String} [type] - The type of content being passed in. Useful if you want to use a non-js detective
    12 | * @return {String[]}


1 code style error found.
```

Will investigate the root cause of this. Might only be the `google` preset.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/6204331-invalid-jsdoc-param-argument?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mrjoelkemp:
The rule has a bug that prevents ordinary use of docblocks claiming that
a param type is incorrect. It's better to use the actual jscs-jsdoc plugin.

Fixes #778

### DIFF
--- a/presets/google.json
+++ b/presets/google.json
@@ -57,12 +57,6 @@
     "disallowSpacesInsideArrayBrackets": "all",
     "disallowSpacesInsideParentheses": true,
 
-
-    "validateJSDoc": {
-        "checkParamNames": true,
-        "requireParamTypes": true
-    },
-
     "disallowMultipleLineBreaks": true,
     "disallowNewlineBeforeBlockStatements": true
 }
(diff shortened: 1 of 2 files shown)


zxqfox:
@mrjoelkemp Unfortunately they won't fail. No rule — no reports.

Do we really need to remove jsdoc blocks from integration tests?

PR #822, opened 2014-12-07
requireMultipleVarDecl: change to true for airbnb preset

mbohal:
The airbnb js coding standard states, that each variable should be initialized with its own ```var``` keyword. (https://github.com/airbnb/javascript#variables).

In  the presets, however, the rule ```requireMultipleVarDecl``` is set to ```onevar```. I think it should be set to ```true```


markelog:
It was decided to wait to apply this policy until we switch to more frequent releases, bottom line is - updated preset will be available in 1.11.

PR #847, opened 2014-12-19
requireCapitalizedComments: corrected letter recognition

### LINKED ISSUE #842: requireCapitalizedComments isn't recognizing letters correctly
While I had a look at [issue 837](https://github.com/jscs-dev/node-jscs/issues/837), I encountered some problems with the rule. Sorry, this report got a little bit longer...

The rule has currently two problems:
1. It only recognizes the letters A-Z as uppercase letters.
2. It only recognizes the letters a-z and A-Z as letters.

This creates a problem with letters like Ä, ä, ∏, π,  ...
### Example for 1.

The following comment should throw an error (the "ü" is lowercase), but isn't throwing one:

```
// über
```

The rule thinks "ü" is no letter, so there is nothing to check. This is not what we want.
### Example for 2.

The following comment shouldn't throw an error (the "Ü" is an uppercase letter and is thus starting a text block, so "diese" can be lowercase), but is throwing one:

```
// Über
// diese Funktion
```

jscs wants it to be

```
// Über
// Diese Funktion
```

For the rule "Über" is not starting a text block, but "diese" is so it should be uppercase. Again, this is not what we want.
## Solution time

The only feasible solution to recognize letter (I know of) is generating a regular expression, which checks the characters.

This should probably be done with Mathias' [regenerate](https://github.com/mathiasbynens/regenerate), a bit like jshint [is doing it](https://github.com/jshint/jshint/blob/master/scripts/generate-identifier-data.js).

AlexanderZeilmann:
fixes #842

### DIFF
--- a/lib/rules/disallow-capitalized-comments.js
+++ b/lib/rules/disallow-capitalized-comments.js
@@ -9,7 +9,6 @@ var Lo = require('unicode-6.3.0/categories/Lo/code-points');
 var letterPattern = regenerate(Lu, Ll, Lt, Lm, Lo).toRegExp();
 var lowerCasePattern = regenerate(Ll).toRegExp();
 
-
 module.exports = function() {};
 
 module.exports.prototype = {
--- a/test/rules/disallow-capitalized-comments.js
+++ b/test/rules/disallow-capitalized-comments.js
@@ -18,7 +18,7 @@ describe('rules/disallow-capitalized-comments', function() {
         assert(checker.checkString('/* Invalid */').getErrorCount() === 1);
         assert(checker.checkString('/*\n Invalid\n */').getErrorCount() === 1);
         assert(checker.checkString('//Über').getErrorCount() === 1);
-        assert(checker.checkString('//∏').getErrorCount() === 1);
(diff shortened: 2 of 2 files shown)


lib/rules/require-capitalized-comments.js
```diff
@@ -1,7 +1,13 @@
...
+var regenerate = require('regenerate');
+var Lu = require('unicode-6.3.0/categories/Lu/code-points');
+var Ll = require('unicode-6.3.0/categories/Ll/code-points');
+var Lt = require('unicode-6.3.0/categories/Lt/code-points');
+var Lm = require('unicode-6.3.0/categories/Lm/code-points');
+var Lo = require('unicode-6.3.0/categories/Lo/code-points');
 
-var upperCasePattern = /[A-Z]/;
-var alphabeticalPattern = /[a-z|A-Z]/;
+var letterPattern = regenerate(Lu, Ll, Lt, Lm, Lo).toRegExp();
```
mathiasbynens: Instead of including `Lu`, `Ll`, `Lt`, `Lm`, and `Lo` separately, why not just use [`L`](http://unicode.org/reports/tr44/#GC_Values_Table)?

PR #850, opened 2014-12-22
requireMatchingFunctionName: requires function names to match member and property names

### LINKED ISSUE #846: Require function names to match member and property names
From time to I see the following "typos":

`MyClass.prototype.myFunction = function anotherFunction() { ... }`

or 

```
var myObject = {
  myFunction: function anotherFunction() { ... }
};
```

So it would be nice to be able to catch such cases and throw an error. I can come up with a PR if it makes sense, but we need to decide on a rule name like `enforceFunctionNameMatch`.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7143435-require-function-names-to-match-member-and-property-names?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xaka:
Example of member name mismatch:

```js
var test = {};
test.foo = function bar() {};
```

Example of property name mismatch:

```js
var test = {foo: function bar() {}};
```

Fixes #846
Closes #850

### DIFF
--- a/lib/rules/require-matching-function-name.js
+++ b/lib/rules/require-matching-function-name.js
@@ -1,0 +1,131 @@
+/**
+ * Requires function names to match member and property names.
+ *
+ * It doesn't affect anonymous functions nor functions assigned to members or
+ * properties named with a reserved word.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireMatchingFunctionName": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var test = {};
+ * test.foo = function foo() {};
+ * ```
+ *
+ * ```js
+ * var test = {};
+ * test['foo'] = function foo() {};
+ * ```
+ *
+ * ```js
+ * var test = {foo: function foo() {}};
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var test = {};
+ * test.foo = function bar() {};
+ * ```
+ *
+ * ```js
+ * var test = {};
+ * test['foo'] = function bar() {};
+ * ```
+ *
+ * ```js
+ * var test = {foo: function bar() {}};
+ * ```
+ */
+
+var assert = require('assert');
+var utils = require('../../lib/utils');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(requireMatchingFunctionName) {
+        assert(
+            requireMatchingFunctionName === true,
+            'requireMatchingFunctionName option requires true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireMatchingFunctionName';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType(['FunctionExpression'], function(node) {
+            switch (node.parentNode.type) {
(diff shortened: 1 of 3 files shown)


lib/rules/require-matching-function-name.js
```diff
@@ -0,0 +1,87 @@
...
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType(['FunctionExpression'], function(node) {
+            var checker;
+            switch (node.parentNode.type) {
+                // object.foo = function bar() {}
+                // object['foo'] = function bar() {}
+                case 'AssignmentExpression':
+                    checker = checkForMember;
```
mikesherov: just call `checkForMember(node.parentNode, errors);` here;

lib/rules/require-matching-function-name.js
```diff
@@ -0,0 +1,87 @@
...
+                    break;
+
+                // object = {foo: function bar() {}}
+                case 'Property':
+                    checker = checkForProperty;
```
mikesherov: just call `checkForProperty(node.parentNode, errors);` here.

lib/rules/require-matching-function-name.js
```diff
@@ -0,0 +1,87 @@
...
+                    break;
+            }
+
+            if (checker) {
```
mikesherov: ditch this variable.

lib/rules/require-matching-function-name.js
```diff
@@ -0,0 +1,87 @@
...
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireMatchingFunctionName';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType(['FunctionExpression'], function(node) {
+            var checker;
```
mikesherov: ditch the variable.

test/rules/require-matching-function-name.js
```diff
@@ -0,0 +1,72 @@
...
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    describe('option value true', function() {
+        beforeEach(function() {
+            checker.configure({ requireMatchingFunctionName: true });
+        });
+
+        // TODO:
```
mikesherov: are these todos done?

mikesherov:
@xaka just a couple of more comments and a rebase should bring this one across the finish line. Thanks again for contributing!

xaka:
@mikesherov Done. Please double-check the documentation part as you changed the way it's generated and I did my best to follow the same path.

test/rules/require-matching-function-name.js
xaka: Nope, these todos are more like a roadmap for this particular rule. I can remove it and we'll let the users to decide what should be done next by opening issues on GitHub.

markelog:
@mikesherov are we good here?

test/rules/require-matching-function-name.js
zxqfox: Voting for an issue if we need it ;-)

markelog:
Thank you for the patience! And sorry it took that long to land.

PR #858, opened 2014-12-24
Add edge cases for require-line-break-after-variable-assignment

jdlrobson:
* use end position rather than start of last declaration

Closes: 540


markelog:
@Krinkle could you verify?

Issue #866, opened 2014-12-30
'requireDotNotation' vs 'requireCamelCaseOrUpperCaseIdentifiers'


necolas:
The docs for `requireDotNotation` describe it as requiring dot notation "when possible". When you are using an external library with method names that violate a local `requireCamelCaseOrUpperCaseIdentifiers` setting, it might be nice to be able to use `lib['snake_method']` and not trigger a `requireDotNotation` error, rather than having to disable style checks.

```js
var lib = require('lib');

// triggers requireCamelCaseOrUpperCaseIdentifiers error
lib.snake_method()
// triggers requireDotNotation error
lib['snake_method']

// only way out is to disable style checking in the file or on the relevant lines
```

PR #868, opened 2014-12-31
requireDotNotation: new value except_snake_case

zxqfox:
Ref #866

### DIFF
--- a/lib/rules/require-dot-notation.js
+++ b/lib/rules/require-dot-notation.js
@@ -1,9 +1,11 @@
 /**
  * Requires member expressions to use dot notation when possible
  *
- * Type: `Boolean`
+ * Type: `Boolean` or `String`
  *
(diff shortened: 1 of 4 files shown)


zxqfox:
:up:

markelog:
Could add tests for https://github.com/jscs-dev/node-jscs/pull/868#discussion_r22388356?

zxqfox:
@markelog Rebased, reverted accidentally dropped docs for es3 flag (website commit), added test for case `___`.

markelog:
Merged, although, i doubt we would ever need `isSnakeCased` and `trimUnderscores` methods somewhere else.

Let's put a pin on those

zxqfox:
@markelog e.g. we don't have a semantic opposite to `requireCamelCaseOrTitleCase` rule. ;-)

PR #871, opened 2015-01-01
Growing coverage for some rules

zxqfox:
I've dropped some checks because I didn't find cases for them.

Strongly need review.

**edit** Locally I see 93% ;-) Here is 95%.

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -151,6 +151,42 @@ JsFile.prototype = {
         this._tokenRangeEndIndex = tokenRangeEndIndex;
     },
     /**
+     * Builds comments index by starting pos for futher navigation.
+     */
+    _buildCommentIndex: function() {
+        var comments = this.getComments();
+        var tokens = this.getTokens();
+        var tokenIndex = 0;
+        var tokensLength = tokens.length;
+        var tokenCommentsBeforeIndex = this._tokenCommentsBeforeIndex = {};
+        var tokenCommentsAfterIndex = this._tokenCommentsAfterIndex = {};
+        var partialComments = [];
+        for (var i = 0, l = comments.length; i < l && comments[i].range[1] <= tokens[0].range[0]; i++) {
+            partialComments.push(comments[i]);
+        }
+        if (partialComments.length) {
+            tokenCommentsBeforeIndex[0] = partialComments;
+            partialComments = [];
+        }
+        for (; i < l; i++) {
+            var startPos = comments[i].range[0];
+            var endPos = comments[i].range[1];
+            if (partialComments.length) {
+                tokenCommentsAfterIndex[tokenIndex] = partialComments;
+                tokenCommentsBeforeIndex[tokenIndex + 1] = partialComments;
+                partialComments = [];
+            }
+            while (tokenIndex + 1 < tokensLength && tokens[tokenIndex + 1].range[1] <= startPos) {
+                tokenIndex++;
+            }
+            partialComments.push(comments[i]);
+        }
+        if (partialComments.length) {
+            tokenCommentsAfterIndex[tokenIndex] = partialComments;
+            tokenCommentsBeforeIndex[tokenIndex + 1] = partialComments;
+        }
+    },
+    /**
      * Returns token position using range start from the index.
      *
      * @returns {Object}
@@ -366,6 +402,32 @@ JsFile.prototype = {
                 cb(token, index, tokens);
             }
         });
+    },
+    /**
+     * Returns comment node directly followed by a token
+     *
+     * @param {Object} token
+     * @returns {Object|undefined}
+     */
+    getCommentAfterToken: function(token) {
+        if (!this._tokenCommentsAfterIndex) {
+            this._buildCommentIndex();
+        }
+        var comments = this._tokenCommentsAfterIndex[token._tokenIndex];
+        return comments ? comments[0] : undefined;
+    },
+    /**
+     * Returns comment node directly followed by a token
+     *
+     * @param {Object} token
+     * @returns {Object|undefined}
+     */
+    getCommentBeforeToken: function(token) {
+        if (!this._tokenCommentsBeforeIndex) {
+            this._buildCommentIndex();
+        }
+        var comments = this._tokenCommentsBeforeIndex[token._tokenIndex];
+        return comments ? comments[comments.length - 1] : undefined;
     },
     /**
      * Returns which dialect of JS this file supports.
--- a/lib/rules/disallow-space-after-keywords.js
+++ b/lib/rules/disallow-space-after-keywords.js
@@ -27,6 +27,14 @@
  *     y++;
  * }
  * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * if (x > y) {
+ *     y++;
+ * }
+ * ```
  */
 
 var assert = require('assert');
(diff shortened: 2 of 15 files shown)


lib/rules/require-curly-braces.js
```diff
@@ -72,7 +72,7 @@ module.exports.prototype = {
         if (typeIndex['case'] || typeIndex['default']) {
             file.iterateNodesByType('SwitchCase', function(node) {
                 // empty case statement
-                if (!node.consequent || node.consequent.length === 0) {
+                if (node.consequent.length === 0) {
```
mrjoelkemp: In general, my problem with the weakening of the conditions is that they were added for edge cases. If we don't have a test hitting that condition, then we should add the test, not remove the condition. 

With the removal, this would throw an error on empty case statements since `.length` wouldn't exist on `undefined`.

mrjoelkemp:
Actually, those files are already 100% covered according to coveralls. What are you using to see coverage locally?

lib/rules/require-curly-braces.js
zxqfox: Parser API said: consequent should exists and be an Array in any cases. So no need to check it.

lib/rules/disallow-space-after-keywords.js
```diff
@@ -31,7 +31,7 @@ module.exports.prototype = {
         file.iterateTokensByType('Keyword', function(token, i, tokens) {
             if (keywordIndex[token.value]) {
                 var nextToken = tokens[i + 1];
-                if (nextToken && nextToken.range[0] !== token.range[1]) {
```
zxqfox: Here I can't imagine a situation where nextToken for Keyword wouldn't exist. Probably I'm wrong.

lib/rules/disallow-space-before-keywords.js
```diff
@@ -52,7 +52,7 @@ module.exports.prototype = {
                 var diff = token.range[0] - prevToken.range[1];
 
                 if (prevToken.loc.end.line === token.loc.start.line && diff !== 0) {
-                    if (prevToken.type !== 'Punctuator' || prevToken.value !== ';') {
```
zxqfox: Same as above. Need help with case.

zxqfox:
> Actually, those files are already 100% covered according to coveralls.

I've just ran `npm run coverage-html` ;-)

These conditions marked yellow, and I think this information doesn't placed in `lcov` and these statements marked as covered. But actually it's overkill.

mdevils:
Would you please try and use our assertion methods?

* Assertions: https://github.com/jscs-dev/node-jscs/blob/master/lib/token-assert.js
* Example: https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-space-before-binary-operators.js

zxqfox:
@mdevils Yeah, why not.

What if there is no `prevToken`? It will be a uncaught error at [lib/token-assert.js#L124](https://github.com/jscs-dev/node-jscs/blob/master/lib/token-assert.js#L124)

mdevils:
Feel free to fix that check.

zxqfox:
@mdevils Okey. I had tried to find test cases before. Can you please provide some ideas how to test these cases.

mdevils:
In this case — no idea.

If you want to test `tokenBefore` with no `prevToken`, you can pass the first document token.

zxqfox:
So I'm not sure but feels like we just don't need that check. ;-\ Okay. I'll rewrite these rules with `token-assert`. Thanks!

mdevils:
That check is needed in `tokenBefore`, not outside of it. So its easy to add and write a test.

lib/rules/require-curly-braces.js
mrjoelkemp: Right. My bad, I was thinking about `SwitchStatement`, not `SwitchCase`. This seems okay then.

zxqfox:
@mdevils There is a problem with comments between tokens. Should I inject comments search logic into `.whitespaceBetween` and `.noWhitespaceBetween` functions?

mdevils:
Which problem did you face?

zxqfox:
@mdevils `/* comment */ if /* yolo */ ( /* another */ condition /* more */) /* again */ { // etc`

Should `whitespaceBetween` ignore them? Should it count spaces outside of comments between tokens?

zxqfox:
@mdevils I suppose to use a helpers `getCommentAfter/BeforeToken` to get the closest comment for token. It simplifies a lot inner rule code for a bunch of rules. https://github.com/zxqfox/node-jscs/commit/7a86811c61e29ec8b47cd309a3c6a6bb349441b9

lib/js-file.js
```diff
@@ -151,6 +151,42 @@ JsFile.prototype = {
         this._tokenRangeEndIndex = tokenRangeEndIndex;
     },
     /**
+     * Builds comments index by starting pos for futher navigation.
+     */
+    _buildCommentIndex: function() {
+        var comments = this.getComments();
+        var tokens = this.getTokens();
+        var tokenIterator = 0;
```
mdevils: `tokenIterator` is a strange name for this. Let's go for `tokenIndex`.

mdevils:
LGTM.

@markelog, please take a look.

lib/js-file.js
```diff
@@ -368,6 +404,32 @@ JsFile.prototype = {
         });
     },
     /**
+     * Returns comment node directly followed by a token
+     *
+     * @param {Object} token
+     * @returns {Object|undefined}
+     */
+    getCommentAfterToken: function(token) {
+        if (!this._tokenCommentsAfterIndex) {
```
zxqfox: Is it fine if we checking it in methods? We can call here `_buildCommentIndex` and move check there. Also we can build that index every time just after token index building.

markelog:
Still a WIP?

test/rules/disallow-space-before-keywords.js
```diff
@@ -45,6 +45,18 @@ describe('rules/dissalow-space-before-keywords', function() {
         assert(checker.checkString('if (true) {\n} /**/else { x++; }').isEmpty());
     });
 
+    it('should not report if tokens placed on different lines', function() {
+        // checking !(prevToken.loc.end.line === token.loc.start.line)
```
mikesherov: Remove this comment.

test/rules/disallow-space-before-keywords.js
```diff
@@ -45,6 +45,18 @@ describe('rules/dissalow-space-before-keywords', function() {
...
+        checker.configure({ disallowSpaceBeforeKeywords: true });
+        assert(checker.checkString('x\nif (true) false').isEmpty());
+    });
+
+    it('should not report if there are no braces', function() {
+        // checking !(prevToken.type !== 'Punctuator' || prevToken.value !== ';')
```
mikesherov: Remove comment.

test/rules/require-space-before-keywords.js
```diff
@@ -64,6 +64,18 @@ describe('rules/require-space-before-keywords', function() {
         assert(errors.explainError(error).indexOf('Should be one space instead of 2, before "else"') === 0);
     });
 
+    it('should not trigger error for different lines', function() {
+        // !(prevToken.loc.end.line === token.loc.start.line)
```
mikesherov: Remove

test/rules/require-space-before-keywords.js
```diff
@@ -64,6 +64,18 @@ describe('rules/require-space-before-keywords', function() {
...
+        checker.configure({ requireSpaceBeforeKeywords: ['else'] });
+        assert(checker.checkString('if (true) {\n} /**/ \nelse { x++; }').isEmpty());
+    });
+
+    it('should not report if there are no braces', function() {
+        // checking !(prevToken.type !== 'Punctuator' || prevToken.value !== ';')
```
mikesherov: Remove.

zxqfox:
Fixed last nuts. Thanks for review, guys!

/cc @markelog

markelog:
Cudos for this hard work!

PR #874, opened 2015-01-01
Preset: add 'catch' to "disallowKeywordsOnNewLine" rule for Wikimedia

jdforrester:
We missed this one. :-)


markelog:
Will merge this, when 1.10 will be around the corner, there is no date for it yet

PR #876, opened 2015-01-03
New rule: requireQuotedKeysInObjects

### LINKED ISSUE #360: Support requireQuotedKeysInObjects
Can we get the opposite of `disallowQuotedKeysInObjects`?

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/2004003-support-requirequotedkeysinobjects?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hpshelton:
If set to 'true', requires all object keys to be surrounded in
quotes. This is the opposite of disallowQuotedKeysInObjects.

Fixes #360 

Notes:
- This is my first PR to jscs, so please be liberal with the feedback. 
- I assume with the new website code, the documentation will get automatically picked up from the source file?

### DIFF
--- a/lib/rules/require-quoted-keys-in-objects.js
+++ b/lib/rules/require-quoted-keys-in-objects.js
@@ -1,0 +1,63 @@
+/**
+ * Requires quoted keys in objects.
+ *
+ * Type: `Boolean`
+ *
+ * Values:
+ *
+ *  - `true` for strict mode
+ *
+ * #### Example
+ *
+ * ```js
+ * "requiresQuotedKeysInObjects": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+  * var x = { 'a': { "default": 1 } };
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var x = { a: 1 };
+ * ```
+ */
+
+var assert = require('assert');
+var utils = require('../utils');
+
+module.exports = function() { };
+
+module.exports.prototype = {
+
+    configure: function(requireQuotedKeysInObjects) {
+        assert(
+            typeof requireQuotedKeysInObjects === 'boolean',
+            this.getOptionName() + ' option requires boolean value'
+        );
+        assert(
+            requireQuotedKeysInObjects === true,
+            this.getOptionName() + ' option requires true value or should be removed'
+        );
+
+        this._mode = requireQuotedKeysInObjects;
+    },
+
+    getOptionName: function() {
+        return 'requireQuotedKeysInObjects';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('ObjectExpression', function(node) {
+            node.properties.forEach(function(prop) {
+                var key = prop.key;
+                if (!(typeof key.value === 'string' && key.type === 'Literal')) {
+                    errors.add('Object key without surrounding quotes', prop.loc.start);
+                }
+            });
+        });
+    }
+};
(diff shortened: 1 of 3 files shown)


lib/rules/require-quoted-keys-in-objects.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Requires quoted keys in objects.
+ *
+ * Type: `String` or `Boolean`
```
markelog: No "allButReserved" option here, so only "boolean" is allowed

test/rules/require-quoted-keys-in-objects.js
```diff
@@ -0,0 +1,70 @@
...
+        assert(checker.checkString('var x = { "1$": 1 }').isEmpty());
+        assert(checker.checkString('var x = { "010": 1 }').isEmpty());
+        assert(checker.checkString('var x = { "6b2ea258-cf24-44fe-b857-c4e37e105d6f": 1 }').isEmpty());
+    });
+
+    it('should check all keys in object', function() {
+        assert(checker.checkString('var x = { a: 1, "b": 2, c: 3 }').getErrorCount() === 2);
+    });
+
+    it('should not report for reserved words', function() {
```
markelog: This and and next test after that are needless?

markelog:
@mdevils it seems we need more methods for `assertion` module?

hpshelton:
Fixed the issues and squashed the commits.

lib/rules/require-quoted-keys-in-objects.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  - `true` for strict mode
+ *
+ * #### Example
+ *
+ * ```js
+ * "requiresQuotedKeysInObjects": true
+ * ```
+ *
+ * ##### Valid for mode `true`
```
zxqfox: Probably it's overkill. We usually describe a value if rule has many.

lib/rules/require-quoted-keys-in-objects.js
```diff
@@ -0,0 +1,63 @@
...
+var utils = require('../utils');
+
+module.exports = function() { };
+
+module.exports.prototype = {
+
+    configure: function(requireQuotedKeysInObjects) {
+        assert(
+            typeof requireQuotedKeysInObjects === 'boolean',
+            this.getOptionName() + ' option requires boolean value'
```
zxqfox: @markelog Did you decided something about these assertions?

zxqfox:
My first PR was much worse. ;-) Nicely done.

lib/rules/require-quoted-keys-in-objects.js
hpshelton: The examples are overkill? If there's no objection to being verbose, I'd prefer to leave them in.

zxqfox: @hpshelton No no. Not an example. Just a "for mode `true`" substring.

hpshelton: Got it. Changed to just "Valid".

lib/rules/require-quoted-keys-in-objects.js
markelog: Nope, i think we need interface that would accept declaration with acceptable values and do this checks automatically.

But i'm not sure if we have talked about this or not. Waiting for @mdevils to respond.

markelog:
Okay, lack of assertion methods should block this PR

PR #877, opened 2015-01-04
requireSpace(Before|After)BinaryOperators: Add tests for error column

### LINKED ISSUE #756: requireSpaceAfterBinaryOperators: issue with location of error
I think the spacing should change to the end of the token right?

I works for single character operators like `>` and `*` but not for `==` and `===`

https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/require-space-after-binary-operators.js#L85

Maybe it should be `var loc = part.loc.start;` -> `var loc = part.loc.end;`? Although i'm not sure since the `===` has it point to the middle character.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/5859520-requirespaceafterbinaryoperators-issue-with-location-of-error?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hpshelton:
Adds additional unit tests to verify the detected error's column value is correct for  requireSpace(Before|After)BinaryOperators, that is, it points to the beginning or end of the operator as appropriate.

Closes #756

### DIFF
--- a/test/rules/require-space-before-binary-operators.js
+++ b/test/rules/require-space-before-binary-operators.js
@@ -20,30 +20,42 @@ describe('rules/require-space-before-binary-operators', function() {
         var notStickedWithParenthesis = 'var test; (test) ' + operator + ' (2)';
 
         [[operator], true].forEach(function(value) {
+
             it('should report sticky operator for ' + sticked + ' with ' + value + ' option',
                 function() {
-                    checker.configure({ requireSpaceBeforeBinaryOperators: value });
+                    checker.configure({ requireSpaceBeforeBinaryOperators: [operator] });
                     assert(checker.checkString(sticked).getErrorCount() === 1);
                 }
             );
+
             it('should not report sticky operator for ' + notSticked + ' with ' + value + ' option',
                 function() {
-                    checker.configure({ requireSpaceBeforeBinaryOperators: value });
+                    checker.configure({ requireSpaceBeforeBinaryOperators: [operator] });
                     assert(checker.checkString(notSticked).isEmpty());
                 }
             );
+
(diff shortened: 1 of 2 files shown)


markelog:
No code changes only tests, i think i love you :-).

Before <a href="https://github.com/jscs-dev/node-jscs/blob/4c31d9ee8039d87aa6db346b5efe7d5679551d21/lib/token-assert.js">assertion API</a> this job was on <a href="https://github.com/jscs-dev/node-jscs/blob/4c31d9ee8039d87aa6db346b5efe7d5679551d21/lib/token-helper.js#L115-117">`getPointerEntities`</a> method, but now it's used only once and with assertions API we no longer need it.

Would you like to remove this method and adapt `requireOperatorBeforeLineBreak` logic to the assertion API?

test/rules/require-space-before-binary-operators.js
```diff
@@ -20,30 +20,42 @@ describe('rules/require-space-before-binary-operators', function() {
...
+                    checker.configure({ requireSpaceBeforeBinaryOperators: [operator] });
                     assert(checker.checkString(notStickedWithParenthesis).isEmpty());
                 }
             );
+
+            it('should highlight the end of the ' + operator + ' operator', function() {
+                checker.configure({ requireSpaceBeforeBinaryOperators: [operator] });
+                var error = checker.checkString(sticked).getErrorList()[0];
+                assert(error.line === 1);
+                assert(error.column === (14));
```
markelog: `(14)`?

PR #879, opened 2015-01-04
requireOperatorBeforeLineBreak: Use the new assertion framework

hpshelton:
Update requireOperatorBeforeLineBreak to use the new assertion framework and remove the call to getPointerEntities since it's no longer used anywhere.

I'm not sure if you want me to do something to clean up the history here since my history's a little confused. Did you apply the previous changes from my master branch as patches rather than merging them? What's the right way to track the upstream repo since that's not discussed in the Contributing documentation?


hzoo:
If i'm not mistaken, I think you can make a new branch for every new feature/PR (and remove when merged), and track/update your master branch with upstream master using something like `git pull --ff upstream master` since changes are rebased into one history? Maybe should be added to docs?

hpshelton:
Closing this PR to fix the history in a new branch. Thanks @hzoo.

hpshelton:
https://github.com/jscs-dev/node-jscs/pull/882

Issue #726, opened 2014-10-28
disallowCurlyBraces


donaldpipowitch:
It looks like I can't `disallowCurlyBraces`? I want to disallow it so switch/case looks like this:

````
switch(a) {
  case 'hello':
    break;
  default:
    break;
}
````

Instead of this:
````
switch(a) {
    case 'hello': {
      break;
    }
    default: {
      break;
    }
}
````
I've never saw someone use that last style.

markelog:
Discussion moved to https://github.com/jscs-dev/node-jscs/issues/796

danny-andrews:
https://github.com/jscs-dev/node-jscs/issues/726#issue-47036167 Does this rule work for this use-case? I wasn't able to get it to work for switch statements.

PR #880, opened 2015-01-05
New rule: disallowCurlyBraces

hzoo:
#726 

The same as requireCurlyBraces except `addError` on `=== 'BlockStatement'` instead of on `!== 'BlockStatement'`.

### DIFF
--- a/lib/rules/disallow-curly-braces.js
+++ b/lib/rules/disallow-curly-braces.js
@@ -1,0 +1,121 @@
+/**
+ * Disallows curly braces after statements.
+ *
+ * Type: `Array` or `Boolean`
+ *
+ * Values: Array of quoted keywords or `true` to disallow curly braces after the following keywords:
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowCurlyBraces": [
+ *     "if",
+ *     "else",
+ *     "while",
+ *     "for",
+ *     "do",
+ *     "with"
+ * ]
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * if (x) x++;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * if (x) {
+ *     x++;
+ * }
+ * ```
+ */
+
+var assert = require('assert');
+var defaultKeywords = require('../utils').curlyBracedKeywords;
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(statementTypes) {
+        assert(
+            Array.isArray(statementTypes) || statementTypes === true,
+            this.getOptionName() + ' option requires array or true value'
+        );
+
+        if (statementTypes === true) {
+            statementTypes = defaultKeywords;
+        }
+
+        this._typeIndex = {};
+        statementTypes.forEach(function(type) {
+            this._typeIndex[type] = true;
+        }.bind(this));
+    },
+
+    getOptionName: function() {
+        return 'disallowCurlyBraces';
+    },
+
+    check: function(file, errors) {
+
+        function isSingleBlockStatement(node) {
+            return node && node.type === 'BlockStatement' &&
+            node.body.length === 1;
+        }
+
+        function addError(typeString, entity) {
+            errors.add(
+                typeString + ' statement with extra curly braces',
+                entity.loc.start.line,
+                entity.loc.start.column
+            );
+        }
+
+        function checkBody(type, typeString) {
+            file.iterateNodesByType(type, function(node) {
+                if (isSingleBlockStatement(node.body)) {
+                    addError(typeString, node);
+                }
+            });
+        }
+
+        var typeIndex = this._typeIndex;
+
+        if (typeIndex['if'] || typeIndex['else']) {
+            file.iterateNodesByType('IfStatement', function(node) {
+                if (typeIndex['if'] && isSingleBlockStatement(node.consequent)) {
+                    addError('If', node);
+                }
+                if (typeIndex['else'] && isSingleBlockStatement(node.alternate)) {
+                    addError('Else', file.findPrevToken(
+                        file.getTokenByRangeStart(node.alternate.range[0]),
+                        'Keyword',
+                        'else'
+                    ));
+                }
+            });
+        }
+
+        if (typeIndex['while']) {
+            checkBody('WhileStatement', 'While');
+        }
+
+        if (typeIndex['for']) {
+            checkBody('ForStatement', 'For');
+            checkBody('ForInStatement', 'For in');
+        }
+
+        if (typeIndex['do']) {
+            checkBody('DoWhileStatement', 'Do while');
+        }
+
+        if (typeIndex['with']) {
+            checkBody('WithStatement', 'With');
+        }
+    }
+
+};
--- a/test/rules/disallow-curly-braces.js
+++ b/test/rules/disallow-curly-braces.js
@@ -1,0 +1,149 @@
+var Checker = require('../../lib/checker');
+var assert = require('assert');
+
+describe('rules/disallow-curly-braces', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    it('should not report missing `if` braces', function() {
+        checker.configure({ disallowCurlyBraces: ['if'] });
+        assert(checker.checkString('if (x) x++;').isEmpty());
+    });
+
+    it('should report `if` with braces', function() {
+        checker.configure({ disallowCurlyBraces: ['if'] });
+        assert(checker.checkString('if (x) { x++; }').getErrorCount() === 1);
+    });
+
+    it('should not report missing `else` braces', function() {
+        checker.configure({ disallowCurlyBraces: ['else'] });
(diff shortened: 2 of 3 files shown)


test/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,143 @@
...
+        checker.configure({ disallowCurlyBraces: ['default'] });
+        assert(checker.checkString('switch(x){case 1:a=b;default:a=c;}').isEmpty());
+    });
+
+    it('should report `default` with braces', function() {
+        checker.configure({ disallowCurlyBraces: ['default'] });
+        assert(checker.checkString('switch(x){default:{a=b;}}').getErrorCount() === 1);
+    });
+
+    it('should ignore method name if it\'s a reserved word (#180)', function() {
```
hzoo: Oh don't need this test?

markelog:
@mikesherov, @mrjoelkemp please check it out

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,149 @@
...
+ *     "while",
+ *     "do",
+ *     "try",
+ *     "catch",
+ *     "case",
+ *     "default"
+ * ]
+ * ```
+ *
+ * ##### Valid
```
mikesherov: Examples are backwards.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,149 @@
...
+            Array.isArray(statementTypes) || statementTypes === true,
+            'disallowCurlyBraces option requires array or true value'
+        );
+
+        if (statementTypes === true) {
+            statementTypes = defaultKeywords;
+        }
+
+        this._typeIndex = {};
+        for (var i = 0, l = statementTypes.length; i < l; i++) {
```
mikesherov: Use a forEach instead of for loop please.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,149 @@
...
+    getOptionName: function() {
+        return 'disallowCurlyBraces';
+    },
+
+    check: function(file, errors) {
+
+        function addError(typeString, node) {
+            var entity = node;
+
+            if (typeString === 'Else') {
```
mikesherov: Don't do this here. Do this at the point of detection.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,149 @@
...
+                if (typeIndex['else'] && node.alternate &&
+                    node.alternate.type === 'BlockStatement' &&
+                    node.alternate.type !== 'IfStatement'
+                ) {
+                    addError('Else', node);
+                }
+            });
+        }
+
+        if (typeIndex['case'] || typeIndex['default']) {
```
mikesherov: Use for notation here and all the places you currently aren't.

mikesherov: I don't think we should check for `case` and `default`. Arbitrarily long lists of statements, including multiple blocks can be children of case or default. They are in no way tied to the control structure, despite people occasionally thinking so.

hzoo:
@mikesherov I didn't change much from `requireCurlyBraces` but it's a good time to refactor/use the new functions we added if possible. I'l also fix the stuff you noted.

zxqfox:
Should we get rid of `Curly` in name? Braces are `{` and `}` already.

markelog:
There is a lot of redundancy and inconsistency in the rule names, everything of which should be addressed in 2.0. 

Since this is opposite rule it wouldn't makes sense to get rid of it now, but totally should be done in the future.

lib/rules/disallow-curly-braces.js
hzoo: What did you mean by "for notation here"

hzoo: And this rule only applies to single line statements since you are required to have braces otherwise right? Yeah most people don't put curly's in their case/default anyway - Oh and that was what the original issue was about.. he wanted to remove curly's for case/default!

hzoo:
Fixed the docs, used forEach, moved else check, removed check for `case` and `default` and associated tests, and made a `isBlockStatement` function

mikesherov:
@hzoo, I don't see how this rule allows:

```
if (x) {
  a();
  b();
}
```

which it needs to allow. Can you also please add some tests for that?

hzoo:
Oops - yeah that's important! added a check for multilines and tests

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,134 @@
...
+        return 'disallowCurlyBraces';
+    },
+
+    check: function(file, errors) {
+
+        function isBlockStatement(node) {
+            return node && node.type === 'BlockStatement';
+        }
+
+        function isMultiline(node) {
```
mikesherov: This is still wrong, as it'll miss:

```
if (a) {
  a = [
    'b'
  ];
}
```

Instead, count the number of statements inside the block statement, and error if it's 1.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,134 @@
...
+                }
+            });
+        }
+
+        var typeIndex = this._typeIndex;
+
+        if (typeIndex['if'] || typeIndex['else']) {
+            file.iterateNodesByType('IfStatement', function(node) {
+                if (typeIndex['if'] && isBlockStatement(node.consequent)) {
+                    if (isMultiline(node.consequent)) {
```
mikesherov: Move this check into `isBlockStatement` rather than calling it after every time.

lib/rules/disallow-curly-braces.js
hzoo: I wasn't sure what the method was - is it `node.body.length`?

test/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,149 @@
...
+        assert(checker.checkString([
+        'if (x) {',
+            'a();',
+            'b();',
+            'c();',
+        '}'
+        ].join('\n')).isEmpty());
+    });
+
+    it('should not report for a block with 3 lines with array', function() {
```
mikesherov: It SHOULD report this!

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,124 @@
...
+    getOptionName: function() {
+        return 'disallowCurlyBraces';
+    },
+
+    check: function(file, errors) {
+
+        function isSingleBlockStatement(node) {
+            console.log(node);
+            return node && node.type === 'BlockStatement' &&
+            // node.body.length === 1;
```
mikesherov: This is correct.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,124 @@
...
+            node.loc.end.line - node.loc.start.line <= 2;
```
mikesherov: This is wrong.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,124 @@
...
+    },
+
+    getOptionName: function() {
+        return 'disallowCurlyBraces';
+    },
+
+    check: function(file, errors) {
+
+        function isSingleBlockStatement(node) {
+            console.log(node);
```
mikesherov: Remove.

lib/rules/disallow-curly-braces.js
```diff
@@ -0,0 +1,124 @@
...
+
+        var typeIndex = this._typeIndex;
+
+        if (typeIndex['if'] || typeIndex['else']) {
+            file.iterateNodesByType('IfStatement', function(node) {
+                if (typeIndex['if'] && isSingleBlockStatement(node.consequent)) {
+                    addError('If', node);
+                }
+                if (typeIndex['else'] && isSingleBlockStatement(node.alternate) &&
+                    node.alternate.type !== 'IfStatement') {
```
mikesherov: This line will always be true if isSingleBlockStatement is true, and can therefore be removed.... right?

hzoo:
Ok I see you meant report an error whenever there is a single statement  - you would want them to change it so its a single line if it takes up multiple like with the array one.

mikesherov:
@hzoo, to be clear: 

```
if (a) {
  b = [
    c
  ];
}
```

can be rewritten to be either:

```
if (a) b = [c];
````

or

```
if (a) b = [
  c
];
```

. Both are valid javascript.

**Comment by mikesherov** [comment]
OK, I have one minor last change, but I'll make it when landing the PR. Thanks again @hzoo

**Comment by hzoo** [comment]
Ok cool

**Comment by mikesherov** [comment]
Thanks again @hzoo landed with a small change:

```
file.findPrevToken(
+                        file.getTokenByRangeStart(node.alternate.range[0]),
+                        'Keyword',
+                        'else'
+                    ));
```

to

```
file.getPrevToken(file.getFirstNodeToken(node.alternate))
```

for future reference.

hzoo:
Ok yeah I need to get more familiar with the API. We could also clean up `requireCurlyBraces` later

mikesherov:
@hzoo if you'd like to clean up requireCurlyBraces, now is a good time to do that. Thanks again!

Issue #887, opened 2015-01-07
TypeError: Cannot read property 'loc' of undefined for disallowNewlineBeforeBlockStatements


tteltrab:
Specifically, JSCS exits with this error if you have a file which begins with a block statement.  Using jscs 1.9.0 on OSX.  Specifically, files beginning with "{" cause the issue as there doesn't exist a token before the openingBrace token and thus the prevToken variable is undefined, causing the error.

markelog:
Thank you for the report, would you like to send us a pull request?

markelog:
Went ahead on this one, would need to fix this asap

tteltrab:
The pull request was #888, had it up for a day or so already.

markelog:
Did a simplified version of it, but still merged yours, since it was better

PR #888, opened 2015-01-07
disallowNewlineBeforeBlockStatements: check if prevToken is defined so t...

tteltrab:
...hat files starting with block statement don't throw error

Fixes: #887

### DIFF
--- a/lib/rules/disallow-newline-before-block-statements.js
+++ b/lib/rules/disallow-newline-before-block-statements.js
@@ -91,18 +91,17 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         file.iterateNodesByType('BlockStatement', function(node) {
-            var tokens = file.getTokens();
-
-            var openingBracePos = file.getTokenPosByRangeStart(node.range[0]);
-            var openingBrace = tokens[openingBracePos];
-            var prevToken = tokens[openingBracePos - 1];
-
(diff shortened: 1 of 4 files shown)


lib/rules/disallow-newline-before-block-statements.js
```diff
@@ -97,7 +97,8 @@ module.exports.prototype = {
             var openingBrace = tokens[openingBracePos];
             var prevToken = tokens[openingBracePos - 1];
 
-            if (openingBrace.loc.start.line !== prevToken.loc.start.line) {
+            if (typeof prevToken === 'object' &&
```
zxqfox: Nice catch & thanks for contributing!

Probably better to use here [new assertions](https://github.com/jscs-dev/node-jscs/blob/master/lib/token-assert.js#L81). Something like:
```
var openingBrace = file.getFirstNodeToken(node);
var prevToken = file.getPrevToken(openingBrace);
if (!prevToken) {
  return;
}

errors.assert.sameLine...
```

tteltrab:
ah, cool.  didn't know about those, but will do - could use an update.  And the strict type check is probably overkill, wasn't sure which way was preferred.

zxqfox:
@tteltrab Definitely ;-D

tteltrab:
Pull request is #888 I believe.

PR #889, opened 2015-01-08
Small improvements for token-assert module

markelog:
/cc @mdevils

### DIFF
--- a/lib/rules/disallow-comma-before-line-break.js
+++ b/lib/rules/disallow-comma-before-line-break.js
@@ -59,7 +59,7 @@ module.exports.prototype = {
             if (token.value === ',') {
                 errors.assert.sameLine({
                     token: token,
-                    nextToken: file.getNextToken(token),
+                    subjectToken: file.getNextToken(token),
                     message: 'Commas should be placed on new line'
                 });
             }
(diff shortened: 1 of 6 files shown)


mdevils:
Hi, Oleg.

Checked out this PR. For me it became less clear.

@mikesherov, can you please check this out?

markelog:
> Checked out this PR. For me it became less clear.

Could you elaborate?

markelog:
Will merge this for now, since i need this for further work on other things, will revert this, if these changes would not be acceptable

mrjoelkemp:
What motivated the property name change? I agree with Murat that `nextToken` was a better name. I'd think what is currently `token` should be `prevToken` and `nextToken/subjectToken` should just be `token` or `currentToken`.

markelog:
This `nextToken/subjectToken` could be very well previous or the next token, so i would choose a name which would be agnostic to the direction, since "direction" doesn't matter, because it doesn't reflected in checks of those methods.  Same with `token` property.

Bikeshedding process is as fun as i remember it.

mdevils:
Errr... why did you merge? For me `nextToken` is way clearer than `subjectToken`.

mikesherov:
@mdevils

> Will merge this for now, since i need this for further work on other things, will revert this, if these changes would not be acceptable

I think Oleg just wanted some solid ground to work off of. 

I do think though that "next" and "previous" imply position in text. @markelog's desire to rename them is good. I am not sure "subjectToken" is a good name either. 

@markelog, you can't start painting a red shed blue and then complain how not fun it is to paint!

PR #890, opened 2015-01-09
Add getReporter method to cli-config module

markelog:
Since this method should be used by the third-party, i think we can release it only in the minor release.


mrjoelkemp:
:+1:

Issue #721, opened 2014-10-26
Add new values and deprecate some for (require | disallow)SpacesInsideObjectBrackets


markelog:
Should use `expect` and `only` values. Deprecated use of `nested` and `allButNested`.

Just like `(require | disallow)SpacesInsideParentheses` rule

PR #893, opened 2015-01-09
Add new values and deprecate some for (require | disallow)SpacesInsideObjectBrackets

markelog:
See https://github.com/jscs-dev/node-jscs/issues/721, although without `only` but with `allExcept` for both

### DIFF
--- a/lib/rules/disallow-spaces-inside-object-brackets.js
+++ b/lib/rules/disallow-spaces-inside-object-brackets.js
@@ -1,14 +1,20 @@
 /**
  * Disallows space after opening object curly brace and before closing.
  *
- * Type: `Boolean` or `String`
+ * Type: `Object`, `Boolean` or `String`
  *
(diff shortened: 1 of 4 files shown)


markelog:
> Coverage decreased (-0.04%) 

Because of the configure `if`s, i find it tedious to add tests for it, should be resolved with https://github.com/jscs-dev/node-jscs/issues/895 instead

lib/rules/disallow-spaces-inside-object-brackets.js
```diff
@@ -20,7 +26,13 @@
  * ##### Valid for mode `"nested"`
  *
  * ```js
- * var x = { a: {b: 1} };
+ * var x = {a: {b: 1} };
```
zxqfox: Space missed?

markelog: Nope, it was a mistake, see description of the rule

Issue #111, opened 2013-12-27
`requireSpacesInsideArrayBrackets` to ignore object literals


SBoudrias:
Right now when setting `requireSpacesInsideArrayBrackets`, you can ignore on nested or require spaces everywhere.

It would be useful to be able to also ignore spaces if an object literal is provided, e.g.:

```
[{ foo: 'bar' }]
[{
  foo: 'bar'
}, {
  foo: 'bar'
}]

// but require
[ 'foo' ]
```

Issue #429, opened 2014-06-10
Edge cases for *SpacesInsideArrayBrackets rule


markelog:
Esprima will treat code like this:

```js
[1][0]
```
as one array and one expression, i.e. `*SpacesInsideArrayBrackets` will fail on cases like this.

PR #896, opened 2015-01-10
Improve spaces inside array brackets rules

### LINKED ISSUE #721: Add new values and deprecate some for (require | disallow)SpacesInsideObjectBrackets

markelog:
Fixes #721 and #429

### DIFF
--- a/lib/rules/disallow-spaces-inside-array-brackets.js
+++ b/lib/rules/disallow-spaces-inside-array-brackets.js
@@ -3,12 +3,19 @@
  *
  * Type: `Boolean` or `String`
  *
- * Values: `"all"` or `true` for strict mode, `"nested"` ignores closing brackets in a row.
+ * Values: `"all"` or `true` for strict mode, `"nested"` (*deprecated* use `"allExcept": [ "[", "]" ]`)
+ * ignores closing brackets in a row.
  *
  * #### Example
  *
  * ```js
  * "disallowSpacesInsideArrayBrackets": "all"
+ *
+ * // or
+ *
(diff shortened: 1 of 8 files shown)


markelog:
> Coverage decreased (-0.02%)
Same as with https://github.com/jscs-dev/node-jscs/pull/893#issuecomment-69401817

presets/wikimedia.json
```diff
@@ -24,7 +24,7 @@
     "requireBlocksOnNewline": 1,
     "disallowEmptyBlocks": true,
     "requireSpacesInsideObjectBrackets": "all",
-    "requireSpacesInsideArrayBrackets": "all",
+    "disallowSpacesInsideArrayBrackets": "all",
```
markelog: /cc @jdlrobson, @jdforrester, @Krinkle 

Is that OK?

markelog:
Wow, it would also fix https://github.com/jscs-dev/node-jscs/issues/111

PR #900, opened 2015-01-12
new rule values: emptyLines for disableTrailingWhitespace and validateIndentation

### LINKED ISSUE #719: Optionally ignore empty lines in disallowTrailingWhitespace
The disallowTrailingWhitespace rule reports validation errors if an otherwise empty line contains whitespace (e.g. automatically inserted by an editor due to indentation).

It would be nice if there was an option for this rule to ignore empty lines.

JonathanGawrych:
Fixes #719 

The new rule value `ignoreEmptyLines` for `disableTrailingWhitespace` allows for indentation (any other trailing whitespace) on lines that have no other content. This allows for a block to continually have the same indentation, even over empty lines.  This new value also match's jshint's old functionality [of skipping empty lines](https://github.com/jshint/jshint/commit/0c0e19319b276b019d285bdfd2cfc49126abd814#diff-ffa5bd6fde803b83317a7f9e87b5bbe1L1320).

The new rule value `includeEmptyLines` for `validateIndentation` requires indentation on lines that have no other content. This is similar to the ignoreEmptyLines above, but instead of not throwing an error on empty lines that have whitespace, this throws errors for empty lines that do not match the current indentation.

I battled with the thought of breaking out these two new rule values into separate rules (`(disallow|require)IndentationOnEmptyLines`) but after putting some thought and code into it, I decided the above way would be better. If they were rules, they would both need to know what the indentation config was set to, would require very similar code to `validateIndentation`, would conflict with `disableTrailingWhitespace`, change the original meaning of the two other rules, and just wasn't a good way to approach the problem.

My only small problem was that for doc changes, jscs would complain for my example trailing whitespace. I used `"\t"` in one place and a comment in the other to overcome that, but I wasn't sure how you would want that invisible character represented.

Thanks!

### DIFF
--- a/lib/rules/validate-indentation.js
+++ b/lib/rules/validate-indentation.js
@@ -1,9 +1,14 @@
 /**
  * Validates indentation for switch statements and block statements
  *
- * Type: `Integer` or `String`
+ * Type: `Integer` or `String` or `Object`
  *
- * Values: A positive integer or `"\t"`
+ * Values:
+ *  - `Integer`: A positive number of spaces
+ *  - `String`: `"\t"` for tab indentation
+ *  - `Object`:
+ *     - `value`: (required) the same effect as the non-object values
+ *     - `includeEmptyLines`: (default: `false`) require empty lines to be indented
  *
  * JSHint: [`indent`](http://jshint.com/docs/options/#indent)
  *
@@ -35,7 +40,7 @@
  * }
  * ```
  *
- * ##### Valid example for mode "\t"
+ * ##### Valid example for mode `"\t"`
  *
  * ```js
  * if (a) {
@@ -46,7 +51,7 @@
  * }
  * ```
  *
- * ##### Invalid example for mode "\t"
+ * ##### Invalid example for mode `"\t"`
  *
  * ```js
  * if (a) {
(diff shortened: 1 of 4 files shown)


lib/rules/validate-indentation.js
```diff
@@ -366,6 +410,15 @@ module.exports.prototype = {
                     linesToCheck[startLine].push = ((newIndents - actualIndents) / indentSize) + 1;
                 }
             });
+
+            if (_this._includeEmptyLines) {
+                var lines = file.getLines();
```
mikesherov:
```
file.getLines.forEach(function(line) {
  if (line.match(/^\s*$/)) {
    linesToCheck[i].check = true;
  } 
});
```

a bit more concise to just forEach.

lib/rules/validate-indentation.js
```diff
@@ -66,11 +93,25 @@ module.exports = function() {};
...
     configure: function(validateIndentation) {
-        assert(
-            validateIndentation === '\t' ||
-                (typeof validateIndentation === 'number' && validateIndentation > 0),
-            'validateIndentation option requires a positive number of spaces or "\\t"'
-        );
+        this._includeEmptyLines = false;
+
+        if (typeof validateIndentation === 'object') {
+            assert(
```
mikesherov: figure out how to avoid the duplicate assertion here... perhaps:

```
this._includeEmptyLines = (validateIndentation.includeEmptyLines === true);
validateIndentation = validateIndentation.value;
```

dropping the else, and then just asserting after?

mikesherov:
@JonathanGawrych please rebase so I can continue reviewing this further. Thanks again for contributing!

lib/rules/validate-indentation.js
JonathanGawrych: updated to use forEach rather than for loop

lib/rules/validate-indentation.js
JonathanGawrych: consolidated into one assert!

JonathanGawrych:
Wow, that was one heck of a rebase. I believe there were 12 commits between when I branched and now. Tests for disableTrailingWhitespace/validateIndention  are passing, but `disallow-spaces-inside-parentheses` is not. I found that this test is also failing on master, so I ignored it.

mikesherov:
Landed! Thanks again so much for contributing @JonathanGawrych !

PR #902, opened 2015-01-12
ESNext: update esprima to properly parse regex tokens

mikesherov:


mikesherov:
@markelog or @mdevils or @mrjoelkemp please review and if LGTM'ed, please merge before we release 1.10

mdevils:
LGTM.

What means `strict`?

lib/string-checker.js
```diff
@@ -89,7 +89,15 @@ StringChecker.prototype = {
         var parseError;
 
         try {
-            tree = this._esprima.parse(str, {loc: true, range: true, comment: true, tokens: true});
+            tree = this._esprima.parse(str, {
+                loc: true,
+                range: true,
+                comment: true,
+                tokens: true,
+                // TODO: sourceType is required to parse 'import' and 'export' statements,
```
markelog: Why this is `TODO`, we can change this later?

lib/string-checker.js
```diff
@@ -89,7 +89,15 @@ StringChecker.prototype = {
...
+                // but has the side effect of forcing all code to be considered strict
```
markelog: Is there a relevant link about this?

lib/string-checker.js
mikesherov: At a later date, we may consider revising this? I suppose I should just leave an issue, and not put the todo in code. Amending.

mikesherov:
By `strict` I mean strict mode. Because modules, as per spec, are always in strict mode. 

But if a user want to use es6, and not modules, they could theoretically want non strict mode. and then we'd need to provide the option of whether to say module or not as an esprima option, which is annoying and we can save for later until someone complains.

markelog:
Could you mention in the commit what issues do you fix?

mikesherov:
@markelog yes, will do.

markelog:
LGTM

Issue #109, opened 2013-12-27
Make `preset` more useful


sindresorhus:
re #54 

As it stands now `preset` isn't very useful as it only has the `jquery` style. Seeing as the config is fairly verbose it would be nice to have some ways to reuse it.

Would be useful if `preset` could have a searchpath in this order:

1. Check it with fs.exists if it's a valid path to a preset file
2. Try to require a node module with that name. Super useful for being able to distribute presets. Since `"main"` in package.json can be a .json file it could be the preset and requiring the module, would give you the preset config object.
3. the built-in preset.

mikesherov:
Can we get this done in 1.12? /cc @mdevils @zxqfox seems we just need to do the original suggestion and can defer the question of loading multiple presets together to a different issue at a different time.

I'll assign @zxqfox for now just so someone owns this.

markelog:
Can take this one if @zxqfox don't mind

zxqfox:
@markelog Feel free! I wanted to have deep extension here if possible: https://github.com/jscs-dev/jscs-jsdoc/issues/83 . There were idea to split google (with closurecompiler specific tags), wiki (with other set), etc.

mikesherov:
Yeah, sounds good to me too. Just trying to get old issues addressed.

zxqfox:
I agree that we can solve original issue (according to current roadmap) and make another issue for future. Just need to know what we can solve here.

mikesherov:
Ability to load a preset by use of `require` so that presets can be distributed via npm.

zxqfox:
@mikesherov We already have `plugins` functionality and it's pretty simple to make it: https://github.com/zxqfox/jscs-bem/blob/master/lib/index.js

I can suggest defining "jscs" or "jscs-plugin" right in package.json with "presets" section if we need something more but it looks pretty complex.
```
{
  "jscs-plugin": {
    "presets": ["hey-ho", "lets-go"]
  }
}
```

We already have package.json usage: https://github.com/jscs-dev/node-jscs/blob/master/test/data/configs/mixedWithPkg/package.json

markelog:
@zxqfox preset option might be simple alternative to the plugin interface, so it would be easy to share such configs without digging into the API.

But i agree, it looks like creating plugin is easy too.

@mikesherov thoughts? 
@sindresorhus would love to hear your opinion about that too.

zxqfox:
[...]

PR #908, opened 2015-01-13
Configuration: Auto-generation

### LINKED ISSUE #797: Autoconfiguration option
JSCS should have a cli option called `--configure`. When ran from the root of your project, it will do the following:

1. ask you to specify the location of your javascript files. 
2. run each preset against your codebase.
3. output the number of violations from each preset with a link to the style guide.
4. ask you to choose a preset based on the output of 3.
5. go through the remaining violations, rule by rule. For each violation, it will ask: "Would you like to create an exception for this rule, or fix this issue?"
  1. for each "exception" created, it will nullfy that rule in .jscsrc.
  2. for each "fix this issue", in the future, will automatically fix it, but for now require you to fix it.
6. at the end, it will output a .jscsrc file for you to commit yourself.

Supercedes #341 

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/6604741-autoconfiguration-option?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mrjoelkemp:
TODO:

- [x] Fix bug when using auto-config with a directory
- [x] Handle case when user does not supply a path for auto-configure. Show error message
- [x] Readme
- [ ] Extract and test my custom promisify for 100% coverage (Only if still needed after Inquire switch) https://github.com/jscs-dev/node-jscs/issues/999
- [ ] Use https://github.com/SBoudrias/Inquirer.js instead of prompt for better error handling https://github.com/jscs-dev/node-jscs/issues/1000
- [x] Add test for cli with `--auto-configure` used without a string value (empty defaults to boolean true)
- [ ] Add test for an es6 codebase. We may need to have the esnext flag passed through to the checker used in generator https://github.com/jscs-dev/node-jscs/issues/1001
- [ ] Add tests for erroneous input handling https://github.com/jscs-dev/node-jscs/issues/1002
- [ ] Speed up auto-configure tests https://github.com/jscs-dev/node-jscs/issues/1003

---

- [x] Refactor
- [x] Tests
 - Choosing a preset with no errors
 - Choosing a preset with errors
 - Console logging
- [x] Fix commit message
- [ ] Check that prompt works when the terminal is resized. Seemed broken when stretching terminal to fit everything on one line. (Could be fixed with a move to Inquire, but not sure) #1000

Fixes #797

### DIFF
--- a/bin/jscs
+++ b/bin/jscs
@@ -15,6 +15,7 @@ program
     .usage('[options] <file ...>')
     .description('A code style linter for programmatically enforcing your style guide.')
     .option('-c, --config [path]', 'configuration file path')
+    .option('--auto-configure [path]', 'auto-generate a JSCS configuration file')
     .option('-e, --esnext', 'attempts to parse esnext code (currently es6)')
     .option('--es3', 'validates code as es3')
     .option('-s, --esprima <path>', 'attempts to use a custom version of Esprima')
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -7,6 +7,7 @@
  */
 var Checker = require('./checker');
 var configFile = require('./cli-config');
+var ConfigGenerator = require('./config/generator');
 
 var Vow = require('vow');
 var supportsColor = require('supports-color');
@@ -50,7 +51,7 @@ module.exports = function(program) {
      * Trying to load config.
      * Custom config path can be specified using '-c' option.
      */
-    if (!config && !program.preset) {
+    if (!config && !program.preset && !program.autoConfigure) {
         if (program.config) {
             console.error('Configuration source', program.config, 'was not found.');
         } else {
@@ -62,7 +63,7 @@ module.exports = function(program) {
         return returnArgs;
     }
 
-    if (!args.length && process.stdin.isTTY) {
+    if (!args.length && process.stdin.isTTY && typeof program.autoConfigure !== 'string') {
         console.error('No input files specified. Try option --help for usage information.');
         defer.reject(1);
 
@@ -93,6 +94,20 @@ module.exports = function(program) {
     } catch (e) {
         console.error(e.message);
         defer.reject(1);
+
+        return returnArgs;
+    }
+    if (program.autoConfigure) {
+        var generator = new ConfigGenerator();
+
+        generator
+        .generate(program.autoConfigure)
+        .then(function() {
+            defer.resolve(0);
+        }, function(error) {
+            console.error('Configuration generation failed due to ', error);
+            defer.reject(1);
+        });
 
         return returnArgs;
     }
--- a/lib/config/generator.js
+++ b/lib/config/generator.js
@@ -1,0 +1,274 @@
+var path = require('path');
+var fs = require('fs');
+
+var Vow = require('vow');
+var Table = require('cli-table');
+var prompt = require('prompt');
+var colors = require('colors');
+var assign = require('lodash.assign');
+
+var Checker = require('../checker');
+var utils = require('../utils');
+
+prompt.message = '';
+prompt.delimiter = '';
+prompt.start();
+
+/**
+ * Script that walks the user through the autoconfig flow
+ *
+ * @type {String[]}
+ * @private
+ */
+var prompts = [
+    {
+        name: colors.green('Please choose a preset number:'),
+        require: true,
+        pattern: /\d+/
+    },
+    {
+        name: 'Create an (e)xception for this rule, or (f)ix the errors yourself?',
+        require: true,
+        pattern: /(e|f)/
+    }
+];
+
+/**
+ * JSCS Configuration Generator
+ *
+ * @name Generator
+ */
+function Generator() {
+    this._config = {};
+}
+
+/**
+ * Generates a configuration object based for the given path
+ * based on the best fitting preset
+ *
+ * @param  {String} path - The path containing file(s) used to guide the configuration
+ *
+ * @return {Promise} Resolved with the generated, JSCS configuration
+ */
+Generator.prototype.generate = function(path) {
+    var checker = getChecker();
+    var _path = utils.normalizePath(path, checker.getConfiguration().getBasePath());
+    var presetNames = Object.keys(checker.getConfiguration().getRegisteredPresets());
+    var statsForPresets;
+
+    console.log('Checking', _path, 'against the presets');
+
+    return Vow
+    .all(presetNames.map(this._checkAgainstPreset.bind(this, _path)))
+    .then(function(resultsPerPreset) {
+        statsForPresets = this._generateStatsForPresets(resultsPerPreset, presetNames);
+        return statsForPresets;
+    }.bind(this))
+    .then(this._showErrorCounts.bind(this))
+    .then(this._getUserPresetChoice.bind(this, prompts[0]))
+    .then(function showViolatedRules(choiceObj) {
+        var presetIndex = choiceObj[prompts[0].name] - 1;
+        var presetName = statsForPresets[presetIndex].name;
+
+        console.log('You chose the ' + presetName + ' preset');
+
+        this._config.preset = presetName;
+
+        var errorList = statsForPresets[presetIndex].errors;
+        errorList = getUniqueErrorNames(errorList);
+
+        var violatedRuleCount = errorList.length;
+
+        if (!violatedRuleCount) { return this._config; }
+
+        console.log(_path + ' violates ' + violatedRuleCount + ' rule' + (violatedRuleCount > 1 ? 's' : ''));
+
+        var errorPrompts = generateRuleHandlingPrompts(errorList);
+
+        return this._getUserViolationChoices(errorPrompts)
+        .then(this._handleViolatedRules.bind(this, errorPrompts, errorList))
+        .then(function() {
+            return this._config;
+        }.bind(this));
+    }.bind(this))
+    .then(function flushConfig(config) {
+        fs.writeFileSync(process.cwd() + '/.jscsrc', JSON.stringify(this._config, null, '\t'));
+        console.log('Generated a .jscsrc configuration file in ' + process.cwd());
+    }.bind(this));
+};
+
+/**
+ * @private
+ * @param  {Object[][]} resultsPerPreset - List of error objects for each preset's run of checkPath
+ * @param  {String[]} presetNames
+ * @return {Object[]} Aggregated datapoints for each preset
+ */
+Generator.prototype._generateStatsForPresets = function(resultsPerPreset, presetNames) {
+    return resultsPerPreset.map(function(presetResults, idx) {
+        var errorCollection = [].concat.apply([], presetResults);
+
+        var presetStats = {
+            name: presetNames[idx],
+            sum: 0,
+            errors: []
+        };
+
+        errorCollection.forEach(function(error) {
+            presetStats.sum += error.getErrorCount();
+            presetStats.errors = presetStats.errors.concat(error.getErrorList());
+        });
+
+        return presetStats;
+    });
+};
+
+/**
+ * @private
+ * @param  {Object[]} statsForPresets
+ */
+Generator.prototype._showErrorCounts = function(statsForPresets) {
+    var table = getTable();
+
+    statsForPresets.forEach(function(presetStats, idx) {
+        table.push([idx + 1, presetStats.name, presetStats.sum]);
+    });
+
+    console.log(table.toString());
+};
+
+/**
+ * Prompts the user to choose a preset
+ *
+ * @private
+ * @param {Object} prompt
+ * @return {Promise}
+ */
+Generator.prototype._getUserPresetChoice = function(prompt) {
+    return this._showPrompt(prompt);
+};
+
+/**
+ * Prompts the user to nullify rules or fix violations themselves
+ *
+ * @private
+ * @param  {Object[]} errorPrompts
+ * @return {Promise}
+ */
+Generator.prototype._getUserViolationChoices = function(errorPrompts) {
+    return this._showPrompt(errorPrompts);
+};
+
+function promisify(fn) {
+    return function() {
+        var deferred = Vow.defer();
+        var args = [].slice.call(arguments);
+
+        args.push(function(err, result) {
+            if (err) {
+                deferred.reject(err);
+            } else {
+                deferred.resolve(result);
+            }
+        });
+
+        fn.apply(null, args);
+
+        return deferred.promise();
+    };
+}
+
+/** @private */
+Generator.prototype._showPrompt = promisify(prompt.get.bind(prompt));
+
+/**
+ * @private
+ * @param  {Object[]} errorPrompts
+ * @param  {String[]} errorList
+ * @param  {Object[]} choices
+ */
+Generator.prototype._handleViolatedRules = function(errorPrompts, errorList, choices) {
+    errorPrompts.forEach(function(errorPrompt, idx) {
+        var associatedRuleName = errorList[idx];
+        var userChoice = choices[errorPrompt.name];
+
+        if (userChoice.toLowerCase() === 'e') {
+            this._config[associatedRuleName] = null;
+        }
+    }, this);
+};
+
+/**
+ * @private
+ * @param  {String} path
+ * @param  {String} presetName
+ * @return {Promise}
+ */
+Generator.prototype._checkAgainstPreset = function(path, presetName) {
+    var checker = getChecker();
+
+    checker.configure({preset: presetName});
+
+    return checker.checkPath(path);
+};
+
+/**
+ * @private
+ * @return {lib/Checker}
+ */
+function getChecker() {
+    var checker = new Checker();
+    checker.registerDefaultRules();
+    return checker;
+}
+
+/**
+ * @private
+ * @param  {String[]} violatedRuleNames
+ * @return {Object[]}
+ */
+function generateRuleHandlingPrompts(violatedRuleNames) {
+    return violatedRuleNames.map(function(ruleName) {
+        var prompt = assign({}, prompts[1]);
+        prompt.name = colors.green(ruleName) + ': ' + prompt.name;
+        return prompt;
+    });
+}
+
+/**
+ * @private
+ * @param  {Object[]} errorsList
+ * @return {String[]}
+ */
+function getUniqueErrorNames(errorsList) {
+    var errorNameLUT = {};
+
+    errorsList.forEach(function(error) {
+        errorNameLUT[error.rule] = true;
+    });
+
+    return Object.keys(errorNameLUT);
+}
+
+/**
+ * @private
+ * @return {Object}
+ */
+function getTable() {
+    return new Table({
+        chars: {
+            top: '', 'top-mid': '', 'top-left': '', 'top-right': '',
+            bottom: '', 'bottom-mid': '', 'bottom-left': '', 'bottom-right': '',
+            left: '', 'left-mid': '',
+            mid: '', 'mid-mid': '',
+            right: '', 'right-mid': '' ,
+            middle: ' '
+        },
+        style: {
+            'padding-left': 0,
+            'padding-right': 0
+        },
+        head: ['', 'Preset', '#Errors']
+    });
+}
+
+module.exports = Generator;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,3 +1,5 @@
+var path = require('path');
+
 // 7.5.2 Keywords
 var ES3_KEYWORDS = {
     'break': true,
@@ -184,6 +186,33 @@ exports.getFunctionNodeFromIIFE = function(node) {
 exports.trimUnderscores = function(name) {
     var res = name.replace(TRAILING_UNDERSCORES_RE, '');
     return res ? res : name;
+};
+
+/**
+ * Whether or not the given path is relative
+ * @param  {String}  path
+ * @return {Boolean}
+ */
+exports.isRelativePath = function(path) {
+    // Logic from: https://github.com/joyent/node/blob/4f1ae11a62b97052bc83756f8cb8700cc1f61661/lib/module.js#L237
+    var start = path.substring(0, 2);
+    return start === './' || start === '..';
+};
+
+/**
+ * Resolves a relative filepath against the supplied base path
+ * or just returns the filepath if not relative
+ *
+ * @param  {String} filepath
+ * @param  {String} basePath
+ * @return {String}
+ */
+exports.normalizePath = function(filepath, basePath) {
+    if (this.isRelativePath(filepath)) {
+        return path.resolve(basePath, filepath);
+    }
+
+    return filepath;
 };
 
 /**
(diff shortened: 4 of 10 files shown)


lib/config/generate.js
```diff
@@ -0,0 +1,169 @@
...
+
+    return Object.keys(errorNameLUT);
+}
+
+function getPresetChoice(choice, presetNames) {
+    var presetIndex = choice[prompts[0]] - 1;
+    var presetName = presetNames[presetIndex];
+
+    if (!presetName) {
+        console.error('You supplied an invalid preset number');
```
mrjoelkemp: Instead, try to use prompt's built-in validation

lib/config/generate.js
```diff
@@ -0,0 +1,169 @@
...
+    prompt.start();
+    prompt.message = '';
+    prompt.delimiter = '';
+
+    return Vow.all(presetNames.map(checkAgainstPreset.bind(null, _path)))
+    .then(function(results) {
+        var errorsCollection = [].concat.apply([], results);
+        var errorCounts = getErrorCounts(errorsCollection);
+
+        outputPresetErrorCounts(presetNames, errorCounts);
```
mrjoelkemp: Only output unique errors to avoid count discrepancy between the table and the number of rules the user is walked through for fixup/exception.

markelog:
I think we can release 1.11 right after it's done.

/cc @mikesherov @mdevils

mikesherov:
Looks good what you have so far. 

One thought is that when no config is found, if you're TTY, go straight to auto configure mode. If you're not TTY, mention the configure option in the error output.

mikesherov:
Second thought is that if you run configure mode without a path, prompt the user for the path. Combined with my previous suggestion, the user can just literally do: `npm install jscs && jscs` from their root and be walked through the setup.

indexzero:
Can't want to try this out. I basically spent week making config files before >_<

lib/config/generate.js
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Generates a configuration object based for the given path
+ * based on the best fitting preset
+ *
+ * @param  {String} path - The path containing file(s) used to guide the configuration
+ *
+ * @return {Promise} Resolved with the generated, JSCS configuration
+ */
+function generateConfig(path) {
```
mdevils: I think a simple class would be nicer and more obvious here.

mdevils:
Great job, Joel!

lib/config/generate.js
mrjoelkemp: Agreed about the clarity. Forcing an instantiation for what should be a function didn't seem correct. This function only looks like a class because the tests need access to internal methods for stubbing. 

I'll convert this to a class though; thanks for reaffirming.

Tests ruin everything :)

lib/config/generator.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+};
+
+function getChecker() {
+    var checker = new Checker();
+    checker.registerDefaultRules();
+    return checker;
+}
+
+function normalizePath(filepath, basePath) {
```
mikesherov: This should be extracted to utils too. This same "if relative, then resolve" code is in 4 spots.

lib/config/generator.js
```diff
@@ -0,0 +1,222 @@
...
+                    _this._processViolationChoice(userChoice, associatedRuleName);
+                });
+
+                return this._config;
+            });
+        }.bind(this))
+        .then(function flushConfig(config) {
+            fs.writeFileSync(process.cwd() + '/.jscsrc', JSON.stringify(config, null, '\t'));
+            console.log('Generated a .jscsrc configuration file in ' + process.cwd());
+        });
```
mikesherov: Do you put a final `.catch` here for unexpected errors? Or this handled by the code this promise is returned to?

lib/config/generator.js
```diff
@@ -0,0 +1,222 @@
...
+
+        return Vow
+        .all(presetNames.map(this._checkAgainstPreset.bind(this, _path)))
+        .then(function generateErrorsCollection(results) {
+            errorsCollection = [].concat.apply([], results);
+            return errorsCollection;
+        })
+        .then(this._showErrorCounts.bind(this, presetNames))
+        .then(this._getUserPresetChoice.bind(this, prompts[0]))
+        .then(function getPresetIndexFromPrompt(choiceObj) {
```
mikesherov: Combine this and the next two thens. If you want to use functions for clarity, go ahead, but no need for multiple thens, it's just noisy.

lib/config/generator.js
```diff
@@ -0,0 +1,222 @@
...
+            var errorPrompts = generateRuleHandlingPrompts(errorList);
+
+            // TODO: Extract method?
+            return this._getUserViolationChoices(errorPrompts)
+            .then(function handleViolatedRules(choices) {
+                errorPrompts.forEach(function(errorPrompt, idx) {
+                    var associatedRuleName = errorList[idx].rule;
+                    var userChoice = choices[errorPrompt];
+
+                    _this._processViolationChoice(userChoice, associatedRuleName);
```
mikesherov: You avoided `_this` everywhere, why not here? Just pass `this` as second arg to forEach.

lib/config/generator.js
```diff
@@ -0,0 +1,222 @@
...
+        .then(function showViolatedRules(errorList) {
+            var violatedRuleCount = getUniqueErrorNames(errorList).length;
+
+            if (!violatedRuleCount) { return this._config; }
+
+            console.log(_path + ' violates ' + violatedRuleCount + ' rule' + (violatedRuleCount > 1 ? 's' : ''));
+
+            var errorPrompts = generateRuleHandlingPrompts(errorList);
+
+            // TODO: Extract method?
```
mikesherov: Yes.

lib/string-checker.js
```diff
@@ -58,6 +58,22 @@ StringChecker.prototype = {
     },
 
     /**
+     * Gets the registered presets
+     * @return {Object}
+     */
+    getPresets: function() {
```
mikesherov: Needs a test.

lib/string-checker.js
```diff
@@ -58,6 +58,22 @@ StringChecker.prototype = {
...
+        return this._configuration.getPresets();
+    },
+
+    /**
+     * Gets the base path
+     * @return {String}
+     */
+    getBasePath: function() {
```
mikesherov: Needs a test.

lib/config/configuration.js
```diff
@@ -157,6 +157,15 @@ Configuration.prototype.getMaxErrors = function() {
 };
 
 /**
+ * Returns the list of registered presets
+ *
+ * @return {String[]}
+ */
+Configuration.prototype.getPresets = function() {
```
mikesherov: Needs a test.

lib/config/generator.js
mikesherov: Actually, this and next *3* thens.

test/string-checker.js
```diff
@@ -356,4 +356,24 @@ describe('modules/string-checker', function() {
             });
         }
     });
+
+    describe('getRegisteredPresets', function() {
+        it('returns the configuration\'s registered presets', function() {
```
mikesherov: This test should actually check that it returns registered presets. The fact that it's a middleman is implementation detail.

test/string-checker.js
```diff
@@ -356,4 +356,24 @@ describe('modules/string-checker', function() {
...
+            var configuration = checker.getConfiguration();
+            var spy = sinon.spy(configuration, 'getRegisteredPresets');
+            checker.getRegisteredPresets();
+            assert.ok(spy.called);
+            spy.restore();
+        });
+    });
+
+    describe('getBasePath', function() {
```
mikesherov: Same middleman comment.

test/string-checker.js
mrjoelkemp: There's already a test for getRegisteredPreset in configuration and this function's sole role is to delegate to that tested method. Doing checks on the returned value is redundant.

What's the minimum checking here that you feel is appropriate?

mikesherov: Then either undelegate the method or copy the test from configuration.

mrjoelkemp: In all fairness, the other delegated methods in string-checker don't have dedicated tests. I'll revert my hide delegate.

mikesherov:
Wooo! Looks like testing's done?

mikesherov:
Just wanted to give a quick brain dump of follow up tickets: 1. run configure if JSCS ran without a Config and in tty. 2. detect esnext.

bin/jscs
```diff
@@ -15,6 +15,7 @@ program
     .usage('[options] <file ...>')
     .description('A code style linter for programmatically enforcing your style guide.')
     .option('-c, --config [path]', 'configuration file path')
+    .option('--configure [path]', 'auto-generate a JSCS configuration file')
```
markelog: `--configure` and `--config`, `-c` are very similar. Can we try another name for it?

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+prompt.delimiter = '';
+prompt.start();
+
+/**
+ * Script that walks the user through the autoconfig flow
+ * @type {String[]}
+ */
+var prompts = [
+    {
+        name: 'Please choose a preset number:'.green,
```
markelog: I guess `colors` is used implicitly, i would add it explicitly

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
+var path = require('path');
+var Vow = require('vow');
+var vowNode = require('vow-node');
+var Table = require('cli-table');
+var prompt = require('prompt');
+var fs = require('fs');
```
markelog: Nit-pick, could we change sequence of those definition like we do in other modules?

Like this:

```js

// Default node modules
var path = require('path');
var fs = require('fs');

// Deps
var Vow = require('vow');
var vowNode = require('vow-node');
var Table = require('cli-table');
var prompt = require('prompt');

// Our stuff
var Checker = require('../checker');
var utils = require('../utils');
```

bin/jscs
zxqfox: `--auto-config` or `--auto-generate`?

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+        pattern: /[0-9]*/
+    },
+    {
+        name: 'Create an (e)xception for this rule, or (f)ix the errors yourself?',
+        require: true,
+        pattern: /(e|f)/
+    }
+];
+
+function Generator() {
```
markelog: Could we add JSDoc to the constructor?

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+    this._config = {};
+}
+
+Generator.prototype = {
```
markelog: Nit-pick: it seems, constructor ref is now lost?

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+var Checker = require('../checker');
+var utils = require('../utils');
+
+prompt.message = '';
+prompt.delimiter = '';
+prompt.start();
+
+/**
+ * Script that walks the user through the autoconfig flow
+ * @type {String[]}
```
markelog: @private?

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+     * @private
+     * @param  {Object[]} errorPrompts
+     * @return {Promise}
+     */
+    _getUserViolationChoices: function(errorPrompts) {
+        return this._showPrompt(errorPrompts);
+    },
+
+    /** @private */
+    _showPrompt: vowNode.promisify(prompt.get.bind(prompt)),
```
markelog: It seems as overkill, to add a whole module just for one function

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+    _checkAgainstPreset: function(path, presetName) {
+        var checker = getChecker();
+
+        checker.configure({preset: presetName});
+
+        return checker.checkPath(path);
+    }
+};
+
+function getChecker() {
```
markelog: Nit-pick: missing JSDoc for those helpers?

lib/utils.js
```diff
@@ -158,6 +160,47 @@ exports.trimUnderscores = function(name) {
...
+    return start === './' || start === '..';
+};
+
+/**
+ * Shallow extend
+ * @param  {Object} obj1 - Object to be extended
+ * @param  {Object} obj2
+ * @return {Object} A reference to obj1
+ */
+exports.extend = function(obj1, obj2) {
```
markelog: Maybe we just should a `lodash` dep?

lib/utils.js
```diff
@@ -158,6 +160,47 @@ exports.trimUnderscores = function(name) {
...
+
+/**
+ * Resolves a relative filepath against the supplied base path
+ * or just returns the filepath if not relative
+ *
+ * @param  {String} filepath
+ * @param  {String} basePath
+ * @return {String}
+ */
+exports.normalizePath = function(filepath, basePath) {
```
markelog: Should it be in `utils` if it used in one module? In some projects `utils` quickly becomes a collection of weird functions, could we avoid that?

lib/utils.js
zxqfox: It weights much more, even if https://www.npmjs.com/package/lodash.assign.

lib/utils.js
```diff
@@ -158,6 +160,47 @@ exports.trimUnderscores = function(name) {
 };
 
 /**
+ * Whether or not the given path is relative
+ * @param  {String}  path
+ * @return {Boolean}
+ */
+exports.isRelativePath = function(path) {
```
markelog: How about 

```js
path.resolve( p ) !== path.normalize( p )
```

lib/config/generator.js
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * Script that walks the user through the autoconfig flow
+ * @type {String[]}
+ */
+var prompts = [
+    {
+        name: 'Please choose a preset number:'.green,
+        require: true,
+        pattern: /[0-9]*/
```
markelog:
```js
/[0-9]*/.test("a")
true
```

I expect something like:

```js
/\d+/
```

But we have only 8 presets, need to limit input to that number or handle it ourselves.

markelog: I would even use https://github.com/SBoudrias/Inquirer.js here, its more popular, and has more features and flexibility, like https://github.com/SBoudrias/Inquirer.js#user-content-prompts-type would work much better here

markelog:
1. For `jscs --configure` it gives
```
/Users/arkel/Workspace/jscs/lib/utils.js:169
    var start = path.substring(0, 2);
                     ^
TypeError: Object true has no method 'substring'
    at Object.exports.isRelativePath (/Users/arkel/Workspace/jscs/lib/utils.js:169:22)
    at Object.exports.normalizePath (/Users/arkel/Workspace/jscs/lib/utils.js:196:14)
    at Object.Generator.generate (/Users/arkel/Workspace/jscs/lib/config/generator.js:47:27)
    at module.exports (/Users/arkel/Workspace/jscs/lib/cli.js:104:10)
    at Object.<anonymous> (/Users/arkel/Workspace/jscs/bin/jscs:36:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
```

It seems `No input files specified.` should we displayed instead.

2. Did `node bin/jscs lib test bin publis` from https://github.com/jscs-dev/node-jscs/blob/ccff2178d35c785190eb5f62a75149a4ae6dd5a6/package.json#L112

Get the same error above, whereas `node bin/jscs lib test bin publish -c .jscsrc` works, looks like a blocker

markelog:
I guess comments above is enough for now

lib/utils.js
markelog: Yeah, but it feels like we could use it in a lot of places

lib/config/generator.js
zxqfox: I think we should allow custom presets somehow. It will be useful feature if we allow to reconfigure already existed `.jscsrc` according to new code if rules were not specified in `.jscsrc` but were in preset. Something like auto-nullifier?

markelog: > I think we should allow custom presets somehow.

You mean from the plugin?

> Something like auto-nullifier?

Hm, not sure what you asking, "auto-nullifier" already exist with these changes

zxqfox: > You mean from the plugin?

Yup. I'm not sure if we can look in `node_modules` or `.env.NODE_PATH` for plugins (how to differentiate them from others, by `jscs-` prefix?), but it would be great if we could load `.jscsrc` and read `plugins` or `preset` options, or just provide `--plugins` option to cli to set them explicitly.

Another idea is to have `~/.jscsrc` file and use it as default (with plugins, presets, etc.).

markelog: This question could be deferred, in fact, i would like to do https://github.com/jscs-dev/node-jscs/issues/109 (if that something we're willing to do).

But it seems identification through `jscs-` is our best bet.

markelog:
I can finish this up if @mrjoelkemp is busy, but would love some help with writing the docs.

mrjoelkemp:
@markelog thanks for the offer, but i'll have time tonight and tomorrow night to address some of the comments. Most importantly, i'll fix the blocker (with a test) and write documentation.

markelog:
Awesome!

Will you check out inquirer module?

mrjoelkemp:
It doesn't seem like it'll be a lot of work to switch it to inquirer. I'll
consider it; however, my priority is to unblock the release of 1.11 and do
follow up changes/bugfixes shortly after.

markelog:
Sounds good

bin/jscs
mrjoelkemp: Opting for `--auto-configure`

lib/config/generator.js
mikesherov: Agreed. Simple enough to promisify it yourself manually.

lib/utils.js
mrjoelkemp: Sure. Opting for lodash.assign.

lib/utils.js
mrjoelkemp: It's used in generator and in configuration. The only suitable location aside from utils would be `checker` if we want to avoid a PathHelper.

lib/utils.js
mrjoelkemp: I think the existing solution is much clearer and more performant. Any edge cases that your method accounts for?

lib/config/generator.js
mrjoelkemp: A minor annoyance that I have is rolling my own functions and writing tests for them only to have it suggested that I scrap that effort and use a 3rd party solution – or the reverse (using 3rd party and asked to roll my own). 

Is a way to avoid the deletion of work to keep functions in closed scope and during a review, have other maintainers suggest the extraction and testing of the function if they know of other places where it could be applied?

zxqfox: If you look at this module code: https://github.com/dfilatov/vow-node/blob/master/lib/vow-node.js there are just 2 functions. It's a kind of micromodule as well as lodash.assign.

I'm not against it as soon as we accept lodash.assign. ;-)

lib/config/generator.js
zxqfox: I have a suggestion. Can we use `jscs-jsdoc` plugin a little bit later and fix all docs?

Sure it would be great if these new functions will have some description to decrease well known bus factor.

lib/config/generator.js
mrjoelkemp: That's handled by the caller: cli.

lib/config/generator.js
mrjoelkemp: I think that's a great idea. We should add an issue to do that in 1.12 or 1.later depending on priorities.

lib/utils.js
zxqfox: Windows maybe? Not sure.

mrjoelkemp:
@markelog Addressed a bunch of your issues. This PR is still not ready yet; the outstanding todo items are listed in the PR's description. Will ping you when this is ready for a re-review.

zxqfox:
@mrjoelkemp You've decided to not use vow-node? ;-) That's strange since lodash.assign is here.

mrjoelkemp:
@zxqfox Yeah :) I do agree that lodash.assign has the potential to be more valuable to the codebase than vow-node. I may end up not even using my custom promisify with the switch to Inquirer.js.

zxqfox:
:+1: That's ok. I just still remember that issue where @dfilatov decided to split promisify to separated module to make core codebase lighter for browser usage ;-)

lib/cli.js
```diff
@@ -62,7 +63,7 @@ module.exports = function(program) {
         return returnArgs;
     }
 
-    if (!args.length && process.stdin.isTTY) {
+    if (!args.length && process.stdin.isTTY && !program.autoConfigure) {
```
mikesherov: as @markelog pointed out, this seems wrong. You still need to call autoconfigure on a set of files.

mrjoelkemp:
@markelog I fixed the bug where using `--auto-configure` without a path shows the "No input files specified" message. This still needs a test (as the todo list suggests).

In terms of usage (as the readme states), this was built to be used like `jscs --auto-configure path` where path is a file or directory. I hadn't considered supplying multiple paths. I now see the utility in that for legacy codebases, but is this a blocker? If we change the format for the auto-configure cli option, is that considered "breaking"?

I'm concerned with continuing to block 1.11. As it stands, auto-configure works for a subset of cases. It does not properly handle erroneous input, does not use Inquire,js, does not turn on automatically the way Mike suggested, doesn't work for ES6 codebases, and only supports single input paths (file or directory, but not multiple args).

We should either ship 1.11 without this (realistically, it'll be at *least* another week to get some of the missing things in) or are we fine with shipping a first draft of this and collectively iterating on it thereafter? cc @mikesherov @mdevils @zxqfox

zxqfox:
I would do it because anybody wishes this a lot and it doesn't break anything. But it should be noticed as experimental. And I hope it does not break any current cli usage.

If we strongly need here some patches I can help. Personally I'm missing ES6 :-(

Great job, Joel! :tada:

markelog:
I'd say better ship early, there probably more issues with it that we didn't consider yet. If we would try to make everything perfect, we would never release anything.

> But it should be noticed as experimental

+1

@mrjoelkemp 
> does not turn on automatically

This is 2.0 stuff https://github.com/jscs-dev/node-jscs/issues/986 it would be great to have open tickets for other issues you mentioned.

mikesherov:
Fuck it, we'll do it live!

@mrjoelkemp, please file a new issue with remaining items, and let's land this thing!

mrjoelkemp:
Will do. I'll merge today after one last pass through.

PR #913, opened 2015-01-13
requireSpacesInForStatement: Requires spaces inbetween for statement.

### LINKED ISSUE #912: Rule: require spaces in between for statement
Requires spaces in between for statement.
 
  Type: `Boolean`
 
  Values: `true` to requires spaces in between for statement.
 
  #### Example
 
  ```js
  "requireSpacesInForStatement": true
  ```
 
  ##### Valid
 
  ```js
  for(var i = 0; i<l; i++) {
      x++;
  }
  ```
 
  ##### Invalid
 
  ```js
  for(var i = 0;i<l;i++) {
      x++;
  }
  ```
 
  ```js
  for(var i = 0; i<l;i++) {
      x++;
  }
  ```
 
  ```js
  for(var i = 0;i<l; i++) {
      x++;
  }
  ```

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7700285-rule-require-spaces-in-between-for-statement?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gero3:
Fixes #912

Requires spaces in between for statement.

Type: Boolean

Values: true to requires spaces in between for statement.

#### Example

  "requireSpacesInForStatement": true
##### Valid

  for(var i = 0; i<l; i++) {
      x++;
  }
##### Invalid

  for(var i = 0;i<l;i++) {
      x++;
  }
  for(var i = 0; i<l;i++) {
      x++;
  }
  for(var i = 0;i<l; i++) {
      x++;
  }

### DIFF
--- a/lib/rules/require-spaces-in-for-statement.js
+++ b/lib/rules/require-spaces-in-for-statement.js
@@ -1,0 +1,84 @@
+/**
+ * Requires spaces inbetween for statement.
+ *
+ * Type: `Boolean`
+ *
+ * Values: `true` to requires spaces inbetween for statement.
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireSpacesInForStatement": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * for(var i = 0; i<l; i++) {
+ *     x++;
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * for(var i = 0;i<l;i++) {
+ *     x++;
+ * }
+ * ```
+ *
+ * ```js
+ * for(var i = 0; i<l;i++) {
+ *     x++;
+ * }
+ * ```
+ *
+ * ```js
+ * for(var i = 0;i<l; i++) {
+ *     x++;
+ * }
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(requireSpacesInForStatement) {
+        assert(
+            requireSpacesInForStatement === true,
+            'requireSpacesInForStatement option requires true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireSpacesInForStatement';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('ForStatement', function(node) {
+            var prevToken;
+            if (node.test) {
+                var testToken = file.getTokenByRangeStart(node.test.range[0]);
+                prevToken = file.getPrevToken(testToken);
+                errors.assert.whitespaceBetween({
+                    token: prevToken,
+                    nextToken: testToken,
+                    spaces: 1,
+                    message: 'Missing space after semicolon'
+                });
+            }
+            if (node.update) {
+                var updateToken = file.getTokenByRangeStart(node.update.range[0]);
+                prevToken = file.getPrevToken(updateToken);
+                errors.assert.whitespaceBetween({
+                    token: prevToken,
+                    nextToken: updateToken,
+                    spaces: 1,
+                    message: 'Missing space after semicolon'
+                });
+            }
+        });
+    }
+};
--- a/test/rules/require-spaces-in-for-statement.js
+++ b/test/rules/require-spaces-in-for-statement.js
@@ -1,0 +1,31 @@
+var Checker = require('../../lib/checker');
+var assert = require('assert');
+
+describe('rules/require-spaces-in-for-statement', function() {
+    var checker;
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    describe('true', function() {
+        beforeEach(function() {
+            checker.configure({ requireSpacesInForStatement: true });
+        });
+        it('should report missing spaces in for statement in both cases', function() {
+            assert(checker.checkString('for(i=0;i<l;i++){}').getErrorCount() === 2);
+        });
+        it('should report missing spaces in for statement before test statement', function() {
+            assert(checker.checkString('for(i=0; i<l;i++){}').getErrorCount() === 1);
+        });
+        it('should report missing spaces in for statement behind test statement', function() {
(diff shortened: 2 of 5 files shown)


markelog:
Could add an opposite rule? 

And add this rule to the following presets:  "jquery", "wikimedia", "google", "airbnb" and "crockford".
Each preset changes with different commit.

lib/rules/require-spaces-in-for-statement.js
```diff
@@ -0,0 +1,83 @@
...
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireSpacesInForStatement';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('ForStatement', function(node) {
+            if (node.test) {
```
markelog: I guess you can use assertion API here, should much cleaner
module called "<a href="https://github.com/jscs-dev/node-jscs/blob/8bd2ffd7bd1fc53a49d017f06322b178124534fa/lib/token-assert.js">token-assert</a>",
use example - https://github.com/jscs-dev/node-jscs/blob/8bd2ffd7bd1fc53a49d017f06322b178124534fa/lib/rules/validate-parameter-separator.js#L67-72

test/rules/require-spaces-in-for-statement.js
```diff
@@ -0,0 +1,28 @@
...
+            checker.configure({ requireSpacesInForStatement: true });
+        });
+        it.skip('should report missing spaces in for statement', function() {
+            assert(checker.checkString('for(i=0;i<l;i++){}').getErrorCount() === 2);
+        });
+        it('should report missing spaces in for statement before test statement', function() {
+            assert(checker.checkString('for(i=0; i<l;i++){}').getErrorCount() === 1);
+        });
+        it('should report missing spaces in for statement behind test statement', function() {
+            assert(checker.checkString('for(i=0;i<l; i++){}').getErrorCount() === 1);
```
markelog: Would be cool to check for case without initialization part, like:
```js
for( ; i<l; i++ ){}
```

test/rules/require-spaces-in-for-statement.js
```diff
@@ -0,0 +1,28 @@
...
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    describe('true', function() {
+        beforeEach(function() {
+            checker.configure({ requireSpacesInForStatement: true });
+        });
+        it.skip('should report missing spaces in for statement', function() {
```
markelog: Why?

lib/rules/require-spaces-in-for-statement.js
gero3: updated

test/rules/require-spaces-in-for-statement.js
gero3: no problem

test/rules/require-spaces-in-for-statement.js
gero3: now with extra info

gero3:
Next commit will add the disallow rule.

I also added the rule to the specified presets and to MDCS.

markelog:
> and to MDCS

Didn't find the mentions about it in the mdcs code-style, could you point me to it?

presets/wikimedia.json
```diff
@@ -65,5 +65,6 @@
     "disallowNewlineBeforeBlockStatements": true,
     "validateLineBreaks": "LF",
     "validateQuoteMarks": "'",
-    "validateIndentation": "\t"
+    "validateIndentation": "\t",
+    "requireSpacesInForStatement": true
```
markelog: /cc @Krinkle

test/rules/require-spaces-in-for-statement.js
markelog: Nuh, i meant why do you `skip` that test?

lib/rules/require-spaces-in-for-statement.js
```diff
@@ -0,0 +1,83 @@
...
+                errors.assert.whitespaceBetween({
+                    token: prevToken1,
+                    nextToken: testToken,
+                    spaces: 1,
+                    message: 'Missing space after semicolon'
+                });
+            }
+            if (node.update) {
+                var updateToken = file.getTokenByRangeStart(node.update.range[0]);
+                var prevToken2 = file.getPrevToken(updateToken);
```
markelog: `prevToken2 ` - not a fun of this kind of naming, maybe something more meaningful would be better?

gero3:
https://github.com/mrdoob/three.js/wiki/Mr.doob%27s-Code-Style%E2%84%A2#for 

Altough it is not very clear it does mention the space behind the semicolon.

test/rules/require-spaces-in-for-statement.js
gero3: copy past error

lib/rules/require-spaces-in-for-statement.js
gero3: fixed this

gero3:
Also added the disallow.

presets/wikimedia.json
Krinkle: This seems already covered by `requireSpacesInsideParentheses`. Without that rule, we'd need a rule for `if`, `while`, `switch`, etc. as well. Not just `for`.

markelog: Look at example a bit more:

```js
// this is valid:
for( var i = 0; i<l; i++ )

// this is isn't:
for( var i = 0;i<l;i++ )
```

markelog:
@gero3 could you put preset changes in the different commit? That might be helpful in the future.

presets/wikimedia.json
Krinkle: @markelog Ah, I see. Nice! I still feel like this should maybe be generalised in some way, but sure, feel free to add this to the wikimedia preset.

mikesherov:
LGTM

Issue #646, opened 2014-09-22
Raise an error when a TODO-comment is found


pomeh:
Is it possible to raise an error whenever a single ligne comment like `// TODO: some stuff` is found in the code ? The rule name could be like `disallowTodoInComments`.

This rule could be extended with a more general rule which disallow a set of user-suplied keywords (or even some regex!):

```json
{
    "disallowKeywordInComments": ["TODO", "TASK", "DEBUG", "WHATEVER"]
}
```

What do you think about this ?

joesepi:
I believe I have a new rule for this:
https://github.com/jscs-dev/node-jscs/pull/930

PR #930, opened 2015-01-16
disallowKeywordsInComments: disallow todo and fixme

joesepi:
@mrjoelkemp

### DIFF
--- a/lib/rules/disallow-keywords-in-comments.js
+++ b/lib/rules/disallow-keywords-in-comments.js
@@ -1,0 +1,103 @@
+/**
+ * Disallows keywords in your comments, such as TODO or FIXME
+ *
+ * Type: `Boolean` or `String` or `Array`
+ *
+ * Values:
+ * - `true`
+ * - `'\b(word1|word2)\b'`
+ * - `['word1', 'word2']`
+ *
+ * #### Examples
+ *
+ * ```js
+ * "disallowKeywordsInComments": true
+ * "disallowKeywordsInComments": "\\b(word1|word2)\\b"
+ * "disallowKeywordsInComments": ["word1", "word2"]
+ * ```
+ *
+ * #### Invalid:
+ * ```
+ * // ToDo
+ * //TODO
+ * /** fixme *\/
+ * /**
+ *  * FIXME
+ *  *\/
+ * ```
+ */
+
+var assert = require('assert');
+
+var getCommentErrors = function(comment, keywordRegEx) {
+    var splitComment = comment.value.split('\n');
+    var errors = [];
+
+    splitComment.forEach(function(commentNode, index) {
+        var lineIndex = index;
+        var matches = commentNode.match(keywordRegEx);
+        var lastIndex = 0;
+
+        if (!matches) { return; }
+
+        errors = errors.concat(matches.map(function(match) {
+            var foundAtIndex;
+
+            lastIndex = lastIndex === 0 ? lastIndex : lastIndex + 1;
+            foundAtIndex = commentNode.indexOf(match, lastIndex);
+            lastIndex = foundAtIndex;
+
+            // line + lineIndex because comment block was split at new lines
+            //   will place carat at correct place within multiline comment
+            // foundAtIndex += 2 because comment opening is stripped
+            //   +2 finds accurate carat position on opening line comment
+            return {
+                line: comment.loc.start.line + lineIndex,
+                column: lineIndex <= 0 ? foundAtIndex += 2 : foundAtIndex
+            };
+        }));
+    });
+
+    return errors;
+};
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(disallowKeywordsInComments) {
+        this._message = 'Comments cannot contain the following keywords: ';
+        this._keywords = ['todo', 'fixme'];
+
+        switch (true) {
+            case Array.isArray(disallowKeywordsInComments):
+                // use the array of strings provided to build RegExp pattern
+                this._keywords = disallowKeywordsInComments;
+                /* falls through */
+            case disallowKeywordsInComments:
+                // use default keywords
+                this._message += this._keywords.join(', ');
+                this._keywordRegEx = new RegExp('\\b(' + this._keywords.join('|') + ')\\b', 'gi');
+                break;
+            case typeof disallowKeywordsInComments === 'string':
+                // use string passed in as the RegExp pattern
+                this._message = 'Comments cannot contain keywords based on the expression you provided';
+                this._keywordRegEx = new RegExp(disallowKeywordsInComments, 'gi');
+                break;
+            default:
+                assert(false, 'disallowKeywordsInComments option requires a value of true, a string or an array');
+        }
+    },
+
+    getOptionName: function() {
+        return 'disallowKeywordsInComments';
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            if (comment.value.search(this._keywordRegEx) === -1) { return; }
+            getCommentErrors(comment, this._keywordRegEx).forEach(function(errorObj) {
+                errors.add(this._message, errorObj);
+            }.bind(this));
+        }.bind(this));
+    }
+  };
--- a/test/rules/disallow-keywords-in-comments.js
+++ b/test/rules/disallow-keywords-in-comments.js
@@ -1,0 +1,49 @@
+var Checker = require('../../lib/checker');
+var assert = require('assert');
+
+describe('rules/disallow-keywords-in-comments', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+        checker.configure({ disallowKeywordsInComments: true });
+    });
+
+    it('should report on an instance of the word todo regardless of case', function() {
+        assert(checker.checkString('// todo').getErrorCount() === 1);
+        assert(checker.checkString('// TODO').getErrorCount() === 1);
+        assert(checker.checkString('/** ToDo */').getErrorCount() === 1);
+        assert(checker.checkString('/**\n * TODO\n */').getErrorCount() === 1);
+        assert(checker.checkString('/* todo */').getErrorCount() === 1);
+        assert(checker.checkString('/*\n\n\n\n    TODO\n */').getErrorCount() === 1);
+    });
+
+    it('should report on an instance of the word fixme regardless of case', function() {
+        assert(checker.checkString('//fixme').getErrorCount() === 1);
+        assert(checker.checkString('// FIXME').getErrorCount() === 1);
+        assert(checker.checkString('/** FixMe */').getErrorCount() === 1);
+        assert(checker.checkString('/**\n * FIXME\n */').getErrorCount() === 1);
+        assert(checker.checkString('/* fixme */').getErrorCount() === 1);
+        assert(checker.checkString('/*\n\n\n\n     FIXME\n */').getErrorCount() === 1);
+    });
+
+    it('should report multiple errors per line', function() {
+        assert(checker.checkString('/* FIXME and ToDO */').getErrorCount() === 2);
+        assert(checker.checkString('//Hi. Todo then fixME').getErrorCount() === 2);
+    });
+
+    it('should not report on a valid comment without the keywords', function() {
+        assert(checker.checkString('//valid').isEmpty());
+        assert(checker.checkString('// valid').isEmpty());
+        assert(checker.checkString('/** valid */').isEmpty());
+        assert(checker.checkString('/*\n\n\n\n     Totally valid\n */').isEmpty());
+    });
+
+    it('should not report on a keywords within words', function() {
+        assert(checker.checkString('//ffixme').isEmpty());
+        assert(checker.checkString('// todos').isEmpty());
+        assert(checker.checkString('/** plzfixme*/').isEmpty());
+        assert(checker.checkString('/*\n\n\n\n     todont\n */').isEmpty());
+    });
+});
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -619,6 +619,7 @@ Configuration.prototype.registerDefaultRules = function() {
 
     this.registerRule(require('../rules/require-spaces-in-for-statement'));
     this.registerRule(require('../rules/disallow-spaces-in-for-statement'));
+    this.registerRule(require('../rules/disallow-keywords-in-comments'));
 };
 
 /**


joesepi:
I did a couple `commit --amend`s after an initial style check fail. That's part of the reason why you see two coverall reports.

joesepi:
This fix is for this feature request:
https://github.com/jscs-dev/node-jscs/issues/646

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,69 @@
...
+ * //TODO
+ * /** fixme *\/
+ * /**
+ *  * FIXME
+ *  *\/
+ * ```
+ */
+
+var assert = require('assert');
+
```
mikesherov: Name this variable `keywordRegex`. More descriptive.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,69 @@
...
+                if (matchIndex >= 0) {
+                    // offset reporter column indicator
+                    // because parser strips the chars that begin comments (//, /*)
+                    if (index === 0) {
+                        matchIndex += 2;
+                    }
+
+                    // adjust line location for reporter
+                    // comment is split by new line and iterated over
+                    errorLoc.line = comment.loc.start.line + index;
```
mikesherov: Inline this object directly into the error.add call. No need to define as empty object first, nor even assign to a var.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,69 @@
...
+
+var r = /\b(todo|fixme)\b/i;
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(disallowKeywordsInComments) {
+        assert(
+            disallowKeywordsInComments === true,
+            'disallowKeywordsInComments option disallows certain keywords in your comments'
```
mikesherov: This error message should say what the required value is. See other rules for suggestions.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            var errorLoc = {};
+            // split by new line to get accurate line number reporting
+            var splitComment = comment.value.split('\n');
+
+            splitComment.forEach(function(commentNode, index) {
+                var matchIndex = commentNode.search(r);
```
mikesherov: What if the word TODO appears twice in the same line? You'll only report one error...

Which I think is fine, but it highlights the fact that getting the exact location of the offending word is probably not important. You could simplify this rule greatly if you just reported the start of the comment as the error location.

mikesherov:
@joesepi, thanks for contributing, friend!

Good work. Found a few nits to address, but also:

 * commit message should say which issue it closes: 

```
disallowKeywordsInComments: new rule 

Fixes #646
```

 * the rule should take in `true` or a string that represents a regex to use instead of the default `"/Joel/"`.

Thanks again!

joesepi:
@mrjoelkemp and @mikesherov and whoever else is interested. I updated the rule and test. I moved the bit where the comment is broken up by lines (for better carat placement) out of the check itself. My intent, after discussing it with Joel earlier, is to move it to a comment util or somewhere similar. I thought this would be a good point to get feedback and if things are good, I can pull it out entirely. Thanks!!

mikesherov:
@joesepi you have a few style violations in your code, which is why travis is failing. Running `npm test` locally will reveal the erros so you don't have to wait for travis.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+var keywords = ['todo', 'fixme'];
+
+var keywordRegEx;
+var buildKeywordRegEx = function(suppliedPattern) {
+    var pattern;
+
+    if (!suppliedPattern) {
+        pattern = '\\b(';
+        pattern += keywords.join('|');
+        pattern += ')\\b';
```
mrjoelkemp: Why not just munge all together on one line? 

I'll take care of that post-merge.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(disallowKeywordsInComments) {
+        // use default keywords
+        if (disallowKeywordsInComments === true) {
+            message = buildMessage();
```
mrjoelkemp: We typically store data on the rule instance to avoid shared state across instances: `this._message = buildMessage()`.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+    var splitComment = comment.value.split('\n');
+    var errorArray = [];
+
+    splitComment.forEach(function(commentNode, index) {
+        var lineIndex = index;
+        var matches = commentNode.match(keywordRegEx);
+        var lastIndex = 0;
+
+        if (matches) {
+            matches.map(function(match) {
```
mrjoelkemp: Alternative
```js
errorArray = errorArray.concat(matches.map(function(match) {
   ... 
   return {
     line: ...,
     column: ...
   };
}))
```

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+
+var message;
+var buildMessage = function() {
+    var keywordsList = keywords.join(', ');
+    return 'Comments cannot contain the following keywords: ' + keywordsList;
+};
+
+var getCommentErrors = function(comment) {
+    var splitComment = comment.value.split('\n');
+    var errorArray = [];
```
mrjoelkemp: `errors` implies it's an array

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+
+    splitComment.forEach(function(commentNode, index) {
+        var lineIndex = index;
+        var matches = commentNode.match(keywordRegEx);
+        var lastIndex = 0;
+
+        if (matches) {
```
mrjoelkemp: `if (!matches) { return; }`

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+    getOptionName: function() {
+        return 'disallowKeywordsInComments';
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            var errorCheck = comment.value.search(keywordRegEx);
+
+            if (errorCheck >= 0) {
+                var errorArray = getCommentErrors(comment);
```
mrjoelkemp: Consider removing the temp var:

```js
getCommentErrors(comment)
.forEach(...)
```

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,131 @@
...
+
+    getOptionName: function() {
+        return 'disallowKeywordsInComments';
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            var errorCheck = comment.value.search(keywordRegEx);
+
+            if (errorCheck >= 0) {
```
mrjoelkemp: Consider introducing a better explaining variable like: `hasViolations = comment.value.search(keywordRegEx) !== -1`

lib/rules/disallow-keywords-in-comments.js
zxqfox: I want this as rule! ;-D

joesepi:
@mikesherov and @mrjoelkemp and others: all good comments. Thank you. I have updated the code and ran npm test locally so Travis shouldn't complain. Please recheck and comment if need be.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+ * Values:
+ * - `true`
+ * - `'\b(word1|word2)\b'`
+ * - `['word1', 'word2']`
+ *
+ * #### Examples
+ *
+ * ```js
+ * "disallowKeywordsInComments": true
+ * "disallowKeywordsInComments": '\b(word1|word2)\b'
```
zxqfox: I'm not sure, but feels like it should be `double-slash b` (`"\\b(...)\\b"`) and in double-quotes ('cause of JSON). Right?

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    return errors;
+};
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(disallowKeywordsInComments) {
+        if (disallowKeywordsInComments === true) {
```
zxqfox: Suggest switch-case here:
```js
switch (true) {
  case disallowKeywordsInComments === true:
  ...
  default:
    assert(false, ...
}
```

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+var assert = require('assert');
+
+// default keywords
+var keywords = ['todo', 'fixme'];
+
+var keywordRegEx;
+var buildKeywordRegEx = function(suppliedPattern) {
+    var pattern;
+
+    if (!suppliedPattern) {
```
zxqfox: mmm... maybe:
```js
pattern = suppliedPattern || ('\\b(' + keywords.join('|') + ')\\b');

File: lib/rules/disallow-keywords-in-comments.js (hunk 12 above)
Comment (mikesherov) [inline lib/rules/disallow-keywords-in-comments.js]: yes. that's right.

File: lib/rules/disallow-keywords-in-comments.js (hunk 14 above)
Comment (mikesherov) [inline lib/rules/disallow-keywords-in-comments.js]: agree with @zxqfox

File: lib/rules/disallow-keywords-in-comments.js [hunk 15]
```diff
@@ -0,0 +1,119 @@
...
+ * ```
+ */
+
+var assert = require('assert');
+
+// default keywords
+var keywords = ['todo', 'fixme'];
+
+var keywordRegEx;
+var buildKeywordRegEx = function(suppliedPattern) {
```
mikesherov: why var not a function declaration?

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+    if (!suppliedPattern) {
+        pattern = '\\b(' + keywords.join('|') + ')\\b';
+    } else {
+        pattern = suppliedPattern;
+    }
+
+    return new RegExp(pattern, 'gi');
+};
+
+var buildMessage = function() {
```
mikesherov: why not function declaration?

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+    var keywordsList = keywords.join(', ');
+    return 'Comments cannot contain the following keywords: ' + keywordsList;
+};
+
+var getCommentErrors = function(comment) {
```
mikesherov: why not function declaration?

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+ *  *\/
+ * ```
+ */
+
+var assert = require('assert');
+
+// default keywords
+var keywords = ['todo', 'fixme'];
+
+var keywordRegEx;
```
mikesherov: this should a member property, not a global variable.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+
+    getOptionName: function() {
+        return 'disallowKeywordsInComments';
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            if (comment.value.search(keywordRegEx) === -1) { return; }
+            getCommentErrors(comment).forEach(function(errorObj) {
+                errors.add(this._message, errorObj);
```
mikesherov: `this` is the wrong thing here... need to bind up.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+            });
```
mikesherov: `}, this);`

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,119 @@
...
+        });
```
mikesherov: `}, this);`

lib/rules/disallow-keywords-in-comments.js
mikesherov: especially considering leveraging the fall through:

```
witch (true) {
  case Array.isArray(disallowKeywordsInComments):
    keywords = disallowKeywordsInComments;
    /*falls through*/
  case disallowKeywordsInComments === true:
    this._message = buildMessage();
    keywordRegEx = buildKeywordRegEx();
    break;
  ...
  default:
    assert(false, ...
}

**Comment by joesepi** [comment]
You guys, these are all great comments. Thank you! The code has improved greatly. Check it out!

@mrjoelkemp @mikesherov @zxqfox

File: lib/rules/disallow-keywords-in-comments.js [hunk 22]
```diff
@@ -0,0 +1,103 @@
...
+        }
+    },
+
+    getOptionName: function() {
+        return 'disallowKeywordsInComments';
+    },
+
+    check: function(file, errors) {
+        file.getComments().forEach(function(comment) {
+            if (comment.value.search(this._keywordRegEx) === -1) { return; }
```
mikesherov: You should do this check inside of the `getCommentErrors` function so your not `search`ing and then `matching`. Just `match` and there are zero matches, return empty array.

test/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,49 @@
+var Checker = require('../../lib/checker');
+var assert = require('assert');
+
+describe('rules/disallow-keywords-in-comments', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+        checker.configure({ disallowKeywordsInComments: true });
```
mikesherov: Do one `describe` per option value. Right now you are only testing `true`. Likely why coverage is listed as going down.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+ * /** fixme *\/
+ * /**
+ *  * FIXME
+ *  *\/
+ * ```
+ */
+
+var assert = require('assert');
+
+var getCommentErrors = function(comment, keywordRegEx) {
```
mikesherov: function declaration instead of assigning to var here please.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+    var splitComment = comment.value.split('\n');
+    var errors = [];
+
+    splitComment.forEach(function(commentNode, index) {
+        var lineIndex = index;
```
mikesherov: Why create this variable? Just name index lineIndex in the function params.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+        var matches = commentNode.match(keywordRegEx);
+        var lastIndex = 0;
+
+        if (!matches) { return; }
+
+        errors = errors.concat(matches.map(function(match) {
+            var foundAtIndex;
+
+            lastIndex = lastIndex === 0 ? lastIndex : lastIndex + 1;
```
mikesherov: If you initialize `lastIndex` to `-1` instead of `0`, this whole line is just `lastIndex++`.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+            foundAtIndex = commentNode.indexOf(match, lastIndex);
```
mikesherov: No need for `foundAtIndex`, just assign directly to `lastIndex` and use it everywhere you used `foundAtIndex` below.

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+            lastIndex = foundAtIndex;
+
+            // line + lineIndex because comment block was split at new lines
+            //   will place carat at correct place within multiline comment
+            // foundAtIndex += 2 because comment opening is stripped
+            //   +2 finds accurate carat position on opening line comment
+            return {
+                line: comment.loc.start.line + lineIndex,
+                column: lineIndex <= 0 ? foundAtIndex += 2 : foundAtIndex
```
mikesherov: `<=` -> `===`

lib/rules/disallow-keywords-in-comments.js
```diff
@@ -0,0 +1,103 @@
...
+    configure: function(disallowKeywordsInComments) {
+        this._message = 'Comments cannot contain the following keywords: ';
+        this._keywords = ['todo', 'fixme'];
+
+        switch (true) {
+            case Array.isArray(disallowKeywordsInComments):
+                // use the array of strings provided to build RegExp pattern
+                this._keywords = disallowKeywordsInComments;
+                /* falls through */
+            case disallowKeywordsInComments:
```
mikesherov: Make sure to test the string option, because I think it'll never get hit.

mikesherov:
*ping* @joesepi

joesepi:
Sorry folks. I wanted to wrap this up before I went on vacation but it didn't happen. I will try to finish it in the coming days.

mikesherov:
Landed with the fixes mentioned! Thanks again for the hard work here @joesepi !

PR #935, opened 2015-01-19
Preset: change value of "requireDotNotation" rule for jquery preset

markelog:


mikesherov:
Where does this come from? I don't doubt the validity, just curios.

markelog:
> Where does this come from?

Just a thought, same use-case as with https://github.com/jscs-dev/node-jscs/issues/866

mikesherov:
@markelog LGTM

mikesherov:
Landing.

PR #937, opened 2015-01-20
 CI: Configure AppVeyor 

am11:
This PR:

* adds appveyor.yml to tests on builds on Windows.
* adds AppVeyor badge to repo's readme.
* updates dependencies' versions.

It is passing all the tests: https://ci.appveyor.com/project/am11/node-jscs/build/6.

<del>Previously, for some crazy reason, it [was failing](https://ci.appveyor.com/project/am11/node-jscs/build/4/job/mk162lsx62x2u3d2) with node version 0.11.14, so I removed [version matrix](https://github.com/am11/node-jscs/blob/2154d7892af388b6a70e8f66bde008ddbdfec397/appveyor.yml#L6-L9) from the configuration which renders it test only on latest stable version of node (v0.10.35). On TravisCI, jscs is also tested against node v0.10.35, so I presume the issue is common (probably due to some dependency?).</del>
See https://github.com/jscs-dev/node-jscs/pull/937#issuecomment-72159148.

**Next steps:**

* After merging this, the repo owner (or collaborator under jscs-dev organisation) would need to logon once on Appveyor with GitHub credentials.
* On account selection screen choose `jscs-dev` from dropdown list (label: Account) and press the GitHub button again.
* On landing page click on Add Project and select node-jscs to keep the ball rolling.

:tada:


markelog:
-1 from me, one CI is enough, i think it's fair to say we depend on other tools for jscs to be cross-platform.

am11:
Alright. jshint repo has it, I thought it might be a good idea to configure appveyor for jscs as well (especially for curious contributors, who are not using Windows machine).

markelog:
@mdevils what is your take on this?

mikesherov:
FWIW, I don't think it can really hurt.

mikesherov:
ping @mdevils

mdevils:
+1 from me.

We had issues with windows and several times we broke windows.

mdevils:
In this example tests were not run, just `linting` and `style checking` were included. Can you please enable tests?

am11:
@mdevils I have enabled the tests. Here are some notes:

* We have now three matrices: v0.10, v0.11 and io.js v1.0 based on http://www.appveyor.com/docs/lang/nodejs-iojs.
* Node.js v0.10 has an issue with piping on Windows which causes the process to terminate as soon as there is some delay in test. This was fixed in v0.11. Meanwhile mocha also provides a workaround for that: https://github.com/mochajs/mocha/issues/333#issuecomment-70276442 (which is unreleased as of this moment).
* Locally [these five tests](https://gist.github.com/am11/7381047e5e17496679e3) are failing on Windows.

Meanwhile I have marked `allow_faliure` for v0.10 branch until mocha next build is release. For rest of matrix (njs v0.11 and io.js v1.0), IMO those tests should be fixed before removing those versions under `allow_faliure` block. Currently, all three of them are allowed to fail.

https://ci.appveyor.com/project/am11/node-jscs/build/20

@dougwilson, @FeodorFitsner, is it normal to get `Command exited with code 4` (for instance in [io.js turn](https://ci.appveyor.com/project/am11/node-jscs/build/20/job/ks42e87x4bimculi) also njs v0.11)? AFAICT, after running all the tests, it should exit with code 0 since the process has completed successfully (although some tests failed). Please CMIIW.

dougwilson:
The `Command exited with code 4` is because there are failing tests. Because mocha tries to print all the failures at the end and then immediately exit, all the messages don't end up making it to the STDIO before Node.js exits (pending a new mocha release eventually). A work-around to at least see them is to add the `--no-exit` flag on `mocha` (assuming your test suite is not leaving something holding open the event loop).

Another option is to use the `spec` reporter. The error output at the end will still get cut off, but you'll be able to tell which tests failed based on if there is a check mark or it still has the number in front of the test name.

dougwilson:
Oddly I didn't have any local test failures on Windows, though @am11 did.

dougwilson:
But, with, for example, `express-generator`, I had a hard time getting that one on AppVeyor because the tests had always passed locally, but yet there were some failures on AppVeyor (due to it reading from the STDIO slowly). I think the best next step is to change the test command to `node_modules/.bin/mocha --no-exit` and see if it shows anything more (until there is a new `mocha` release).

am11:
@dougwilson, thanks for the insights. With `--no-exit` at least v0.10 target is executing all the tests. But now all three targets are reporting `Command exited with code 4`. See https://ci.appveyor.com/project/am11/node-jscs/build/21.

Also, notice that the tests which are failing locally for me are different than those are failing on AppVeyor CI (but all three targets on CI are failing exact same four tests). Hence the weirdness.

dougwilson:
> With --no-exit at least v0.10 target is executing all the tests. But now all three targets are reporting Command exited with code 4

The only difference with `--no-exit` wasn't that it would fix the failures, only that it would now print what they were :) Which I see is happening now. It was actually executing all the tests before, but the output was getting cut off because of the <= 0.10 Node.js bug.

Now that I can see what tests are failing on AppVeyor, I may have time to look at the tests in this project and see why they fail.

dougwilson:
So the reason the tests fail with text vs console is because your tests are fragile: they assume that this line works a certain way: https://github.com/jscs-dev/node-jscs/blob/master/lib/cli-config.js#L153

When you run the tests on your local machine, it's in an environment that is detected as supporting colors; on AppVeyor it is correctly detected as an environment that does not support colors. Those tests need to be fixed so they are either not testing something influenced by the environment, duplicate the environment check to know what to expect, or control the environment completely.

am11:
The root cause is the set of conditions in `supports-color` package.

@sindresorhus, would it make sense to have another rule for CI before [this line] (https://github.com/sindresorhus/supports-color/blob/2ae39320b7ca26c2803b4da531ed9bbcc40d78a3/index.js#L18)? Something like:

```javascript
if ('CI' in process.env) { 
  return true; 
}
```

Both AppVeyor and Travis CI define `CI` env var.

AppVeyor: http://www.appveyor.com/docs/environment-variables
Travis CI: http://docs.travis-ci.com/user/ci-environment/#Environment-variables

mikesherov:
I think @dougwilson gave the correct answer: modify the tests so that the tests themselves detect color support and adjust expectations based on that.

Sure @sindresohrus can add more checks to the package, but still our tests aren't accounting for environment properly.

am11:
@mikesherov, should the test modification be part of this PR?

mikesherov:
Yes please

dougwilson:
There are two ways to support this.

1) Make the environment adjust to the tests. I have a similar problem testing my `depd` module. This is what I did there:
https://github.com/dougwilson/nodejs-depd/blob/v1.0.0/test/test.js#L657-L675

2) Make the tests adjust to the environment:
```patch
From a1c4d967b24537dfd0aa540e8bb91c7194d6b12e Mon Sep 17 00:00:00 2001
From: Douglas Christopher Wilson <doug@somethingdoug.com>
Date: Sat, 31 Jan 2015 13:11:47 -0500
Subject: [PATCH] Example change for color tests

---
 test/cli-config.js | 16 ++++++++++++----
 test/cli.js        | 16 ++++++++++++----
 2 files changed, 24 insertions(+), 8 deletions(-)

diff --git a/test/cli-config.js b/test/cli-config.js
index 4275701..70ce76a 100644
--- a/test/cli-config.js
+++ b/test/cli-config.js
@@ -2,6 +2,7 @@ var path = require('path');
 
 var assert = require('assert');
 var sinon = require('sinon');
+var supportsColor = require('supports-color');
 var rewire = require('rewire');
 
 var configFile = rewire('../lib/cli-config');
@@ -133,18 +134,25 @@ describe('modules/cli-config', function() {
     });
 
     describe('getReporter method', function() {
-        it('should get console reporter if called without arguments', function() {
+        var basicReporter;
+
+        before(function() {
+          // The basic reporter varies on environment.
+          basicReporter = supportsColor ? 'console' : 'text';
+        });
+
+        it('should get a basic reporter if called without arguments', function() {
             var reporter = configFile.getReporter();
 
             assert.equal(typeof reporter.writer, 'function');
-            assert.equal(reporter.path, path.resolve('./lib/reporters/console'));
+            assert.equal(reporter.path, path.resolve('./lib/reporters/', basicReporter));
         });
 
-        it('should get console reporter if called with arguments with default values', function() {
+        it('should get a basic reporter if called with arguments with default values', function() {
             var reporter = configFile.getReporter(undefined, true);
 
             assert.equal(typeof reporter.writer, 'function');
-            assert.equal(reporter.path, path.resolve('./lib/reporters/console'));
+            assert.equal(reporter.path, path.resolve('./lib/reporters/', basicReporter));
         });
 
         it('should get text reporter if called with colors = false', function() {
diff --git a/test/cli.js b/test/cli.js
index 314e366..56e0f45 100644
--- a/test/cli.js
+++ b/test/cli.js
@@ -6,6 +6,7 @@ var assert = require('assert');
 var glob = require('glob');
 var hasAnsi = require('has-ansi');
 var rewire = require('rewire');
+var supportsColor = require('supports-color');
 
 var cli = rewire('../lib/cli');
 var startingDir = process.cwd();
@@ -317,7 +318,14 @@ describe('modules/cli', function() {
     });
 
     describe('reporter option', function() {
-        it('should implicitly set console reporter', function() {
+        var basicReporter;
+
+        before(function() {
+          // The basic reporter varies on environment.
+          basicReporter = supportsColor ? 'console' : 'text';
+        });
+
+        it('should implicitly set a basic reporter', function() {
             var result = cli({
                 args: ['test/data/cli/error.js'],
                 colors: true,
@@ -325,7 +333,7 @@ describe('modules/cli', function() {
             });
 
             return result.promise.always(function() {
-                assert.equal(path.basename(result.reporter), 'console');
+                assert.equal(path.basename(result.reporter), basicReporter);
                 rAfter();
             });
         });
@@ -492,7 +500,7 @@ describe('modules/cli', function() {
             });
         });
 
-        it('should have colors output', function() {
+        it('should ' + (supportsColor ? '' : 'not ') + 'have colors output', function() {
             var result = cli({
                 colors: true,
                 args: ['test/data/cli/error.js'],
@@ -500,7 +508,7 @@ describe('modules/cli', function() {
             });
 
             return result.promise.fail(function() {
-                assert(hasAnsi(console.log.getCall(0).args[0]));
+                assert(hasAnsi(console.log.getCall(0).args[0]) === supportsColor);
             });
         });
     });
-- 
1.9.5.msysgit.0
```

am11:
@dougwilson thanks. But ICYMI, without changing the test spec, there is a simple way to fix it by passing `--color` argument.
All tests are passing for the entire matrix:
https://ci.appveyor.com/project/am11/node-jscs/build/22/

dougwilson:
Oh, nice, I didn't know mocha had a `--colors` argument and that it would force it on the tests. If it works, I agree it's the easiest change.

markelog:
It seems hassle for nothing, we didn't yet receive at least one ticket of windows failures and now we need to change a bunch of things for no apparent reason.

dougwilson:
Sure, though the colors issue is in no way Windows-specific. Do the tests not fail on linux if you run `npm test > results.log`?

am11:
Actually the arguments are also parsed by [`supports-color` code](https://github.com/sindresorhus/supports-color/blob/2ae39320b7ca26c2803b4da531ed9bbcc40d78a3/index.js#L11) and the good thing is mocha just ignores the arguments which it does not understand (as opposed to throwing error like: *invalid argument*). 

I have removed the [`allow_faliure` block](https://github.com/am11/node-jscs/blob/8ebead8dcc0ab49ba9ba511f27a70d750e1bc842/appveyor.yml#L14-L20) from latest commit.

@mikesherov, @mdevils, this PR is ready. After merging, please follow the *Next Steps* from [aforementioned comment](https://github.com/jscs-dev/node-jscs/pull/937#issue-54850678) to set it up.

am11:
@markelog I have made zero changes in the code.

markelog:
Meant this comment - https://github.com/jscs-dev/node-jscs/pull/937#issuecomment-72328916

> Actually the arguments are also parsed by supports-color code and the good thing is mocha just ignores the arguments which it does not understand (as opposed to throwing error like: invalid argument).

This is unsettling, it should work without any hacks.

> - node --version
  - npm --version
  - npm install
  - node_modules/.bin/jshint .
  - node bin/jscs lib test bin publish
  - ./node_modules/.bin/mocha --no-exit --color

It should do only `npm test` and that's it.

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+init:
+  - git config --global core.autocrlf input
+
+environment:
+  matrix:
+    # node.js
+    - nodejs_version: "0.10"
+    - nodejs_version: "0.11"
+    # io.js
+    - nodejs_version: "1.0"
```
markelog: Only production-ready envs, i.e. 0.10

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+
+test_script:
+  - node --version
+  - npm --version
+  - npm install
```
zxqfox: I think this line should be moved to install section (because it's about installation ;-)

zxqfox:
> It should do only npm test and that's it.

I think it should do `npm run mocha` or `npm run appveyor` with just `mocha`. We don't need to test codestyle several times on different platforms. We just need to test jscs itself therefore run `mocha` should be sufficient.

I'm for `npm run appveyor` or `npm run mocha`.

am11:
@markelog, 

1. `npm test` does not work when `jshint` is not globally installed. 
2. && works in cmd on Windows, but not in PS (you need to use `;`). Similarly both forward and back slashes are acceptable in PS but cmd only allows backslashes. But this should not be the problem on AppVeyor, since it has mingw's msys in its path for both cmd and PowerShell.

Which also means `npm test` for node-jscs does not work in Windows box in its current form unless you already have jshint (which should be the problem in Linux too, where jshint is not already installed) and Cygwin/Mingw installed. 

To fix it, either we need to change [this](https://github.com/jscs-dev/node-jscs/blob/df7474c1c3adc8ee501b82097f9e7a43f8c519a0/package.json#L112):

```json
"lint": "jshint . && node bin/jscs lib test bin publish",
```

to:

```json
"lint": "./node_modules/.bin/jshint . && node bin/jscs lib test bin publish",
```

Or explicitly install jshint globally in preinstall script.

Or keep the concerns separated as AppVeyor has a separate config file which we have now: run mocha on repo's root. And this is the only way to run tests on Windows anyway.

am11:
@zxqfox 
> btw. we can get rid of  --no-exit  and just left  --color  option

v0.10 has a bug which was fixed in v0.11 and mocha vNext provides a fix for v0.10 (which is unreleased): https://github.com/jscs-dev/node-jscs/pull/937#issuecomment-72208342. That is why we cannot get rid of --no-exit yet.

markelog:
@zxqfox 
> I think it should do npm run mocha or npm run appveyor with just mocha

Yeah, this is what i call "hassle"

> We don't need to test codestyle several times on different platforms.

Don't want to know about :-(, i would let it execute as it does on travis.

@am11 
> npm test does not work when jshint is not globally installed.

It does, this why we have it as dev dep, user doesn't required to install anything globally, see npm docs for more info.

appveyor.yml
am11: Is io.js not production ready?

markelog: Fun, huh :-)? it is 1.0.4, but its not ready to be used, see - https://iojs.org/faq.html#version

@mikeal do you have a ticket about that? It seems it confuses a lot of people.

am11:
@markelog 
npm test does not work when jshint is not globally installed *on Windows*, even with dev dependencies.

zxqfox:
@markelog Just travis also pushes coverage and it looks like overhead.

@am11 Did you tried to set up Path manually? I've cloned repo onto windows machine, run PS, run `npm i`, and run `npm test` — all is fine.

After that I did one more test:
```
$env:Path = $env:Path + ";./node_modules/.bin"
mocha
```
And it also works well.

Also I've tried to remove global jshint — but `npm test` still works.

I can't repeat that ;-( Sorry.
Can you try to add `- ps: $env:Path = $env:Path + ";./node_modules/.bin"` after installation or before `npm test`? Also would be useful to have `- ps: echo $env:Path` before `npm test`.

zxqfox:
@am11 Also I would try this syntax:
```
  - cmd: npm test
```
instead of
```
  - npm test
```

markelog:
@am11 btw, thank you for the all work you doing here, i'm skeptical, but we're all appreciative

dougwilson:
Though, as a user who uses jscs on Windows all the time, I appreciate the effort :)

am11:
@zxqfox, 

Thanks for the env var tip, `npm test` is working now. Also, I have removed `--no-exit` and it is working as well (I don't know how and why)! :stuck_out_tongue_winking_eye: 

The default runtime in AppVeyor is cmd, so let it follow the natural course without explicitly specifying it. Moreover, there is a separate task in package.json `test-appveyor`.

@markelog, I have removed v0.11 and io.js from the test matrix, for those being non-production-ready. Although node.js community is [very touchy](https://github.com/sass/node-sass/issues/563#issuecomment-70072413) about this particular subject: *regardless of how Joyent chooses to label that branch as "unstable"*. :smiley:

package.json
```diff
@@ -111,6 +111,7 @@
   "scripts": {
     "lint": "jshint . && node bin/jscs lib test bin publish",
     "test": "npm run lint && mocha",
+    "test-appveyor": "mocha --color",
```
markelog: There is a lot of comments here now, so i lost track of why this is needed, i'd like to fully understand why it is required to have `--color` pseudo-option in appveyor but not in travis.

am11: Copy https://github.com/jscs-dev/node-jscs/pull/937#issuecomment-72307626.

`supports-color` code has this [`isTTY` condition](https://github.com/sindresorhus/supports-color/blob/2ae39320b7ca26c2803b4da531ed9bbcc40d78a3/index.js#L18), which satisfies for AppVeyor as teletyping is not supported in AppVeyor. On Travis CI, it does not satisfies this condition as stdout has `isTTY` set to `true`.

I made this [test commit](https://github.com/am11/node-jscs/commit/bc518d41e1729f02aa4a30aa0ea27383d52f771c) to verify it. [appveyor result](https://ci.appveyor.com/project/am11/node-jscs/build/33#L148) and here is [travis-ci result](https://travis-ci.org/jscs-dev/node-jscs/builds/49057659#L94). I will rebase it out now.

markelog: Could you do the

```
"test": "npm run lint && mocha --color"
```
and make appveyor execute `npm test`

package.json
```diff
@@ -62,30 +62,30 @@
...
-    "has-ansi": "~1.0.0",
+    "coveralls": "~2.11.2",
+    "has-ansi": "~1.0.1",
     "jshint": "~2.6.0",
     "mocha": "~2.1.0",
     "regenerate": "~1.2.1",
-    "rewire": "~2.1.0",
-    "sinon": "~1.12.0",
+    "rewire": "~2.1.4",
+    "sinon": "~1.12.2",
```
markelog: Is dependencies update required?

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: "{build}"
+
+init:
+  - git config --global core.autocrlf input
+  - ps: $env:Path = $env:Path + ";./node_modules/.bin"
+
+environment:
+  matrix:
+    # node.js
```
markelog: I guess you could remove it now

appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+    - nodejs_version: "0.10"
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - node --version
+  - npm --version
+  - npm install
+
+test_script:
+  - ps: echo $env:Path
```
markelog: Is this required?

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: "{build}"
+
+init:
+  - git config --global core.autocrlf input
+  - ps: $env:Path = $env:Path + ";./node_modules/.bin"
```
markelog: If it could works without it then i guess we could remove it

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+version: "{build}"
+
+init:
+  - git config --global core.autocrlf input
```
markelog: Why do we need that?

package.json
am11: Nops, just did it in a separate commit. All is working fine though.

appveyor.yml
am11: Just output the path as per @zxqfox's suggestion.

package.json
markelog: Alrighty, it's fine, could merge it with it

appveyor.yml
dougwilson: Because git will check out text files with CRLF line endings by default. he code style checker that is part of the tests here will fail since the line endings won't be LF.

markelog: Isn't `.gitattributes` not enough?

am11: Nops. Removed.
Also removed the environment matrix below. The build is successful. https://ci.appveyor.com/project/am11/node-jscs/build/38 with latest commit: ee2bb1dc5cfaf6f46f55a81f84070b20ef6dfbb7. :tada:

README.md
```diff
@@ -1,4 +1,5 @@
-[![Build Status](https://travis-ci.org/jscs-dev/node-jscs.svg?branch=master)](https://travis-ci.org/jscs-dev/node-jscs)
+﻿[![Build Status](https://travis-ci.org/jscs-dev/node-jscs.svg?branch=master)](https://travis-ci.org/jscs-dev/node-jscs)
```
mikesherov: Why is this line changed? Seems no different to me? @am11, are you using correct line endings?

am11: @mikesherov, yeah that was both CRLF and BOM. Fixed by 3477d8a76b2d3dfeee16d0e2c6610f169791f12c.

mikesherov:
@am11 can you please rebase this. @markelog seems like we're good now with landing this?

appveyor.yml
```diff
@@ -0,0 +1,12 @@
...
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - node --version
+  - npm --version
+  - npm install
+
+test_script:
+  - npm test
+
+build: off
```
markelog: Can you give me link to this setting?

am11: Sure: http://www.appveyor.com/docs/appveyor-yml (Ctrl+F, search for: '174')

am11:
@mikesherov, done. :)

appveyor.yml
am11: They use `build` slug for msbuild systems (mostly Visual Studio projects).

markelog:
LGTM

PR #941, opened 2015-01-20
StringChecker: leading hashbangs still report correct error line numbers

mikesherov:


lib/string-checker.js
```diff
@@ -76,6 +76,41 @@ StringChecker.prototype = {
...
+     * @return {Object}     the output AST
+     * @throws Error on invalid source code.
+     */
+    _parse: function(str) {
+        var hashbang = (str[0] === '#' && str[1] === '!');
+        var tree;
+
+        // Convert bin annotation to a comment
+        if (hashbang) {
+            str = '//' + str.substr(2, str.length);
```
mrjoelkemp: Why not just strip out the line (from `#!` to `\n` from `str`?

lib/string-checker.js
```diff
@@ -76,6 +76,41 @@ StringChecker.prototype = {
...
 
     /**
+     * Parses source code into an AST
+     *
+     * @param  {String} str the source code to parse
+     * @return {Object}     the output AST
+     * @throws Error on invalid source code.
+     */
+    _parse: function(str) {
+        var hashbang = (str[0] === '#' && str[1] === '!');
```
mrjoelkemp: Same thing? `str.indexOf('#!') === 0`

lib/string-checker.js
mikesherov: because then you're creating multiple line breaks which throws off `disallowMultipleLineBreaks`

mikesherov:
@mrjoelkemp can I get a :+1: here.

lib/string-checker.js
```diff
@@ -76,6 +76,41 @@ StringChecker.prototype = {
...
+     * @return {Object}     the output AST
+     * @throws Error on invalid source code.
+     */
+    _parse: function(str) {
+        var hashbang = str.indexOf('#!') === 0;
+        var tree;
+
+        // Convert bin annotation to a comment
+        if (hashbang) {
+            str = '//' + str.substr(2);
```
markelog: Could that conflict with `requireSpaceAfterLineComment ` rule?

mrjoelkemp:
:+1: assuming Oleg's comment is not applicable.

lib/string-checker.js
markelog: https://github.com/jscs-dev/node-jscs/pull/941/files#diff-c4ec64ef009f51578bddc596b5b6ddc2R107

I see, okay :-)

test/rules/disallow-multiple-line-breaks.js
```diff
@@ -3,29 +3,37 @@ var assert = require('assert');
...
         assert(checker.checkString('x = 1;\n\ny = 2').isEmpty());
     });
+
     it('should not report single line break when first line starts with #!', function() {
-        checker.configure({ disallowMultipleLineBreaks: true });
         assert(checker.checkString('#!/usr/bin/env node\nx = 1;').isEmpty());
         assert(checker.checkString('#!/usr/bin/env node\n\nx = 1;').isEmpty());
     });
+
+    it('should report error in the correct location when first line starts with #!', function() {
```
markelog: This pull for the `string-checker` module, but test for the rule, could we add one to for the relevant module too?

Like here - https://github.com/jscs-dev/node-jscs/blob/30049c1da94b590b783c1ba948da1d639b8c06bc/test/string-checker.js#L38

mikesherov:
Closed by 0253717989b475999e2deea19f124ff1cbb502fd

Issue #698, opened 2014-10-19
New format attempt


markelog:
<b>WIP</b>

Related to https://github.com/jscs-dev/node-jscs/issues/136
This is what i got so far - https://gist.github.com/markelog/752e53faf9f27de4024d

/cc @mikesherov @mdevils @mrjoelkemp @nschonni

hzoo:
For new/modified rules now, is it still

```
 * "disallowDanglingUnderscores": { allExcept: ["_exception"] }
```

because some rules like requireSpacesInsideParentheses are using

```
 * "requireSpacesInsideParentheses": {
 *     "all": true,
 *     "except": [ "{", "}" ]
 * }
```

mikesherov:
Just wanted to drop a note that the "2-level" approach to configuration here means:

1. the way plugins and getOptionName work needs to change.
2. the way enable/disable comments works needs to change.

I'm not sure again as to the benefit of having 2 levels for rules. It introduces the question of what belongs in the top level vs. the bottom level.

If the options are all normalized, simplified (meaning combining beforeAndAfter and require/disallow) and some names shortened, 1 level looks fine too IMO:

```
"binarySpacing": "require",
"unarySpacing": { require: { allExcept: "++" } },
"curlyBraceBlocks": "require",
"curlyBraceNewlines": "require"
````

Just something to consider.

markelog:
I think we're at the point of prototyping this thing, i'm sure will find more caveats along the way or possible improvements.

mrjoelkemp:
A bit of discussion on this issue going on in Gitter: https://gitter.im/jscs-dev/node-jscs?at=55566279811d64626eebcb40. 

We're torn between forcing users to change their configuration in 2.0 versus having an internal conversion proxy to allow users to not have to change their config.

markelog:
For this issue to resolve i think we need to:
- declaratively specify rule options #480, #895
- preserve back-compat - both old and new format should work
- create classification strategy of which rules should go to which block, like "block", "vars", etc; see the proposed format
- actually add new parsing for the new format
- rules clean-up, at least #853, #538 #1349

Only last one item is a breaking change.

markelog:
In other words, is there any takers for the proposed items?

mrjoelkemp:
Yes, I'm looking at #895.

PR #942, opened 2015-01-20
requireCapitalizedConstructors: accept list of exempt constructors

### LINKED ISSUE #390: Ability to provide exceptions to requireCapitalizedConstructors
`requireCapitalizedConstructors` is triggering exceptions on native classes declared by the browser, such as `mozRTCPeerConnection`.

Please detect such conditions automatically (I'm not sure that's possible) or provide a mechanism to provide exceptions to this rule.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/2274423-ability-to-provide-exceptions-to-requirecapitalizedconstructors?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

samlecuyer:
This allows requireCapitalizedConstructors to accept either `true`
or a list of constructors that are allowed to remain uncapitalized
By passing a list of identifiers, `true` is assumed.

fixes #390

### DIFF
--- a/lib/rules/require-capitalized-constructors.js
+++ b/lib/rules/require-capitalized-constructors.js
@@ -1,9 +1,9 @@
 /**
  * Requires constructors to be capitalized (except for `this`)
  *
- * Type: `Boolean`
+ * Type: `Boolean` or `Object`
  *
- * Values: `true`
+ * Values: `true` or Object with `allExcept` Array of quoted identifiers which are exempted
  *
  * JSHint: [`newcap`](http://jshint.com/docs/options/#newcap)
  *
@@ -11,6 +11,9 @@
  *
  * ```js
  * "requireCapitalizedConstructors": true
+ * "requireCapitalizedConstructors": {
+ *     allExcept: ["somethingNative"]
+ * }
  * ```
  *
  * ##### Valid
@@ -18,6 +21,7 @@
  * ```js
  * var a = new B();
  * var c = new this();
+ * var d = new somethingNative();
  * ```
  *
  * ##### Invalid
@@ -32,16 +36,19 @@ var assert = require('assert');
 module.exports = function() {};
 
 module.exports.prototype = {
-
     configure: function(requireCapitalizedConstructors) {
         assert(
-            typeof requireCapitalizedConstructors === 'boolean',
-            'requireCapitalizedConstructors option requires boolean value'
+            requireCapitalizedConstructors === true || Array.isArray(requireCapitalizedConstructors.allExcept),
+            'requireCapitalizedConstructors option requires object of exceptions or true value'
         );
-        assert(
-            requireCapitalizedConstructors === true,
-            'requireCapitalizedConstructors option requires true value or should be removed'
-        );
+        this._allowedConstructors = {};
+
+        var allExcept = requireCapitalizedConstructors.allExcept;
+        if (allExcept) {
+            for (var i = 0, l = allExcept.length; i < l; i++) {
+                this._allowedConstructors[allExcept[i]] = true;
+            }
+        }
     },
 
     getOptionName: function() {
@@ -49,8 +56,11 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
+        var allowedConstructors = this._allowedConstructors;
+
         file.iterateNodesByType('NewExpression', function(node) {
             if (node.callee.type === 'Identifier' &&
+                !allowedConstructors[node.callee.name] &&
                 node.callee.name[0].toUpperCase() !== node.callee.name[0]
             ) {
                 errors.add(
--- a/test/rules/require-capitalized-constructors.js
+++ b/test/rules/require-capitalized-constructors.js
@@ -4,25 +4,50 @@ var assert = require('assert');
 describe('rules/require-capitalized-constructors', function() {
     var checker;
 
+    function baseCases() {
+        it('should report uncapitalized construction', function() {
+            assert(checker.checkString('var x = new y();').getErrorCount() === 1);
+        });
+
+        it('should not report capitalized construction', function() {
+            assert(checker.checkString('var x = new Y();').isEmpty());
+        });
+
+        it('should not report member expression construction', function() {
+            assert(checker.checkString('var x = new ns.y();').isEmpty());
+        });
+
+        it('should not report construction with "this" keyword', function() {
+            assert(checker.checkString('var x = new this();').isEmpty());
+        });
+    }
+
     beforeEach(function() {
         checker = new Checker();
         checker.registerDefaultRules();
-        checker.configure({ requireCapitalizedConstructors: true });
     });
 
-    it('should report uncapitalized construction', function() {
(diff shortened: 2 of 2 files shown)


samlecuyer:
Can someone please restart the travis build?

lib/rules/require-capitalized-constructors.js
```diff
@@ -11,6 +11,9 @@
  *
  * ```js
  * "requireCapitalizedConstructors": true
+ * "requireCapitalizedConstructors": [
```
mrjoelkemp: As per https://github.com/jscs-dev/node-jscs/issues/698, it might be more appropriate for this to be `"requireCapitalizedConstructors": { allExcept: ['somethingNative', 'anotherNativeThing'] }`

samlecuyer: Would something like
```
"requireCapitalizedConstructors": { 
    'somethingNative': false, 
    'anotherNativeThing': false
}
```
work for you?

mrjoelkemp: Brainstorming is definitely appreciated and welcomed. However, the new config format has been agreed upon by the community and maintainers, so let's stick with using the `allExcept` format unless @markelog feels there's another format from #698 that makes more sense.

Thanks for the PR @samlecuyer. I'll do a more careful review once the config change is done. Feel free to ping me.

samlecuyer: @mrjoelkemp thanks for the feedback, I've updated to follow the new format.

lib/rules/require-capitalized-constructors.js
```diff
@@ -32,16 +36,19 @@ var assert = require('assert');
...
-        );
-        assert(
-            requireCapitalizedConstructors === true,
-            'requireCapitalizedConstructors option requires true value or should be removed'
+            requireCapitalizedConstructors === true || Array.isArray(requireCapitalizedConstructors.allExcept),
+            'requireCapitalizedConstructors option requires object of exceptions or true value'
         );
+        this._allowedConstructors = {};
+
+        var allExcept = requireCapitalizedConstructors.allExcept;
```
mrjoelkemp: If a `true` value is supplied as `requireCapitalizedConstructors`, this will throw an error.

samlecuyer: `(true).allExcept` returns `undefined`, the `if (allExcept)` one line down prevents an error.

mrjoelkemp: Well damn. Thanks for the tip :) I assumed booleans had a similar behavior to other primitives (e.g., `1.foobar` throws) – but ya know, JavaScript :)

lib/rules/require-capitalized-constructors.js
```diff
@@ -49,8 +56,11 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
+        var allowedConstructors = this._allowedConstructors;
+
         file.iterateNodesByType('NewExpression', function(node) {
             if (node.callee.type === 'Identifier' &&
+                allowedConstructors[node.callee.name] !== true &&
```
mrjoelkemp: `!allowedConstructors[node.callee.name]`?

test/rules/require-capitalized-constructors.js
```diff
@@ -4,25 +4,59 @@ var assert = require('assert');
...
+            checker = new Checker();
+            checker.registerDefaultRules();
+            checker.configure({
+                requireCapitalizedConstructors: {
+                    allExcept: ['require']
+                }
+            });
+        });
+
+        it('should report uncapitalized construction', function() {
```
mrjoelkemp: Maybe extract the cases that are redundant between the describes into a `checkBaseCases()` function? That'll make it clearer which cases are unique to the different configuration value.

lib/rules/require-capitalized-constructors.js
alawatthe: ```1.foobar``` throws, because JavaScript thinks that the ```.``` is the decimal point and expects more numbers after it (and ```foobar``` is not a number). So you could say ```1.foobar``` is an illegal numerical expression, which therefore throws.
If you add a second point like ```1..foobar``` the second point ends the numerical expression and acts as "property access point" (is there an official word for this?). And since there is no ```.foobar``` on numbers it returns undefined.

mrjoelkemp: That's awesome :)

samlecuyer: @mrjoelkemp like @alawatthe said, it's a parsing error vs a type error issue.

lib/rules/require-capitalized-constructors.js
zxqfox: +1 for `allExpect`.

It's not clear to declare exceptions immediately after rule. That will look like "...require capitalized constructors. which ones? these... A B C" — definitely the opposite sense.

Good job at all. Thanks for contributing!

samlecuyer:
@mrjoelkemp I cleaned up the things you mentioned and squashed them into a single commit (sorry to rewrite the history :|)

mrjoelkemp:
Thanks @samlecuyer. Looks good from a quick glance. I'll give it a more careful look and merge today. Nice work!

Oh, and rewriting history is totally fine for the commits that have yet to be merged into master. We typically strive for a single commit unless you're touching another file as part of a cleanup.

test/rules/require-capitalized-constructors.js
```diff
@@ -4,25 +4,49 @@ var assert = require('assert');
...
+        });
+
+        baseCases();
     });
 
-    it('should not report construction with "this" keyword', function() {
-        assert(checker.checkString('var x = new this();').isEmpty());
+    describe('with `allExcept` value', function() {
+        beforeEach(function() {
+            checker = new Checker();
```
mikesherov: move this and the next line back into a top level `beforeEach`, and only have the `configure` call in each inner `beforeEach`.

.gitignore
```diff
@@ -1,5 +1,6 @@
 /node_modules
 .idea
+.*.swp
```
mikesherov: This belongs in your personal global `.gitignore`, not in our project specific one. Please remove.

mikesherov:
@samlecuyer left a few more comments, please address! Thanks again for contributing!

.gitignore
alawatthe: I agree, that ```.*.swp``` belongs in the personal global ```.gitignore```. But isn't this also true for ```.idea```and ```*.sublime-*```? Just wondering...

samlecuyer:
@mikesherov fixed!

mrjoelkemp:
This looks good. I'll merge.

mrjoelkemp:
Landed! Thanks again @samlecuyer! Great job on this rule.

Issue #918, opened 2015-01-13
Forbid usage of object


janbaer:
Is there a rule with that we can forbid the usage of some specific functions or objects like *console.log* ?

alawatthe:
I think such a rule belongs more to a tool like eslint or jshint. jscs is about the format of the code.

If you are only concerned about *console*, there is a [rule in eslint](http://eslint.org/docs/rules/no-console.html) addressing your needs.

mikesherov:
@janbaer thanks for contributing. @alawatthe is exactly correct, as JSCS is only concerned with style, and this is better suited for either JSHint or ESLint. Thanks again!

janbaer:
Thanks for the answer. We're using currently jshint but it's not really extendable with custom rules. So we decide to use additionally **jscs** to have more flexibility with code checks in a semantic way and with formatting. There is a rule that can check for forbidden keywords like **debugger** in the production code . Why not extend the **jscs** to check also the usage of some functions. The **console** was just an example. Before some weeks I used the jasmine function **ddescribe** to just run a specific test. But I forgot to change it back to **describe**. So for some days our **ci** just executed one test. With a check with **jscs** for example in this wouldn't happen.

alawatthe:
So I gave this a second thought. I think a rule like
```js
disallowVariableNames: ['temp', 'foo', 'bar', 'baz', 'asdf']
```
could make sense, as this is a stylistic choice about the code.
@janbaer This should fit your use case, too.

I am just not a too big fan of a rule like
```js
disallowUseOfObject: ['console']
```
Since it is impossible to completely prohibit the use of an object through static analysis.

@mikesherov should we reopen this issue?

mrjoelkemp:
Duplicate of #41

Issue #612, opened 2014-08-31
add rule for disallowing naive variables


mocheng:
Some variable patterns are not preferred since it hurts readability. For example,

var tmp = 123; // what the heck of tmp? It should be specific
var arr = "hello world".split(" "); // arr is meaningless. Variable words would be better

Is this kind of rule aligned with JSCS goals? If yes, I can contribute a PR.

markelog:
This issue has been open more than a month without supported PR for it. 

Unfortunately, we don't have the resources to implement rules that could not be included in any of the supported [presets](https://github.com/jscs-dev/node-jscs/tree/master/presets) (if it could be, please point that out).

However, if you or anybody else submit PR for it, we will gladly consider it.

mrjoelkemp:
Duplicate of #41

pohutukawa:
I think blacklisting variable names is a possible solution, but prone to circumvention through "inventiveness" of other developers trickiness. However, together with a min length, this might actually already be quite helpful. In fact anything that helps to kerb "silly" variables and widely scoped "short" variables helps.

PR #943, opened 2015-01-20
implemented disallow-variable-names

### LINKED ISSUE #41: Validate bad variable names
I think we should add a config field which can validate bad variable names.

All one- two-symbol vars, expect in-loop vars like i, j. Abstract names like `item`.

alawatthe:
disallowVariableNames: first implementation

disallowVariableNames takes an array of strings, which are considered as illegal variable names.

Fixes #41, #612, #918


alawatthe:
So here is a little longer description:

The issue of restricting the set of legal variable names has come up in the past several times, see #41, #612, #918.

The feedback in the discussions for such a rule was positive (I think), but no one created a pull request.


Therefore I'm starting with a pull request.

One open point in the past (see #612) was, if (and if "yes" how?) single letter variable names should be handled differently depending on whether they occur in loops or not.
In my opinion this problem can't be solved in a general way.
This is the reason why my pull request does not include any special handling for single letter variable names.


What do you think?

P.S.: This is my first rule I implemented from ground up, so don't spare me with critic :-)
For example I did know if there is a specified order for all the ```registerRule``` functions.

test/rules/disallow-variable-names.js
```diff
@@ -0,0 +1,18 @@
...
+var assert = require('assert');
+
+describe('rules/disallow-variable-names', function() {
+    var checker;
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+        checker.configure({ disallowVariableNames: ['temp', 'foo'] });
+    });
+    it('should report illegal variable names', function() {
```
mrjoelkemp: spaces before every it please

test/rules/disallow-variable-names.js
```diff
@@ -0,0 +1,18 @@
...
+        assert(checker.checkString('temp = 1;').getErrorCount() === 1);
+        assert(checker.checkString('obj[foo] = 1;').getErrorCount() === 1);
```
mrjoelkemp: Can you also add tests for vars in loops and functions for completeness?

test/rules/disallow-variable-names.js
```diff
@@ -0,0 +1,18 @@
...
+    });
+    it('should not report legal variable names', function() {
+        assert(checker.checkString('good = 1').isEmpty());
```
mrjoelkemp: Same comment about completeness of the tests for other scenarios where vars are declared.

mrjoelkemp:
Do you need to rebase off of master?

lib/config/configuration.js
```diff
@@ -616,6 +616,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-line-break-after-variable-assignment'));
 
     this.registerRule(require('../rules/disallow-semicolons'));
+
+    this.registerRule(require('../rules/disallow-variable-names'));
```
mrjoelkemp: No specific order. This is fine.

mrjoelkemp:
Nice work @alawatthe!

alawatthe:
@mrjoelkemp thank you, for your feedback :-)
I rebased, added the new lines and more tests. If you have more ideas for tests, I can add them, too.

lib/rules/disallow-variable-names.js
```diff
@@ -0,0 +1,57 @@
...
+    },
+
+    getOptionName: function() {
+        return 'disallowVariableNames';
+    },
+
+    check: function(file, errors) {
+        var disallowVariableNames = this._disallowVariableNames;
+        file.iterateNodesByType('Identifier', function(node) {
+            if (disallowVariableNames.indexOf(node.name) !== -1) {
```
mrjoelkemp: The iteration over disallowVariableNames is redundant. In `configure`, you could create an object of var name -> true (ex: `foo`: true) for a constant time lookup. You'd then do `if (disallowVariableNames[node.name])`.

mrjoelkemp: What about uppercase var names? Your check should be case insensitive. I guess an alternative to the object lookup above is to create a regex from all of the disallowed vars and do a case insensitive match on node.name.

mrjoelkemp: Be sure to add tests for the uppercase var names.

alawatthe: My take on the uppercase/lowercase issue was, that one should be able to prohibit the use of ```foo``` but allow the use of ```Foo```. If you want to prohibit the use of ```Foo```, too, you can just add it to the array. Of course, this would still allow ```fOo``` and so on. And thus one could say this leads to exponential complexity...
That said, I can be persuaded on this one :-D
Are there any other opinions on this?

On the object lookup: Yes, the object lookup is probably faster. But this feels a little bit premature opimization like. I am the last person to not improve the code performance wise, but then we should do it right and do performance testing as well. Let me do some experiments with this.
But depending on the uppercase/lowercase issue I can of course implement a object lookup.

mrjoelkemp: @mikesherov any opinion on the uppercase/lowercase issue above?

@alawatthe I was iffy on mentioning the object notation for the same fear of "the root of all evil," but:

1. This is one of the more obvious improvements that doesn't obfuscate the code (it actually clarifies the condition).
2. We have other rules that do this: https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-dangling-underscores.js#L84-L87, https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-implicit-type-conversion.js#L41-L44, https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-keywords.js#L31-L34, and a bunch more.

We'll wait for other opinions on casing to decide on the course of action.

mikesherov: @alawatthe I agree variable should be case sensitive match. With regards to hash lookup instead of indexOf I agree with @mrjoelkemp that our idiomatic way is using hashes instead of array lookups. Please change.

alawatthe: I changed it an object lookup.

Just for the record: I wasn't against the object lookup, I just think we can to do more performance wise. :-)

alawatthe:
Sorry for the noise on this bug :-/

I have been thinking a bit about the name for this rule. Since it also prevents property names like ```obj.bad``` a better name for the rule would be disallowIndentifierNames, i think.

I already changed the name and added tests on my local branch. So, if you agree with the name change I can upload the code in no time.

mikesherov:
@alawatthe sounds good.

mikesherov:
@alawatthe can you all rebase while you're at it? Thanks!

alawatthe:
I screwed the rebase :weary:. Therefore I created a completely new pull request https://github.com/jscs-dev/node-jscs/pull/960.

Sorry for the inconvenience :confused:

PR #948, opened 2015-01-23
disallowStringLiteralStatements: create new rule

samlecuyer:
A common error I see is a misspelled 'use strict' directive. By
adding a rule to not allow string literal statements, we can
avoid this whole class of errors.  This new rule causes an error
any time a string statement other than 'use strict' appears.

### DIFF
--- a/lib/rules/disallow-string-statements.js
+++ b/lib/rules/disallow-string-statements.js
@@ -1,0 +1,51 @@
+/**
+ * Disallows string literal statements.
+ *
+ * Type: `Boolean`
+ *
+ * Values: true
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowStringLiteralStatements": true
+ * ```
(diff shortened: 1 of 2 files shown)


samlecuyer:
@mrjoelkemp would you mind taking a look?

mikesherov:
@samlecuyer, thanks for contributing! 

We currently have a policy of only focusing on style issues, and not ones of program correctness, which are better suited for JSHint or ESLint. Also, you can write custom additionalRules for jscs and open source them as well. Not everything needs to get into core :-)

Also, it's best to open an issue first before implementing so we can discuss a feature's merits before you go and do all that work.

With that said I'm going to close this pull request, but please don't let it discourage you from taking a crack at some of our other open issues.

samlecuyer:
@mikesherov thanks for the feedback!

Issue #949, opened 2015-01-23
validateParameterSeparator: issue with error message with more than 1 space


hzoo:
When you have too many spaces with value: `", "` 
```
Missing space before function parameter '$q' at src/ngMock/angular-mocks.js :
   456 |angular.mock.$IntervalProvider = function() {
   457 |  this.$get = ['$browser', '$rootScope', '$q', '$$q',
   458 |       function($browser,   $rootScope,   $q,   $$q) {
-----------------------------------------------
```

https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/validate-parameter-separator.js#L67
```js
errors.assert.whitespaceBetween({
    token: prevParamToken,
    nextToken: punctuatorToken,
    spaces: 1,
    message: 'Missing space after function parameter \'' + prevParamToken.value + '\''
});
```

I forgot to change the message when changing validateParameterSeparator to be stricter - https://github.com/jscs-dev/node-jscs/issues/701. I think it should say something like `"Exactly one space is allowed after function parameter"`

And maybe we need another option to allow one or more spaces?

hzoo:
Oh this is the same issue as https://github.com/jscs-dev/node-jscs/issues/927.

I guess there must be a few places we should change the error message - for rules starting with `requireSpaces` and are using `errors.assert.whitespaceBetween`

PR #950, opened 2015-01-24
Fix error messages for whitespaceBetween

hzoo:
For #949.

Since the rules enforce only one space.

Also removed `spaces: 1` for `noWhitespaceBetween` since it doesn't use that as an option?


zxqfox:
But... `spaces : 1` is a good thing! :cry:

mikesherov:
@zxqfox @hzoo removed it from *no*WhiteSpaceBetween, because the number is always 0. This LGTM. Landing.

PR #951, opened 2015-01-24
requireSemicolons rule implemetation

lahmatiy:
Boolean rule that requires semicolon after:
* var declaration
* expression statement
* return
* throw
* break
* continue
* do-while

Valid:
```js
var a = 1;
var fn = function(){ return; };
if (a > 0)
  b = a * 5;
```

Invalid:
```js
var a = 1
var fn = function(){ return };
if (a > 0)
  b = a * 5
```


mikesherov:
@lahmatiy rather than commenting out the tests, use `it.skip` and put a comment above them as to why they are skipped. Also, please search https://code.google.com/p/esprima/issues to see if there is an issue already logged in esprima for this.

lahmatiy:
@mikesherov 
I thought comments are better here as `it.skip` too hard to find. I decided do not remove those test as they were before and were successful. Probably those tests should be removed at all.
I've read about Esprima changes about semicolon, but I can't found it for now. Actually it behaves better than before, as semicolon on next line is not an error here (previously it produces empty expression with single semicolon). Also you could format your code in this way:
```js
var foo = 1
  , bar = 2
  , baz = 3
  ;
```
I saw this sort of formatting.
First implementation fails on it (returns error). But now it's ok. And it looks reasonable, as rule should just checks semicolon is not missed.
Should I remove those tests than?
// cc @mdevils

lib/rules/require-semicolons.js
```diff
@@ -0,0 +1,98 @@
...
+            if (node.type === 'VariableDeclaration') {
+                if ((node.parentNode.type === 'ForInStatement' && node.parentNode.left === node) ||
+                    (node.parentNode.type === 'ForStatement' && node.parentNode.init === node)) {
+                    return;
+                }
+            }
+
+            // search for last token inside node
+            var token;
+            for (var i = file.getTokenPosByRangeStart(node.range[0]) + 1; i < tokens.length; i++) {
```
mdevils: Please, use `JsFile::getLastNodeToken()` method (https://github.com/jscs-dev/node-jscs/blob/master/lib/js-file.js#L230).

mdevils:
Great job, @lahmatiy!
Feel free to remove those tests, it is fine without them.

Please use `getLastNodeToken` as I noted above.

lahmatiy:
@mdevils 
Done.
But I moved commented tests to valid. I think rule should has opition about edge cases. It could be changed in future, and those test will a reminder about those cases.

lahmatiy:
@mikesherov @mdevils is there any chance for merge?

mikesherov:
Yes, just give it a bit because we just released 1.11 and need a few days to plan out 1.12

mrjoelkemp:
@lahmatiy Mind rebasing off of master and fixing the jshint violation?

lahmatiy:
@mrjoelkemp rebased and fixed. Looking forward for merge.

lib/rules/require-semicolons.js
```diff
@@ -0,0 +1,88 @@
...
+            'DoWhileStatement',
+            'ReturnStatement',
+            'ThrowStatement',
+            'BreakStatement',
+            'ContinueStatement',
+            'DebuggerStatement'
+        ], function(node) {
+            // ignore variable declaration inside for and for-in
+            if (node.type === 'VariableDeclaration') {
+                if ((node.parentNode.type === 'ForInStatement' && node.parentNode.left === node) ||
```
zxqfox: What about `ForOf`? ;-)

lib/rules/require-semicolons.js
```diff
@@ -0,0 +1,88 @@
...
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(requireSemicolons) {
+        assert(
+            typeof requireSemicolons === 'boolean',
```
zxqfox: No need to check type when later we have explicit assert.

hzoo: A few of the rules are checking boolean and true - we should probably remove those so no one gets confused

lib/rules/require-semicolons.js
lahmatiy: I added `ForOfStatement` support and basic ES6 support as well (tests for `let` and `const` was added). But I'm not sure ES6 is fully supported still (even by `Esprima`).

lahmatiy:
@zxqfox @mrjoelkemp Believe it's ok now?

zxqfox:
Yeah, nicely done! :+1:

I'm only thinking about assert api to make it autofixable. Probably we can left it as is becase the opposite rule still need update.

mikesherov:
Assertion API is whitespace only for 1.12. I don't want to have to tie the two together.

mikesherov:
:+1:

mrjoelkemp:
Thanks for contributing @lahmatiy! Landed.

PR #953, opened 2015-01-25
JSHint: remove "forin" option

mikesherov:
We do not and will never include any module that tamper
with the Object property, so this check is unnecessary
and will prevent us from achieving 100% code coverage.

/cc @mdevils @markelog


markelog:
Yep, remember arguing for that :-)

mrjoelkemp:
:+1:

Issue #747, opened 2014-11-05
Move traversing module (js-file) to its own npm-module


markelog:
`js-file` module is pretty powerful, it might be used in different projects and help others.

Just like [estraverse](https://github.com/Constellation/estraverse) or tools alike standalone we could do the same, but most (all?) of those tools use only AST, whereas we also use tokens and basically created methods that link AST with tokens, which mean you could do more, in some cases much more.

hzoo:
Should we close this specific issue since we're talking about a CST now?

mrjoelkemp:
Seems so. +1 to closing. Will let @markelog decide.

PR #957, opened 2015-01-25
cleanup: remove lib/token-helper

mikesherov:


mikesherov:
/cc @mdevils @markelog @mrjoelkemp Everything in token-helper is superceded by the simpler methods in `js-file`. It has never been part of a documented public API, and only plugin developers who were specifically scrounging around and looking for it would be effected.

Can we remove this for 1.11 or must we wait for 2.0?

mikesherov:
Whatever we decide, we should document it in our versioning guide, and we should also document the api of JsFile, etc.

markelog:
> 98.03%

Wow!

> Can we remove this for 1.11 or must we wait for 2.0?

Wasn't documented, i think it's safe to remove in minor. 
Perhaps we should start documenting `js-file` and `cli-config` modules, one of the steps would be to solve https://github.com/jscs-dev/node-jscs/issues/747

mikesherov:
Ok. I'm going to land this then and agree we should start thinking how to solve. #747

PR #958, opened 2015-01-26
(require|disallow)SpacesInsideArrayBrackets: only check for ArrayExpressions, add new rules for Brackets

mikesherov:


mikesherov:
ping @markelog ... should I also switch all of the presets to use the new rules? None of them have caveats for expression vs. accessor AFAIK.

lib/rules/disallow-spaces-inside-brackets.js
```diff
@@ -0,0 +1,114 @@
...
+ * }
+ * ```
+ *
+ * ##### Valid for mode `true`
+ *
+ * ```js
+ * var x = [[1]];
+ * ```
+ *
+ * ##### Valid for mode `"allExcept"`
```
mrjoelkemp: Can you add `"allExcept": [ "[", "]", "{", "}" ]` instead for clarity?

mrjoelkemp:
:+1: Nice work!

lib/rules/disallow-spaces-inside-brackets.js
mikesherov: will do.

markelog:
> ping @markelog ... should I also switch all of the presets to use the new rules? None of them have caveats for expression vs. accessor AFAIK.

Yeah.

I think after https://github.com/jscs-dev/node-jscs/pull/958#discussion_r23536197 we are good to go.

Nicely done!

mikesherov:
@markelog should we be switching any presets over to the new rule? Or should I make a ticket for 2.0 to switch them over?

markelog:
It seems new rules behave according to code styles.

But i also think we also should address https://github.com/jscs-dev/node-jscs/pull/822 and https://github.com/jquery/contribute.jquery.org/pull/104.

And do this sort of thing until 2.0 is released. After that, switch to strict release tactics.

mikesherov:
@markelog OK, so I'm going to merge as is, and then file issue to add new rules later.

zxqfox:
I'm sorry, but what about block comments between bracket and the next token?
```
var x = [/*1*/ 'a' ];
var x = [ /*2*/'a' ];
```
etc.

mikesherov:
@zxqfox we can address that in another PR. This was about restoring old functionality.

zxqfox:
almost managed ;-(

mikesherov:
@zxqfox sorry, I pushed right before your comment! Block comments inside brackets is extremely rare... but you should file bugs against the existing rules.

markelog:
> but you should file bugs against the existing rules.

+1

PR #960, opened 2015-01-26
implemented disallow-identifier-names

alawatthe:
Related to https://github.com/jscs-dev/node-jscs/pull/943. I screwed the rebase :weary: Sorry

### DIFF
--- a/lib/rules/disallow-identifier-names.js
+++ b/lib/rules/disallow-identifier-names.js
@@ -1,0 +1,63 @@
+/**
+ * Disallows a specified set of identifier names.
+ *
+ * Type: `Array`
+ *
+ * Values: Array of strings, which should be disallowed as identifier names
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowIdentifierNames": ['temp', 'foo']
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var good = 1;
+ * object[fine] = 2;
+ * object.fine = 3;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var temp = 1;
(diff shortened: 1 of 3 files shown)


lib/rules/disallow-identifier-names.js
```diff
@@ -0,0 +1,63 @@
...
+ * var good = 1;
+ * object[fine] = 2;
+ * object.fine = 3;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var temp = 1;
+ * object[foo] = 2;
```
mrjoelkemp: `foo` should be quoted

lib/rules/disallow-identifier-names.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * ```js
+ * "disallowIdentifierNames": ['temp', 'foo']
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var good = 1;
+ * object[fine] = 2;
```
mrjoelkemp: `fine` should be quoted

mrjoelkemp:
A test asserting case sensitivity would be great. In other words, there should be a test that shows that `FOO` does not result in an error.

lib/rules/disallow-identifier-names.js
alawatthe: The rule currently does not prohibit ```object['foo']``` (therefore, the foo should not have quotes in the example).

That said, I agree that we want to prohibit ```object['foo']```, too. Let me implement it.

mrjoelkemp:
@alawatthe Sorry this fell through the cracks. Great work here. Thanks for contributing. Landed!

I think this will be a popular rule.

PR #963, opened 2015-01-27
Obey the Rule of Silence

### LINKED ISSUE #962: jscs violates the Rule of Silence
When `jscs` finds no errors it prints:

```bash
No code style errors found.
```

This violates the Rule of Silence:

> The rule of silence, also referred to as the silence is golden rule, is an important part of the Unix philosophy that states that when a program has nothing surprising, interesting or useful to say, it should say nothing. It means that well-behaved programs should treat their users' attention and concentration as being valuable and thus perform their tasks as unobtrusively as possible. That is, silence in itself is a virtue. (http://www.linfo.org/rule_of_silence.html)

What would you say to a PR that made `jscs` silent by default?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8055740-jscs-violates-the-rule-of-silence?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

feross:
Fixes #962

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -22,7 +22,5 @@ module.exports = function(errorsCollection) {
          * Printing summary.
          */
         console.log('\n' + errorCount + ' code style ' + (errorCount === 1 ? 'error' : 'errors') + ' found.');
-    } else {
-        console.log('No code style errors found.');
     }
 };
--- a/lib/reporters/text.js
+++ b/lib/reporters/text.js
@@ -22,7 +22,5 @@ module.exports = function(errorsCollection) {
          * Printing summary.
          */
         console.log('\n' + errorCount + ' code style ' + (errorCount === 1 ? 'error' : 'errors') + ' found.');
-    } else {
-        console.log('No code style errors found.');
     }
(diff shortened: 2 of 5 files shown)


mrjoelkemp:
Nice job @feross! :+1:

mikesherov:
I'm not sure we could land this till 2.0, as there maybe programs that are (albeit incorrectly) checking for the "no errors" message rather than the success error code.

feross:
Yep, I agree. This is definitely a major version bump.

Though, you could publish this as 2.0.0 immediately, if you're so inclined. Bumping the major version isn't that big of a deal :) Browserify is already on 8.x.x. Not to mention the version Chrome is on! :p

mdevils:
Agree with @mikesherov

domenic:
+1

PR #966, opened 2015-01-27
requireParenthesesAroundIIFE: ensure to get the closingToken

### LINKED ISSUE #965: requireParenthesesAroundIIFE: ensure to get the closingToken
rule `requireParenthesesAroundIIFE`:

```javascript
// foo.js
var a = function() { return 1; }()
```
```sh
$ jscs foo.js

TypeError: Cannot read property 'value' of undefined
    at isWrapped (/usr/local/lib/node_modules/jscs/lib/rules/require-parentheses-around-iife.js:29:77)
......
```

If the semicolon is omitted，the [`tokens[closingToken + 1]`](https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/require-parentheses-around-iife.js#L64) is not reachable

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8076381-requireparenthesesaroundiife-ensure-to-get-the-closingtoken?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zhangyuheng:
This fixes bug for checking string with no trailing semicolon
since the closing token is not reachable.

Fixes #965

### DIFF
--- a/lib/rules/require-parentheses-around-iife.js
+++ b/lib/rules/require-parentheses-around-iife.js
@@ -61,7 +61,10 @@ module.exports.prototype = {
             var openingToken = file.getTokenPosByRangeStart(node.range[0]);
             var closingToken = file.getTokenPosByRangeStart(node.range[1] - 1);
 
-            return tokens[openingToken - 1].value + tokens[closingToken + 1].value === '()';
+            var closingTokenValue = tokens[closingToken + 1] ?
+                tokens[closingToken + 1].value : undefined;
+
+            return tokens[openingToken - 1].value + closingTokenValue === '()';
         }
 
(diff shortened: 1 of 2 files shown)


mikesherov:
Thanks for contributing! This is one of our oldest rules and there are now better APIs to handle the edge case.

I'll take the tests from your PR, but I'll clean up the file when I land it.

Thanks again!

PR #967, opened 2015-01-28
Speed up builds by using Docker-based Travis CI

xaka:
It should reduce the build time down to ~20s. More details are at http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 language: node_js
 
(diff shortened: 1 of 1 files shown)


mikesherov:
Thanks @xaka! I'll land this tomorrow.

mikesherov:
Thanks again!

PR #970, opened 2015-01-28
 requireSpace(AfterPrefix|BeforePostfix)UnaryOperators: use tokenAssert ...

hzoo:
...methods

Refactoring to use

```js
file.getFirstNodeToken(node);
file.getNextToken(node);
errors.assert.whitespaceBetween({});
```

Also newlines in the test files


mikesherov:
Thanks!

PR #975, opened 2015-01-29
spaces* rules: use tokenAssert methods

hzoo:
checked for anything like

```
var openingBracket = file.getTokenPosByRangeStart(node.range[0]);
openingBracket = tokens[openingBracketPos];
var nextToken = tokens[openingBracketPos + 1];
```
Used `file.get[First|Last]NodeToken` and `file.get[Next|Prev]Token` instead

I didn't change the IIFE rule since there's already a PR for that.

These methods definitely make it easier to read/create new whitespace rules!

### DIFF
--- a/lib/rules/disallow-spaces-in-anonymous-function-expression.js
+++ b/lib/rules/disallow-spaces-in-anonymous-function-expression.js
@@ -80,7 +80,6 @@ module.exports.prototype = {
     check: function(file, errors) {
         var beforeOpeningRoundBrace = this._beforeOpeningRoundBrace;
         var beforeOpeningCurlyBrace = this._beforeOpeningCurlyBrace;
-        var tokens = file.getTokens();
 
         file.iterateNodesByType(['FunctionExpression'], function(node) {
             var parent = node.parentNode;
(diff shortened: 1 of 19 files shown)


lib/rules/disallow-spaces-in-anonymous-function-expression.js
```diff
@@ -96,8 +96,7 @@ module.exports.prototype = {
                 if (beforeOpeningRoundBrace) {
                     var nodeBeforeRoundBrace = node;
 
-                    var functionTokenPos = file.getTokenPosByRangeStart(nodeBeforeRoundBrace.range[0]);
-                    var functionToken = tokens[functionTokenPos];
+                    var functionToken = file.getFirstNodeToken(nodeBeforeRoundBrace);
 
                     var nextTokenPos = file.getTokenPosByRangeStart(functionToken.range[1]);
```
mikesherov: Update this.

mikesherov: Just doublecheck that this is replaced everywhere it can be.

lib/rules/require-space-after-object-keys.js
```diff
@@ -46,12 +46,13 @@ module.exports.prototype = {
...
-                var colon = tokens[keyPos + 1];
-                if (colon.range[0] === key.range[1]) {
-                    errors.add('Missing space after key', key.loc.end);
-                }
+                var token = file.getFirstNodeToken(property.key);
+                errors.assert.whitespaceBetween({
+                    token: token,
+                    nextToken: file.getNextToken(token),
+                    message: 'Missing space after key',
+                    spaces: 1
```
mikesherov: Remove spaces: 1 to maintain backCompat.

mikesherov:
Thanks as always for contributing! Just a few minor updates on this one.

hzoo:
Ok I added some more for `(disallow|require)SpacesIn*` rules.

I'l also go back and fix the things you commented and rebase

lib/rules/require-space-after-object-keys.js
hzoo: I should probably change the other commit as well then https://github.com/jscs-dev/node-jscs/commit/f02aa48c5c6f5b12d6bc71356c7cc354dfe23f9b#diff-e64981daea06ee3997fe34d106961f0eR79 to remove spaces

mikesherov:
Thanks again!

PR #979, opened 2015-01-29
Removed requireMultipleVarDecl option

stefanosala:
Looking at [here](https://github.com/airbnb/javascript#variables) it seems that it's not required to have only one `var` declaration, but the opposite.


mikesherov:
Thanks for contributing! This is a dupe of #822

stefanosala:
Ops, thanks :)

markelog:
Updated preset will be available in 1.11

PR #980, opened 2015-01-29
Update: .editorconfig - fix for invalid indent_style value

valorkin:
@mikesherov 
`indent_style` value - `spaces` is not valid,
changed to `space`


markelog:
Thank you!

Issue #436, opened 2014-06-12
Feature Request: Exception to Indentation Rules for Module Pattern


danvk:
When using the module pattern and JSCS's ```validateIndentation``` option, you're forced to indent your entire JS file. For example, this:

    (function() {
    
    function foo() {
      alert("hello");
    }

    window.foo = foo;
    
    })();

becomes

    (function() {
    
      function foo() {
        alert("hello");
      }

      window.foo = foo;
    
    })();

i.e. you have to indent your entire file because you've wrapped it in the module pattern. This makes your effective line limit shorter, and generally seems like a waste of space.

I like the ```validateIndentation``` option, but it would be nice if there were a way to make an exception for this particular pattern.

PR #981, opened 2015-01-30
validateIndentation: exception to indentation rules for module pattern

mikesherov:


lib/utils.js
```diff
@@ -147,6 +147,33 @@ exports.isSnakeCased = function(name) {
 };
 
 /**
+ * Returns the function expression node if the provided node is an iffe,
+ * other returns undefined.
+ *
+ * @param  {Object} node
+ * @return {Object|undefined}
+ */
+exports.getFunctionNodeFromIIFE = function(node) {
```
markelog: This method could be used anywhere else?

markelog: Strike that, see the use

lib/utils.js
```diff
@@ -147,6 +147,33 @@ exports.isSnakeCased = function(name) {
 };
 
 /**
+ * Returns the function expression node if the provided node is an iffe,
+ * other returns undefined.
+ *
+ * @param  {Object} node
+ * @return {Object|undefined}
```
markelog: Usually, if there is no object to return, null is used instead of undefined.

test/rules/validate-indentation.js
```diff
@@ -150,6 +150,67 @@ describe('rules/validate-indentation', function() {
...
+        beforeEach(function() {
+            checker.configure({ validateIndentation: 4 });
+        });
+
+        var cases = {
+            'no indentation': 'a++;',
+            indentation: '    a++;'
+        };
+
+        Object.keys(cases).forEach(function(title) {
```
markelog: These kind of test declarations are harder to isolate, not sure if copy/paste is worse, i did it couple times myself though

markelog:
I would divide this commit, `require-parentheses-around-iife` changes are nothing to do with #436.

Code looks better, comments especially helpful, but i'm afraid bus factor of the validateIndentation rule is equal to one. @zxqfox, @mdevils, @mrjoelkemp i'm i right?

mikesherov:
Yes, hit by a bus factor is still high. 

I recently spent a lot of time cleaning up this rule to reduce the complexity of the algorithm. 

I also spent a bit of time with @mrjoelkemp reviewing the rule and how it works.

Next, I'm going to write up how this rule works, because the plan is to have other indentation rules extend from it and use its interface (which I have not exposed yet).

The hope is to have someone else write one of the other indentation rules using the interface. That way, when I buy a private island, all knowledge of indentation will not be lost :-)

lib/utils.js
mikesherov: Good point. I'll change that.

zxqfox: `{?Object}` ?

test/rules/validate-indentation.js
mikesherov: I think copy/paste is worse in this case. I'm going to leave it unless you have strong objection.

lib/utils.js
mikesherov: @zxqfox, if that's valid JSDoc, I'll use it.

zxqfox: @mikesherov Absolutely valid. http://usejsdoc.org/tags-type.html It's called nullable type.

mrjoelkemp: @zxqfox That's cool. Great tip.

mikesherov:
Thanks for the review. Landed with the proposed modifications!

PR #983, opened 2015-01-31
rules*: use tokenAssert methods

hzoo:
- requireCurlyBraces: use tokenAssert (make it similar to disallowCurlyBraces)

### DIFF
--- a/lib/rules/disallow-space-between-arguments.js
+++ b/lib/rules/disallow-space-between-arguments.js
@@ -46,15 +46,10 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-
-        var tokens = file.getTokens();
-
         file.iterateNodesByType(['CallExpression'], function(node) {
-
             node.arguments.forEach(function(param, i) {
-
-                var punctuatorToken = file.getTokenByRangeStart(param.range[1]);
-
+                var token = file.getFirstNodeToken(param);
+                var punctuatorToken = file.getPrevToken(token);
                 if (punctuatorToken.value === ',') {
                     errors.assert.noWhitespaceBetween({
                         token: punctuatorToken,
--- a/lib/rules/require-curly-braces.js
+++ b/lib/rules/require-curly-braces.js
@@ -67,17 +67,11 @@ module.exports.prototype = {
 
     check: function(file, errors) {
 
-        function addError(typeString, node) {
-            var entity = node;
+        function isNotABlockStatement(node) {
+            return node && node.type !== 'BlockStatement';
+        }
 
-            if (typeString === 'Else') {
-                entity = file.findPrevToken(
-                    file.getTokenByRangeStart(node.alternate.range[0]),
-                    'Keyword',
-                    'else'
-                );
-            }
-
+        function addError(typeString, entity) {
             errors.add(
                 typeString + ' statement without curly braces',
                 entity.loc.start.line,
@@ -87,7 +81,7 @@ module.exports.prototype = {
 
         function checkBody(type, typeString) {
             file.iterateNodesByType(type, function(node) {
-                if (node.body.type !== 'BlockStatement') {
+                if (isNotABlockStatement(node.body)) {
                     addError(typeString, node);
                 }
             });
@@ -97,14 +91,12 @@ module.exports.prototype = {
 
         if (typeIndex['if'] || typeIndex['else']) {
             file.iterateNodesByType('IfStatement', function(node) {
-                if (typeIndex.if && node.consequent.type !== 'BlockStatement') {
+                if (typeIndex['if'] && isNotABlockStatement(node.consequent)) {
                     addError('If', node);
                 }
-                if (typeIndex['else'] && node.alternate &&
-                    node.alternate.type !== 'BlockStatement' &&
-                    node.alternate.type !== 'IfStatement'
-                ) {
-                    addError('Else', node);
+                if (typeIndex['else'] && isNotABlockStatement(node.alternate) &&
+                    node.alternate.type !== 'IfStatement') {
+                    addError('Else', file.getPrevToken(file.getFirstNodeToken(node.alternate)));
                 }
             });
         }
--- a/lib/rules/require-spaces-in-for-statement.js
+++ b/lib/rules/require-spaces-in-for-statement.js
@@ -58,22 +58,19 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         file.iterateNodesByType('ForStatement', function(node) {
-            var prevToken;
             if (node.test) {
-                var testToken = file.getTokenByRangeStart(node.test.range[0]);
-                prevToken = file.getPrevToken(testToken);
+                var testToken = file.getFirstNodeToken(node.test);
                 errors.assert.whitespaceBetween({
-                    token: prevToken,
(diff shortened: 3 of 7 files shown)


lib/rules/require-curly-braces.js
```diff
@@ -97,14 +91,12 @@ module.exports.prototype = {
...
+                if (typeIndex['if'] && isNotABlockStatement(node.consequent)) {
                     addError('If', node);
                 }
-                if (typeIndex['else'] && node.alternate &&
-                    node.alternate.type !== 'BlockStatement' &&
-                    node.alternate.type !== 'IfStatement'
-                ) {
-                    addError('Else', node);
+                if (typeIndex['else'] && isNotABlockStatement(node.alternate) &&
+                    node.alternate.type !== 'IfStatement') {
```
hzoo: ` && node.alternate.type !== 'IfStatement'` is kept b/c otherwise some of the tests regarding presets failed

```
modules/string-checker presets preset crockford should not report any errors from the sam
ple file:
     AssertionError: false == true
```

Not sure if it means we need to add that to disallow

mikesherov: Yes, it needs to be kept only for the require case, because otherwise `else if...` would need to be `else { if...` because technically that else doesn't have curly braces.

mikesherov: But that also means we should add a test for `else if` to this PR in the require-curly-braces suite so we don't lose coverage should we ever remove the preset tests.

hzoo: Oh ok that makes sense

hzoo:
added some other files that I missed before

lib/rules/disallow-space-between-arguments.js
```diff
@@ -46,20 +46,13 @@ module.exports.prototype = {
...
-
-        var tokens = file.getTokens();
-
         file.iterateNodesByType(['CallExpression'], function(node) {
-
             node.arguments.forEach(function(param, i) {
-
-                var punctuatorToken = file.getTokenByRangeStart(param.range[1]);
-
+                var punctuatorToken = file.getPrevToken(file.getFirstNodeToken(param));
```
zxqfox: You using here `punctuatorToken = getPrevToken` and later `nextToken: file.getNextToken(punctuatorToken)`.

So `file.getFirstNodeToken` is already a `nextToken` for `punctuatorToken`.

You can get rid of unnecessary `getNetToken` call by storing `firstNodeToken`.

lib/rules/require-spaces-in-for-statement.js
```diff
@@ -58,22 +58,19 @@ module.exports.prototype = {
...
     check: function(file, errors) {
         file.iterateNodesByType('ForStatement', function(node) {
-            var prevToken;
             if (node.test) {
-                var testToken = file.getTokenByRangeStart(node.test.range[0]);
-                prevToken = file.getPrevToken(testToken);
+                var testToken = file.getFirstNodeToken(node.test);
                 errors.assert.whitespaceBetween({
-                    token: prevToken,
+                    token: file.getPrevToken(testToken),
```
zxqfox: What if there is a block comment between these tokens?

lib/rules/disallow-space-between-arguments.js
hzoo: The `punctuatorToken.value === ','` is important
Ok I should just rename to this then.

```js
var nextToken = file.getFirstNodeToken(param);
                var punctuatorToken = file.getPrevToken(nextToken);
                if (punctuatorToken.value === ',') {
                    errors.assert.noWhitespaceBetween({
                        token: punctuatorToken,
                        nextToken: file.getNextToken(punctuatorToken)
                    });
                }
```

lib/rules/require-spaces-in-for-statement.js
hzoo: How are we solving this for all other rules - maybe that should be part of the whitespaceBetween method? I don't know how common it is to have comments in for loop statements

mikesherov: Right, only certain rules where it's common to have comment blocks in between (such as function arguments) should we consider them. But again, I think it's beyond scope of this pull request. @zxqfox, file a new issue for that please.

lib/rules/disallow-space-between-arguments.js
```diff
@@ -46,20 +46,13 @@ module.exports.prototype = {
...
                 if (punctuatorToken.value === ',') {
                     errors.assert.noWhitespaceBetween({
                         token: punctuatorToken,
-                        nextToken: file.getNextToken(punctuatorToken),
-                        message: 'Illegal space between arguments'
```
mikesherov: why is this removed?

lib/rules/disallow-space-between-arguments.js
mikesherov: why isn't this line just: `var punctuatorToken = file.getFirstNodeToken(param);` previously it was `var punctuatorToken = file.getTokenByRangeStart(param.range[1]);`?

mikesherov: never mind, I get it.

test/rules/require-curly-braces.js
```diff
@@ -153,4 +153,14 @@ describe('rules/require-curly-braces', function() {
         assert(error.line === 4);
         assert(error.column === 12);
     });
+
+    it('should not report missing `else` braces for `else if`', function() {
+        checker.configure({ requireCurlyBraces: ['else'] });
+        assert(checker.checkString('if (x) { x++; } else if (x) x++;').isEmpty());
```
mikesherov: please change this to: `assert(checker.checkString('if (x) { x++; } else if (x) { x++; }').isEmpty());` it is misleading otherwise.

test/rules/require-curly-braces.js
```diff
@@ -153,4 +153,14 @@ describe('rules/require-curly-braces', function() {
...
+    });
+
+    it('should report missing `if` braces for `else if`', function() {
```
mikesherov: remove this test, it's not needed.

mikesherov:
Thanks again @hzoo for all your hard work here.

lib/rules/disallow-space-between-arguments.js
zxqfox: :+1:

lib/rules/require-spaces-in-for-statement.js
zxqfox: @hzoo I've thought about it, but better no ;-) We did this: [getCommentAfterToken](https://github.com/jscs-dev/node-jscs/search?utf8=%E2%9C%93&q=getCommentAfterToken&type=Code) — helpers to get all comments before and after token.

→ #987

lib/rules/disallow-space-between-arguments.js
hzoo: Wasn't sure if we wanted to just use the default messages or not

hzoo:
Made the suggested changes

mikesherov:
@hzoo, please rebase. I've already changed the callExpressions rules as part of a code coverage sweep.

hzoo:
Ok rebased and changed the last commit (requireSpaces) to match the disallow one

mikesherov:
Thanks again!

PR #990, opened 2015-02-02
JsFile: fix duplicate `(` token in `export default` statements

mikesherov:


mrjoelkemp:
:+1:

lib/js-file.js
```diff
@@ -139,6 +139,22 @@ JsFile.prototype = {
...
+        // in duplicate tokens on `export default function() {}`
+        // (https://code.google.com/p/esprima/issues/detail?id=631)
+        if (this.getDialect() === 'es6') {
+            var lastToken = this._tree.tokens[0];
+            this._tree.tokens = this._tree.tokens.filter(function(token, i) {
+                if (i === 0) {
+                    return true;
+                }
+
+                var repeatedToken = lastToken.range[0] === token.range[0] && lastToken.range[1] === token.range[1];
```
markelog: A bit dangerous, but i can't think of the case when ranges would overlap.

lib/js-file.js
```diff
@@ -139,6 +139,22 @@ JsFile.prototype = {
      * Builds token index by starting pos for futher navigation.
      */
     _buildTokenIndex: function() {
+        // Temporary fix (i hope) for esprima tokenizer, which results
+        // in duplicate tokens on `export default function() {}`
+        // (https://code.google.com/p/esprima/issues/detail?id=631)
+        if (this.getDialect() === 'es6') {
+            var lastToken = this._tree.tokens[0];
```
markelog: Would move it to the filter function, but i guess it's developer choice

lib/js-file.js
markelog: This line looks like an `if` statement then as assignment considering how it used below

mikesherov: Yeah, tokens can't overlap. Nodes can. It's safe. Will make your other suggestions.

Issue #901, opened 2015-01-12
move esprima parsing to inside JsFile and out of StringChecker


mikesherov:
the `tree` variable is created in string checker by parsing the `str` variable. The `str` and `tree` variables are then passed to the `JsFile` constuctor. The tree is not used anywhere else.

We should just parse in the constructor of JsFile given the `str`. This will also help us avoid having to do `esprima.parse` directly in tests.

I don't think this need to move to 2.0 I don't imagine anyone is really using JsFile directly. Thoughts?

hzoo:
This probably isn't as relevant anymore - especially after the cst changes right?

hzoo:
Ok closing

mdevils:
I have a prototype in `v3` where it was implemented, so feel free to close.

PR #1012, opened 2015-02-07
Autofixing: token data

mdevils:
Getting ready for JS-code reassembly.

In this commit I:
* Merged code tokens with comment tokens.
* Prepared tokens for future reassembling into JS-source.
* Applied whitespace info to the tokens.
* Refactored `JsFile` class to increase simplicity.

/cc @mikesherov @markelog @mrjoelkemp @zxqfox

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -18,59 +18,38 @@ var JsFile = function(filename, source, tree, options) {
 
     this._filename = filename;
     this._source = source;
-    this._tree = tree || {tokens: []};
+    this._tree = tree || {tokens: [], comments: []};
 
     this._es3 = options.es3 || false;
     this._es6 = options.es6 || false;
 
     this._lines = source.split(/\r\n|\r|\n/);
-    this._tokenRangeStartIndex = null;
-    this._tokenRangeEndIndex = null;
-    var index = this._index = {};
-    var _this = this;
 
-    this._buildTokenIndex();
+    this._tree.tokens = this._fixEs6Tokens(this._tree.tokens);
+    this._tokens = this._buildTokenList(this._tree.tokens, this._tree.comments);
+    this._addEOFToken();
+    this._applyWhitespaceData(this._tokens, source);
 
-    this.iterate(function(node, parentNode, parentCollection) {
-        var type = node.type;
+    var tokenIndexes = this._buildTokenIndex(this._tokens);
+    this._tokenRangeStartIndex = tokenIndexes.tokenRangeStartIndex;
+    this._tokenRangeEndIndex = tokenIndexes.tokenRangeEndIndex;
 
-        node.parentNode = parentNode;
-        node.parentCollection = parentCollection;
-        (index[type] || (index[type] = [])).push(node);
-
-        // Temporary fix (i hope) for esprima tokenizer
-        // (https://code.google.com/p/esprima/issues/detail?id=481)
-        // Fixes #83, #180
-        switch (type) {
-            case 'Property':
-                convertKeywordToIdentifierIfRequired(node.key);
-                break;
-            case 'MemberExpression':
-                convertKeywordToIdentifierIfRequired(node.property);
-                break;
-        }
-    });
+    this._index = this._buildNodeIndex();
+    this._fixEsprimaIdentifiers();
 
     this._buildDisabledRuleIndex();
-
-    // Part of temporary esprima fix.
-    function convertKeywordToIdentifierIfRequired(node) {
-        var tokenPos = _this.getTokenPosByRangeStart(node.range[0]);
-        var token = _this._tree.tokens[tokenPos];
-        if (token.type === 'Keyword') {
-            token.type = 'Identifier';
-        }
-    }
 };
 
 JsFile.prototype = {
     /**
      * Builds an index of disabled rules by starting line for error suppression.
+     *
+     * @private
      */
     _buildDisabledRuleIndex: function() {
         this._disabledRuleIndex = [];
 
-        var comments = this.getComments() || [];
+        var comments = this.getComments();
         var commentRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
 
         comments.forEach(function(comment) {
@@ -116,6 +95,7 @@ JsFile.prototype = {
      * @param {Boolean} enabled whether the rule is disabled or enabled on this line
      * @param {String} rulesStr the string containing specific rules to en/disable
      * @param {Number} line the line the comment appears on
+     * @private
      */
     _addToDisabledRuleIndex: function(enabled, rulesStr, line) {
         rulesStr = rulesStr || '*';
@@ -137,24 +117,12 @@ JsFile.prototype = {
 
     /**
      * Builds token index by starting pos for futher navigation.
+     *
+     * @param {Object[]} tokens
+     * @returns {{tokenRangeStartIndex: {}, tokenRangeEndIndex: {}}}
+     * @private
      */
-    _buildTokenIndex: function() {
-        // Temporary fix (i hope) for esprima tokenizer, which results
-        // in duplicate tokens on `export default function() {}`
-        // (https://code.google.com/p/esprima/issues/detail?id=631)
-        if (this.getDialect() === 'es6') {
-            var tokenHash = {};
-            this._tree.tokens = this._tree.tokens.filter(function(token) {
-                var hashKey = token.range[0] + '_' + token.range[1];
-                var isDuplicate = tokenHash[hashKey];
-
-                tokenHash[hashKey] = true;
-
-                return !isDuplicate;
-            });
-        }
-
-        var tokens = this._tree.tokens;
+    _buildTokenIndex: function(tokens) {
         var tokenRangeStartIndex = {};
         var tokenRangeEndIndex = {};
         for (var i = 0, l = tokens.length; i < l; i++) {
@@ -162,52 +130,22 @@ JsFile.prototype = {
             tokenRangeEndIndex[tokens[i].range[1]] = i;
             tokens[i]._tokenIndex = i;
         }
-        this._tokenRangeStartIndex = tokenRangeStartIndex;
-        this._tokenRangeEndIndex = tokenRangeEndIndex;
+        return {
+            tokenRangeStartIndex: tokenRangeStartIndex,
+            tokenRangeEndIndex: tokenRangeEndIndex
+        };
     },
-    /**
-     * Builds comments index by starting pos for futher navigation.
-     */
-    _buildCommentIndex: function() {
-        var comments = this.getComments();
-        var tokens = this.getTokens();
-        var tokenIndex = 0;
-        var tokensLength = tokens.length;
-        var tokenCommentsBeforeIndex = this._tokenCommentsBeforeIndex = {};
-        var tokenCommentsAfterIndex = this._tokenCommentsAfterIndex = {};
-        var partialComments = [];
-        for (var i = 0, l = comments.length; i < l && comments[i].range[1] <= tokens[0].range[0]; i++) {
-            partialComments.push(comments[i]);
-        }
-        if (partialComments.length) {
-            tokenCommentsBeforeIndex[0] = partialComments;
-            partialComments = [];
-        }
-        for (; i < l; i++) {
-            var startPos = comments[i].range[0];
-            if (partialComments.length) {
-                tokenCommentsAfterIndex[tokenIndex] = partialComments;
-                tokenCommentsBeforeIndex[tokenIndex + 1] = partialComments;
-                partialComments = [];
-            }
-            while (tokenIndex + 1 < tokensLength && tokens[tokenIndex + 1].range[1] <= startPos) {
-                tokenIndex++;
-            }
-            partialComments.push(comments[i]);
-        }
-        if (partialComments.length) {
-            tokenCommentsAfterIndex[tokenIndex] = partialComments;
-            tokenCommentsBeforeIndex[tokenIndex + 1] = partialComments;
-        }
-    },
+
     /**
      * Returns token position using range start from the index.
      *
+     * @deprecated
      * @returns {Object}
      */
     getTokenPosByRangeStart: function(start) {
         return this._tokenRangeStartIndex[start];
     },
+
     /**
      * Returns token using range start from the index.
      *
@@ -215,8 +153,9 @@ JsFile.prototype = {
      */
     getTokenByRangeStart: function(start) {
         var tokenIndex = this._tokenRangeStartIndex[start];
-        return tokenIndex === undefined ? undefined : this._tree.tokens[tokenIndex];
+        return tokenIndex === undefined ? undefined : this._tokens[tokenIndex];
     },
+
     /**
      * Returns token using range end from the index.
      *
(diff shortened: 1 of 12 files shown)


lib/js-file.js
```diff
@@ -425,12 +425,10 @@ JsFile.prototype = {
      * @returns {Object|undefined}
      */
     getCommentAfterToken: function(token) {
```
mikesherov: This likely becomes deprecated now, as its usage is superceded by getNextToken(true). We can fix that later though.

lib/js-file.js
```diff
@@ -438,12 +436,10 @@ JsFile.prototype = {
      * @returns {Object|undefined}
      */
     getCommentBeforeToken: function(token) {
```
mikesherov: This also is likely deprecated.

lib/js-file.js
```diff
@@ -543,7 +546,196 @@ JsFile.prototype = {
...
+        return result;
+    },
+
+    /**
+     * Adds `_isComment` flat to the comment-tokens.
+     *
+     * @param {Object[]} commentTokens
+     * @private
+     */
+    _markCommentTokens: function(commentTokens) {
```
mikesherov: Possibly unnecessary if comment tokens already have a "type".

mdevils: There are two types: `Block` and `Line`

lib/js-file.js
```diff
@@ -543,7 +546,196 @@ JsFile.prototype = {
...
+    },
+
+    /**
+     * Applies whitespace information to the token list.
+     *
+     * @param {Object[]} tokens
+     * @param {String} source
+     * @private
+     */
+    _applyWhitespaceData: function(tokens, source) {
```
mikesherov: Should be `whitespaceBefore` and need to introduce an `EOF` token at end. Otherwise, whitespace before first real comment is missed.

lib/js-file.js
```diff
@@ -543,7 +546,196 @@ JsFile.prototype = {
...
+    /**
+     * Temporary fix (I hope) for esprima tokenizer, which results
+     * in duplicate tokens on `export default function() {}`
+     * (https://code.google.com/p/esprima/issues/detail?id=631)
+     *
+     * @param {Object[]} tokens
+     * @returns {Object[]}
+     * @private
+     */
+    _fixEs6Tokens: function(tokens) {
```
mikesherov: The fix for this will be in next Esprima release :-)

lib/js-file.js
```diff
@@ -543,7 +546,196 @@ JsFile.prototype = {
...
+
+/**
+ * Parses a JS-file.
+ *
+ * @param {String} source
+ * @param {Object} esprima
+ * @param {Object} esprimaOptions
+ * @returns {Object}
+ */
+JsFile.parse = function(source, esprima, esprimaOptions) {
```
mikesherov: Hooray! There's an open ticket for this btw.

mdevils: Found it #901 :)

lib/js-file.js
mdevils: Oh, yes, empty files with whitespaces will be truncated. Will add EOF token.

lib/js-file.js
mikesherov: Also, can clean up the JsFile tests now that JsFile has a parse function.

lib/js-file.js
mikesherov: I see. OK, then. Perhaps do this step inside buildTokenList so that an additional loop can be saved?

lib/rules/disallow-padding-newlines-in-blocks.js
```diff
@@ -51,7 +51,6 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         var lines = file.getLines();
-        var tokens = file.getTokens();
```
mikesherov: Perhaps we need to add jshint's `unused` option if this was just dangling.

lib/rules/disallow-semicolons.js
```diff
@@ -43,15 +43,17 @@ module.exports.prototype = {
...
-                    errors.add('semicolons are disallowed at the end of a line.', token.loc.end);
-                }
+        file.iterateTokensByTypeAndValue('Punctuator', ';', function(token) {
+            errors.assert.sameLine({
+                token: token,
+                nextToken: file.getNextToken(token),
+                requireNextToken: true,
+                message: 'Semicolons are disallowed at the end of a line.'
             });
+            //var nextToken = file.getNextToken(token);
```
mikesherov: obvious, but remove this.

lib/string-checker.js
```diff
@@ -83,17 +83,6 @@ StringChecker.prototype = {
      * @throws Error on invalid source code.
      */
     _parse: function(str) {
```
mikesherov: hmmm, perhaps instead of a static call to `JsFile.parse`, we simple move the parse call into the constructor of JsFile, along with passing `this._esprima` and `this._esprima` as options to JsFile?

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
...
+        if (index >= 0) {
+            if (includeComments) {
+                return this._tokens[index];
+            } else {
+                do {
+                    if (!this._tokens[index]._isComment) {
+                        return this._tokens[index];
+                    }
+                    index--;
+                } while (index >= 0);
```
mikesherov: `} while (--index >= 0);`

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
...
-        return index >= 0 ? this._tree.tokens[index] : undefined;
+        if (index >= 0) {
+            if (includeComments) {
+                return this._tokens[index];
+            } else {
+                do {
+                    if (!this._tokens[index]._isComment) {
+                        return this._tokens[index];
+                    }
+                    index--;
```
mikesherov: remove line.

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
...
      *
      * @param {Object} token
+     * @param {Boolean} [includeComments=false]
      * @returns {Object|undefined}
      */
-    getPrevToken: function(token) {
+    getPrevToken: function(token, includeComments) {
         var index = token._tokenIndex - 1;
-        return index >= 0 ? this._tree.tokens[index] : undefined;
+        if (index >= 0) {
```
mikesherov: perhaps gaurd clause instead to avoid 4 levels of indent below?

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
...
+            if (includeComments) {
+                return this._tokens[index];
+            } else {
```
mikesherov: no need for else if the line before is early returning.

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
...
      *
      * @param {Object} token
+     * @param {Boolean} [includeComments=false]
      * @returns {Object|undefined}
      */
-    getNextToken: function(token) {
+    getNextToken: function(token, includeComments) {
         var index = token._tokenIndex + 1;
-        return index < this._tree.tokens.length ? this._tree.tokens[index] : undefined;
+        if (index < this._tokens.length) {
```
mikesherov: same comments here as function above.

lib/js-file.js
```diff
@@ -543,7 +546,196 @@ JsFile.prototype = {
...
+            var tokenHash = {};
+            return tokens.filter(function(token) {
+                var hashKey = token.range[0] + '_' + token.range[1];
+                var isDuplicate = tokenHash[hashKey];
+
+                tokenHash[hashKey] = true;
+
+                return !isDuplicate;
+            });
+        } else {
```
mikesherov: perhaps make this a guard clause at top instead?

lib/rules/disallow-semicolons.js
mdevils: Oh, sure!

lib/string-checker.js
mdevils: The problem is that StringChecker catches parse errors. If we would parse in JsFile constructor, we won't be able to create new JsFile instance. Currently JsFile is created with empty esprima node tree information.

mikesherov: OK, we can talk about this another time because it's actually not relevant to the rest of the work you've done here, and I don't want to take attention away from it.

lib/js-file.js
mdevils: OK, thank you!

lib/js-file.js
```diff
@@ -245,26 +184,55 @@ JsFile.prototype = {
     getLastNodeToken: function(node) {
         return this.getTokenByRangeEnd(node.range[1]);
     },
+
     /**
      * Returns the first token before the given.
      *
      * @param {Object} token
+     * @param {Boolean} [includeComments=false]
```
TheSavior: I don't think we should have magic booleans. I think I'd much prefer this taking an options object.

mikesherov: Yeah, @ariya would frown on our use of the boolean trap. However, this is more for pragmatism than anything else at the moment.

lib/rules/disallow-padding-newlines-in-blocks.js
mdevils: Yep, will PR this separately.

lib/js-file.js
mdevils: Agreed.

mdevils:
I have added `JsFile::render()` method and several tests for it. Seems like it works.

lib/js-file.js
```diff
@@ -543,7 +556,236 @@ JsFile.prototype = {
...
+     */
+    _addEOFToken: function() {
+        var loc = {
+            line: this._lines.length,
+            column: this._lines[this._lines.length - 1].length
+        };
+        this._tokens.push({
+            type: 'EOF',
+            value: '',
+            range: [this._source.length, this._source.length + 1],
```
mikesherov: I suppose technically, the range of this should be: `range: [this._source.length, this._source.length],` zero length.

lib/rules/disallow-padding-newlines-in-blocks.js
mikesherov: already on it :-)

lib/string-checker.js
```diff
@@ -83,17 +83,6 @@ StringChecker.prototype = {
...
-
-        // Convert bin annotation to a comment
-        if (hashbang) {
-            str = '//' + str.substr(2);
-        }
-
-        // Strip special case code like iOS instrumentation imports: `#import 'abc.js';`
-        str = str.replace(/^#!?[^\n]+\n/gm, '');
-
         var esprimaOptions = {
```
mikesherov: I suppose this, and the best of the function body here, should move into in `JsFile.parse` too. You're doing one too many copies of the options.

lib/js-file.js
mdevils: I've added +1 for `tokenEndPosition` indexes... But I can just ignore EOF there...

lib/string-checker.js
mdevils: Yeah, though so.

lib/js-file.js
mikesherov: It's probably fine either way.

mikesherov:
@mdevils this is probably good to go after you move the last bits out of `_parse` and into `JsFile.parse`. I'd like for you to land it, so we can move onto the next step. :-)

mdevils:
I think this is almost ready to merge. I plan adding some test.

TheSavior:
Awesome, looking forward to being able to fix all the rules that do this:

`var nextToken = file.getCommentAfterToken(token) || file.getNextToken(token);`

mikesherov:
@mdevils please rebase. I just finished up the JSHint `unused` ticket, which causes conflicts.

lib/token-assert.js
```diff
@@ -77,11 +77,21 @@ TokenAssert.prototype.noWhitespaceBetween = function(options) {
...
  * @param {String} [options.message]
+ * @param {Boolean} [options.requireNextToken=false]
  */
 TokenAssert.prototype.sameLine = function(options) {
     var token = options.token;
     var nextToken = options.nextToken;
 
+    if (options.requireNextToken && (!nextToken || nextToken.type === 'EOF')) {
+        this.emit('error', {
+            message: options.message || token.value + ' should not be the last in the file',
```
TheSavior: It seems weird to be messaging about the end of the file in a generic helper like this.

mdevils: I'm not sure if it is weird.

mdevils:
I have rebased this PR.

lib/js-file.js
```diff
@@ -244,26 +185,60 @@ JsFile.prototype = {
...
+
+        if (options && options.includeComments) {
+            return this._tokens[index];
+        }
+
+        do {
+            if (!this._tokens[index].isComment) {
+                return this._tokens[index];
+            }
+        } while (++index < this._tokens.length);
```
zxqfox: Probably it's groundlessly slow because we can put `_nextTokenIndex` value on building index. `getNext/PrevToken` methods are very popular

lib/js-file.js
```diff
@@ -542,7 +554,238 @@ JsFile.prototype = {
...
+     * @returns {String}
+     */
+    render: function() {
+        var result = '';
+
+        // For-loop for maximal speed.
+        for (var i = 0; i < this._tokens.length; i++) {
+            var token = this._tokens[i];
+
+            result += token.whitespaceBefore;
```
zxqfox: Should we use here `result.push` + `result.join('')` for speed?

lib/js-file.js
mikesherov: @zxqfox, if we want to actually go the performance route, we should add some perf tests and establish a perf budget. Otherwise, we're just guessing that this is slow or an actual bottleneck.

lib/js-file.js
mikesherov: http://jsperf.com/string-concat-vs-array-join-10000/15 suggests that this doesn't matter.

lib/js-file.js
zxqfox: @mikesherov At least we can run tests before and after that PR.
I know there are a couple of additional tests but...
Executing `time mocha` in `master`:
```
real    0m13.345s
user    0m12.157s
sys     0m0.304s
```
And in `mdevils.autofixing-token-data`:
```
real    0m21.480s
user    0m19.745s
sys     0m0.528s
```
It's like 1.5x slower ;-(

mikesherov: how many times did you run this?

zxqfox: Once, but... Oh. I can't repeat that. Sorry.

mikesherov: I just ran the same thing you did and got:

master
```
real	0m9.353s
user	0m8.401s
sys	0m0.313s
```


jscs-dev-mdevils.autofixing-token-data
```
real	0m10.031s
user	0m8.752s
sys	0m0.415s
```

Which is a negligible difference. But then again, BOTH of our results are invalid because they are not rigorous done by a perf testing framework, which exist and we should integrate if we care about this sort of thing.

mikesherov: Furthermore, we have no idea where the bottleneck is in this case. I seriously doubt it has anything to do with this line.

mikesherov: BTW, no need to apologize! We *should* integrate perf testing.

zxqfox: > are not rigorous done by a perf testing framework, which exist and we should integrate

Looking into v8.log, what you suggest as perf testing framework?

mikesherov: Let's move this discussion to a new issue.

lib/js-file.js
mdevils: Yep, strings perform good these times :)

Issue #516, opened 2014-07-08
Auto formatting API and option. [$315]


mikesherov:
JSCS should have the ability to automatically format a given file with the specified rules. There are several challenges here:

For now, we are only going to consider whitespace changes. This approach is outlined here:

1. Change all rules to use the JsFile traversal methods instead of directly referencing the token by index directly. This allows us to change the underlying token data structure at will while only having to change the traversal methods.
2. Make all whitespace only rules use the token-assert framework.
3. Modify the token list so that it includes comment tokens, and a new terminal token called "EOF".
3a. Modify traversal methods to ignore comment tokens as necessary.
4. Modify comment tokens to include a property called "prefix", which contains the preceding whitespace.
5. Add a function called errors.fixToken (or some other name), which takes in a token, and the updated prefix whitespace, which modifies the prefix of that token.
6. Make token-assert call errors.fixToken appropriately.
7. Modify StringChecker so that it can output the modifed source and save it to disk. It's simply a loop over the token list, printing the prefix followed by the value of the token.
8. Modify CLI to respect a `--fix` parameter, which triggers the autoformat option.

Please see https://github.com/mdevils/node-jscs/pull/514 as one possible implementation that ultimately wasn't how we wanted to approach the problem.

jzaefferer:
So #751 was merged, introducing the infrastructure for autofixing, but not the autofixing itself, is that right? Does someone have plans to work on the next step? If not, would a bigger bounty make a difference?

mikesherov:
Yes, it's the next thing im working on. Bounty makes no difference here.

indexzero:
That's awesome @mikesherov! I didn't realize #752 was merged. Is there a running list of rules that need formatters? Now that the API is defined I could help write a few.

soliton4:
hi there
i found this in my inbox:
https://github.com/soliton4/nodeMirror/issues/18

would be a totally cool feature, but it looks like the formatting option is still under construction.
is there any part of it usable and can you point me to a working piece of code?

i basically need the option to input text and get formated text as output.

thank you very much ;D

jdlrobson:
Hey @mikesherov I'm super interested in this too. Is there any branch I should be watching where this is happening? I would be interested to see how you are thinking about this and see if I can help out!

mikesherov:
[...]

PR #1014, opened 2015-02-07
New Rule: requirePaddingNewLinesAfterBlockDeclarations

### LINKED ISSUE #1010: Add Rule requirePaddingNewLinesAfterBlockDeclarations
Any thoughts on a rule like this?

Valid:
```javascript
if (true) {
}

while (true) {
}

// and

function foo() {
  for (var i = 0; i < 2; i++) {
  }

  console.log('foo');
}

// and

var obj = {
  a: 1,
  foo: function() {
  },

  bar: function() {
  }
};

// and

if (true) {
  (function() {
  })();

  var a = 2;
}

// and

Foo.prototype.a = function() {
};

Foo.prototype.b = function() {
};
```
Invalid:

```javascript
if (true) {
  var a = 2;
}
while (true) {
  var b = 3;
}

// and

function foo() {
  for (var i = 0; i < 2; i++) {
  }
  console.log('foo');
}

// and

var obj = {
  a: 1,
  foo: function() {
  },
  bar: function() {
  }
};

// and

if (true) {
  (function() {
  })();
  var a = 2;
}

// and

Foo.prototype.a = function() {
};
Foo.prototype.b = function() {
};
```

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8368807-require-padding-new-lines-after-block-declarations?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
requirePaddingNewLinesAfterBlockDeclarations / disallowPaddingNewLinesAfterBlockDeclarations

Fixes #1010


TheSavior:
This PR updated differentLine in tokenAssert. Most of the whitespace only rules that will need to be updated to accomplish the next step in #516 will require an even more robust differentLine function. 

I don't want to update those until after this is merged, otherwise there will be conflicts or two barely related PRs relying on each other.

mikesherov:
@TheSavior thanks a bunch for contributing! The rule names should be named `(require|disallow)PaddingNewLinesAfterBlocks`. Technically, it's a block statement, and never a block declaration.

test/rules/disallow-padding-newlines-after-block-declarations.js
```diff
@@ -0,0 +1,65 @@
...
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+        checker.configure({ disallowPaddingNewLinesAfterBlockDeclarations: true });
+    });
+
+    it('should report padding after block', function() {
+        assert(checker.checkString('if(true){}\n\nvar a = 2;').getErrorCount() === 1);
+    });
+    it('should report padding after nested block', function() {
```
mikesherov: padding newline before each it. I know this is inconsistent elsewhere, but that's what we've decided on.

TheSavior:
Updated and squashed.

lib/token-assert.js
```diff
@@ -110,9 +112,22 @@ TokenAssert.prototype.differentLine = function(options) {
...
 
-    if (token.loc.end.line === nextToken.loc.start.line) {
+    var gap = options.hasOwnProperty('lines') ? options.lines : 1;
+    var allowExtra = options.hasOwnProperty('allowExtra') ? options.allowExtra : false;
+
+    var errorMessage = token.value + ' and ' + nextToken.value + ' should be ';
+    if (gap === 1) {
+        errorMessage += 'on different lines';
+    } else {
+        errorMessage += 'separated by ' + gap + ' lines';
```
zxqfox: Not `gap - 1` lines?

lib/token-assert.js
```diff
@@ -101,6 +101,8 @@ TokenAssert.prototype.sameLine = function(options) {
  * @param {Object} options.token
  * @param {Object} options.nextToken
  * @param {Object} [options.message]
+ * @param {Number} [options.lines=1] Amount of lines between tokens.
```
zxqfox: It's not a real amount of lines between. It's more like difference between line numbers. Not sure how to name that properly. Gap?

lib/rules/require-newline-before-block-statements.js
```diff
@@ -97,6 +97,7 @@ module.exports.prototype = {
             errors.assert.differentLine({
                 token: prevToken,
                 nextToken: openingBrace,
+                allowExtra: true,
```
zxqfox: Should we?

zxqfox:
Great work here! :+1:

lib/rules/require-newline-before-block-statements.js
TheSavior: Yeah, I didn't have this originally but it's tests were assuming that differentLine wasn't strict and only guaranteed that they were in fact on *different* lines, not that they were on touching lines.

lib/token-assert.js
TheSavior: :+1:

TheSavior:
Updated to fix naming / usage of 'options.lines'. I think all of these TokenAssert whitespace functions will need to be changed anyways once we update all of the whitespace rules to use them.

lib/rules/require-newline-before-block-statements.js
zxqfox: I'd add a test for that then. ;-)

TheSavior: Is there a test you can think of that I'm missing? require-newline-block-statements had tests for both and I made sure those worked. I also have tests for token-assert differentLine that tests allowing extra, as well as requiring strict. :-)

zxqfox: Ye. https://github.com/jscs-dev/node-jscs/blob/master/test/rules/require-newline-before-block-statements.js#L30 
You're right. ;-) Nothing to do already.

TheSavior: :+1:

TheSavior:
What do we need to land this?

mikesherov:
@savior, nothing. Just wait until I come through and land a bunch of these.

Although, I wonder if we need requirePaddingNewLinesBeforeBlocks too?

TheSavior:
Cool.

That rule kind of exists already, but it obviously has to be very different logic. There is this rule: requireNewLineBeforeBlockStatements

TheSavior:
I've gone ahead and taken this opportunity to set up differentLine for us to be able to use it when porting the other whitespace rules to use it. I didn't particularly want to land this with an incompatible version of differentLine.

mikesherov:
@mdevils or @markelog I'm a fan of the additions to the assertions, but just wanted to get a +1 from either of you before I land this.

TheSavior:
Rebasing from master

TheSavior:
This rule has been added to the Airbnb style guide: https://github.com/airbnb/javascript#whitespace

TheSavior:
@mdevils or @markelog bump :+1:

lib/rules/disallow-padding-newlines-after-blocks.js
```diff
@@ -0,0 +1,104 @@
...
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(value) {
+        assert(
+            typeof value === 'boolean',
+            'disallowPaddingNewLinesAfterBlocks option requires boolean value'
+        );
+        assert(
```
zxqfox: @markelog Do you remember what we decided about double assertions in configure?

lib/token-assert.js
```diff
@@ -110,12 +113,50 @@ TokenAssert.prototype.differentLine = function(options) {
...
-            message: options.message || token.value + ' and ' + nextToken.value + ' should be on different lines',
+            message: options.message || errorMessage + message + ' line(s) apart',
             line: token.loc.end.line,
             column: token.loc.end.column
         });
+    }).bind(this);
+
+    var lineDifference = Math.abs(token.loc.end.line - nextToken.loc.start.line);
+
+    if (atLeast !== undefined) {
```
zxqfox: I'm not sure about atMost, atLeast, and exactly (cc @mdevils), but if it's ok — can we make a switch-case here?
Like:
```
swtich (true) {
  case !!exactly:
    error = 'exactly' + exactly;
    break;
  case !atLeast:
    error = 'at least ' + atLeast;
    break;
  case !atLeast:
    error = 'at most ' + atMost;
    break;
  default:
    error = 'more than ' + atLeast + ' and less than ' + atMost;
    break;
}
```

lib/rules/disallow-padding-newlines-after-blocks.js
TheSavior: Nope, I copied this from another rule.

Are you telling me that this is wrong, or actually asking what it should be? haha.

zxqfox: Just asking. I'm not sure that we have a strong decision about that ;-) Half of rules has double check, and the rest just `value === true` or something.

TheSavior: value === true is probably reasonable here

zxqfox: Definitely. Just both assertions will throw if `value !== true`, and it's only possible value for rule. And also `option requires boolean value` can mislead. So why we need to assert twice? ;-) That's why we started that discussion earlier.

lib/token-assert.js
TheSavior: atLeast or exactly could be 0 in which case !0 wouldn't really give us the logic we are looking for. I agree that this set of if/elses could be cleaned up, I'm not convinced a `switch(true)` is the best way to do it. :-)

TheSavior: Also, I'm not completely convinced on `atLeast`, `atMost`, and `exactly` as options either. The other option is to have multiple functions. Any opinions?

zxqfox: Oh, right. Probably not but it allows you to make a flat human readable tree of conditions and will report an error just one for one free lines block. So it's a nice option imo ;-)

Anyway, this change affects assertion API and we should wait @mdevils with his minds about autoformatter.

TheSavior:
@mdevils or @markelog bump :+1: 

We're still looking for some more concrete thoughts about the change to the differentLine assertion here.

mdevils:
I'd add a new assertion method `blankLinesBetween`. What do you think about it?

zxqfox:
@mdevils It's totally fine if it'd make autofix easier. For now I feel that assertion methods should be as simple as possible so new method should be better than extended old one.

TheSavior:
So we will have sameLine, differentLine, and blankLinesBetween. BlankLinesBetween will take a number of lines and lessThan, greaterThan, and exactly. Are we good with that? 

There is definitely some cases where you want to say that there should be exactly one line between them. disallowMultipleLineBreaks(3) is an example that would want to say at most 3 line breaks, and then most other cases would want greaterThan so that it doesn't conflict with disallowMultipleLineBreak.

mikesherov:
`BlankLinesBetween` should be able to simultaneously take `lessThan` and `greaterThan`, but should be mutually exclusive of `exactly`. 

@TheSavior sounds like you got what you need. Once you finish up here I can land this, and we can move on to #1023 :-)

TheSavior:
Agreed that it should be mutually exclusive with exactly. I'll plan on throwing an Error if exactly and (lessThan or greaterThan) are set.

I should be able to get to this in the next two days. I'm really excited to move forward on these PRs.

mikesherov:
@TheSavior thanks again for sticking with this!

TheSavior:
To be consistent with `whitespaceBetween`, I'm going to call it `linesBetween` instead of `blankLinesBetween`. Any issues with that?

mikesherov:
Go ahead!

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -0,0 +1,128 @@
...
+ *             return 2;
+ *         }
+ *     };
+ * }
+ * ```
+ */
+
+var assert = require('assert');
+
+var excludes = [
```
mikesherov: This should be an object instead of an array.

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -0,0 +1,128 @@
...
+            var parentNode = node.parentNode;
+
+            var nextToken = file.getNextToken(closingBracket);
+
+            function tokenShouldBeExcluded(item) {
+                return parentNode.type === item[0] && nextToken.value === item[1];
+            }
+
+            while (nextToken !== undefined) {
+                var excludedStatement = excludes.some(tokenShouldBeExcluded);
```
mikesherov: If excludes is an object, this becomes:

`if (excludes[parentNode] === nextToken.value) {`

test/token-assert.js
```diff
@@ -370,6 +386,252 @@ describe('modules/token-assert', function() {
...
+
+                var tokens = file.getTokens();
+                tokenAssert.differentLine({
+                    token: tokens[1],
+                    nextToken: tokens[0],
+                    exactly: 1
+                });
+
+                assert(!onError.calledOnce);
+            });
```
mikesherov: `!onError.called` here and everywhere you're using once to test that it wasn't called.

TheSavior: :+1:

TheSavior:
@mdevils @mikesherov, can you guys take a look at the new assertion? I want to make sure that is something we agree on before I go through the merge conflicts (adding whitespace fixing to the assertions)

lib/rules/require-padding-newlines-after-blocks.js
TheSavior: Since TryStatement is in there twice, if I make it into an object I would have to have this:

```
var excludes = {
    'IfStatement': ['else'],
    'DoWhileStatement': ['while'],
    'TryStatement': ['catch', 'finally'],
    'CatchClause': ['finally'],
    'FunctionExpression': ['.'],
};
```

and then my check would be:
```
if (excludes[parentNode.type].indexOf(nextToken.value) !== -1) {
                    return;
                }
```

I'm okay with this, are you?

mikesherov: Yes.

TheSavior:
@mdevils There are no tests for the functionality of adding the whitespace prefixed data in the assertions. Since this required changing a bunch of that logic for the differentLine assertions and then dealing with that part of the merge conflicts, this PR currently doesn't have autofixing for lines. 

Can you add tests (at least 1 or two) in the style that you would want to see those tests for token-assert and then we can get these assertions adding whitespace properly.

lib/token-assert.js
```diff
@@ -102,19 +103,16 @@ TokenAssert.prototype.sameLine = function(options) {
...
-            line: token.loc.end.line,
-            column: token.loc.end.column,
-            fixed: true
-        });
-    }
+    this.linesBetween({
+        token: token,
+        nextToken: nextToken,
+        message: options.message || token.value + ' and ' + nextToken.value + ' should be on the same line',
+        exactly: -1
```
mikesherov: why isn't this `exactly: 0`?

lib/token-assert.js
```diff
@@ -93,7 +93,8 @@ TokenAssert.prototype.sameLine = function(options) {
         this.emit('error', {
             message: options.message || token.value + ' should not be the last in the file',
             line: token.loc.end.line,
-            column: token.loc.end.column
+            column: token.loc.end.column,
+            fixed: true
```
mikesherov: how is this fixed?

lib/token-assert.js
```diff
@@ -128,14 +126,73 @@ TokenAssert.prototype.differentLine = function(options) {
         return;
     }
 
-    if (token.loc.end.line === nextToken.loc.start.line) {
-        nextToken.whitespaceBefore = '\n';
+    this.linesBetween({
+        token: token,
+        nextToken: nextToken,
+        message: options.message || token.value + ' and ' + nextToken.value + ' should be on different lines',
+        atLeast: 0
```
mikesherov: this should be `atLeast: 1`

lib/token-assert.js
```diff
@@ -128,14 +126,73 @@ TokenAssert.prototype.differentLine = function(options) {
...
+
+    if (exactly !== undefined && (atLeast !== undefined || atMost !== undefined)) {
+        throw new Error('You cannot specify atLeast or atMost with exactly');
+    }
+
+    if (atLeast !== undefined && atMost !== undefined && atMost < atLeast) {
+        throw new Error('atLeast and atMost are in conflict');
+    }
+
+    var blankLines = Math.abs(token.loc.end.line - nextToken.loc.start.line) - 1;
```
mikesherov: you shouldn't have to subtract 1 here, and then the other comments I made will make sense.

TheSavior: It wouldn't be `linesBetween` then, it would be `lineDifference`...right?

lib/token-assert.js
TheSavior: merge conflict. I'll fix.

lib/token-assert.js
mikesherov: It would still be linesBetween. It's impossible for there to be `-1` lines between.

mikesherov:
@TheSavior I'm going to land this with the slight modifications I suggested so we can move forward on the rest of the great work you did.

mikesherov:
@TheSavior landed thanks!

lib/token-assert.js
TheSavior: FWIW, I don't think I agree with this.

```javascript
var x = 2;

var y = 3;
```

I think if we asked people how many lines are between those two lines, the answer we would get is 1. However, if we asked how many new line characters there are, I'm sure we'd get two. I do agree though that specifying -1 linesBetween is weird for sameLine, but I believe it makes more sense for everything else.

mikesherov: I would say two, because my editor would say:
```
1) var x = 2;
2)
3) var y = 3;
```

If the `-1` case is weird, the 0 case is just as weird.

zxqfox: For me the last case is "1 line between" or "2 line breaks between" or "2 new lines between".
Actually, I'm agree that `linesBetween` is more correct here if we apply -1.

This discuss about 0-based and 1-based numberings. I'd say that 0-based for measurement, while 1-based is for counting. So we should decide what we need here more. If we need to count lines between 2 other lines — we should use 1-based numbering.
© Your Captain Obvious ;-)

zxqfox: Personally I prefer counting (because I'm a human) so `"linesBetween": 1` is preferable for the last case.

TheSavior: how do you feel about `whiteSpaceBetween` with `spaces: 2`?

Would it be one or two space characters between the tokens?
`var x = 2` or `var  x  =  2`

lib/rules/disallow-padding-newlines-after-blocks.js
zxqfox: Btw, drop the first assertion if you can ;-) It's ok if not. I'll do it at weekend for all rules.

lib/token-assert.js
zxqfox: I think it's a special case we shouldn't check here. Probably `sameLine: true` would be more informative. Or just a separate method to check that. Or rename of `linesBetween` to `newLinesBetween` ;-(

mikesherov: this is `sameLine` :-) ! It's using linesBetween under the hood.

lib/token-assert.js
hzoo: So for `disallowPaddingNewlinesInBlocks` I'm using `linesBetween`.
Looks like I need `atMost: 1,` to make the tests pass. At first I used `exactly: 0` but I was confusing myself or something

mikesherov: yes, `atMost: 1` there can be either 1 or 0 newline characters between the two tokens.

PR #1015, opened 2015-02-07
Preset: JSCS

mrjoelkemp:
Useful for standardizing styles across all JSCS org projects.

Also useful if we want to eventually support the JSCS preset as the default.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,18 +1,6 @@
 {
-    "preset": "google",
+    "preset": "jscs",
     "fileExtensions": [ ".js", "jscs" ],
-
-    "requireParenthesesAroundIIFE": true,
-    "maximumLineLength": 120,
-    "validateLineBreaks": "LF",
-    "validateIndentation": 4,
-
-    "disallowKeywords": ["with"],
-    "disallowSpacesInsideObjectBrackets": null,
-    "disallowImplicitTypeConversion": ["string"],
-
-    "safeContextKeyword": "_this",
-
     "excludeFiles": [
         "test/data/**"
     ]
--- a/presets/jscs.json
+++ b/presets/jscs.json
@@ -1,0 +1,14 @@
+{
+    "preset": "google",
+
+    "requireParenthesesAroundIIFE": true,
+    "maximumLineLength": 120,
+    "validateLineBreaks": "LF",
+    "validateIndentation": 4,
+
+    "disallowKeywords": ["with"],
+    "disallowSpacesInsideObjectBrackets": null,
+    "disallowImplicitTypeConversion": ["string"],
+
+    "safeContextKeyword": "_this"
+}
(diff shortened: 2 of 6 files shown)


.jscsrc
```diff
@@ -1,19 +0,0 @@
-{
```
zxqfox: Symlink?

mrjoelkemp: Yes. Works well. Feel free to +1 if this looks good.

zxqfox: I'm afraid that it won't work on windows. I'd suggest here to use `{ "preset": "jscs" }`. ;-)

mrjoelkemp: Great tip. Will try it out.

zxqfox:
:+1: Nicely done!

presets/jscs.json
```diff
@@ -0,0 +1,19 @@
...
+    "validateLineBreaks": "LF",
+    "validateIndentation": 4,
+
+    "disallowKeywords": ["with"],
+    "disallowSpacesInsideObjectBrackets": null,
+    "disallowImplicitTypeConversion": ["string"],
+
+    "safeContextKeyword": "_this",
+
+    "excludeFiles": [
```
mikesherov: Move this out to jscsrc

presets/jscs.json
```diff
@@ -0,0 +1,19 @@
+{
+    "preset": "google",
+    "fileExtensions": [ ".js", "jscs" ],
```
mikesherov: Move this out to jscsrc

mikesherov:
Reopening because I reverted this change for 1.11.1

presets/jscs.json
```diff
@@ -0,0 +1,14 @@
+{
+    "preset": "google",
```
TheSavior: Should this also say jscs?

mrjoelkemp: Good question, but I don't believe so. This is the definition of the jscs preset (which is an extension of the google preset).

TheSavior: Oh, gotcha. Super confusing. This isn't the name of the preset itself, this is a case of presets extending each other.

mrjoelkemp: I hear ya. Preset name registration happens here: https://github.com/jscs-dev/node-jscs/blob/master/lib/config/configuration.js#L657-L681

TheSavior: What if the rule in configuration files was extendPreset instead of preset? Just like you can extend by path. Or in some other way merge those two options.

markelog:
> Useful for standardizing styles across all JSCS org projects.

Not sure if that is needed, `grunt-jscs`, for example, uses `jquery` preset and pretty happy with the choice.

> Also useful if we want to eventually support the JSCS preset as the default.

That's undecided, i'd say, if would choose a default one, we should choose most popular one.

mrjoelkemp:
> Not sure if that is needed, grunt-jscs, for example, uses jquery preset and pretty happy with the choice.

Style uniformity across the JSCS ecosystem would be great for external contributions. For example, I wanted to build jscs-bot and thought it wouldn't be right to use my own style for a tool that would (may) eventually live in the jscs org.

zxqfox:
For now I'm synchronizing `.jscsrc` manually in `jscs-jsdoc`, so `jscs` preset would be handy. @markelog I'm not sure why you against that.

mdevils:
This `preset` is good to have :+1:

markelog:
> @markelog I'm not sure why you against that.

Because you the only one how does this. 

- "JSCS" preset doesn't have a formal and detailed definition.
- Who would use it, aside from us?

It seems the only reason why we want to add it, is only because we can. Seems like a plugin job.

zxqfox:
@markelog Can you create a repo for our preset?

markelog:
@zxqfox would be a perfect reference in the plugin/preset wiki

PR #1020, opened 2015-02-08
Added check for es6 methods (harmony)

### LINKED ISSUE #1013: Cannot read property '_tokenIndex' of undefined in v.1.11.0
Thanks for all the great work on this tool and congrats on the latest release. However, I just upgraded from 1.10.0 to 1.11.0 and ran into this error:

```
TypeError: Cannot read property '_tokenIndex' of undefined
    at Object.JsFile.getNextToken (/Users/joe/brigade/node_modules/jscs/lib/js-file.js:265:26)
    at /Users/joe/brigade/node_modules/jscs/lib/rules/require-space-before-object-values.js:53:37
    at Array.forEach (native)
    at /Users/joe/brigade/node_modules/jscs/lib/rules/require-space-before-object-values.js:43:29
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/Users/joe/brigade/node_modules/jscs/lib/js-file.js:379:42)
    at Object.module.exports.check (/Users/joe/brigade/node_modules/jscs/lib/rules/require-space-before-object-values.js:42:14)
    at null.<anonymous> (/Users/joe/brigade/node_modules/jscs/lib/string-checker.js:157:22)
    at Array.forEach (native)
    at StringChecker.checkString (/Users/joe/brigade/node_modules/jscs/lib/string-checker.js:155:35)
```

Removing this line from my .jscsrc file makes the error go away:

```
"requireSpaceBeforeObjectValues": true,
```

Here's some of my config that might be interesting?

```
  "fileExtensions": [".js", ".jsx"],
  "esnext": true,
  "esprima": "./node_modules/esprima-fb",
```

Let me know if I can provide any more info, or if there is something that I am doing wrong here.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8378622-cannot-read-property-_tokenindex-of-undefined-in-v-1-11-0?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zxqfox:
Fixes #1013

### DIFF
--- a/lib/rules/disallow-space-after-object-keys.js
+++ b/lib/rules/disallow-space-after-object-keys.js
@@ -41,7 +41,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand) {
+                if (property.shorthand || property.method) {
                     return;
                 }
 
--- a/lib/rules/disallow-space-before-object-values.js
+++ b/lib/rules/disallow-space-before-object-values.js
@@ -41,7 +41,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand) {
+                if (property.shorthand || property.method) {
                     return;
                 }
 
--- a/lib/rules/require-aligned-object-values.js
+++ b/lib/rules/require-aligned-object-values.js
@@ -67,7 +67,7 @@ module.exports.prototype = {
 
             var maxKeyEndPos = 0;
             var skip = node.properties.some(function(property, index) {
-                if (property.shorthand) {
+                if (property.shorthand || property.method) {
                     return true;
                 }
 
--- a/lib/rules/require-quoted-keys-in-objects.js
+++ b/lib/rules/require-quoted-keys-in-objects.js
@@ -45,7 +45,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(prop) {
-                if (prop.shorthand) {
+                if (prop.shorthand || prop.method) {
                     return;
                 }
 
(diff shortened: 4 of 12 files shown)


lib/rules/disallow-space-after-object-keys.js
```diff
@@ -41,7 +41,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand) {
+                if (property.shorthand || property.method) {
```
mikesherov: might be a good util: `isKeylessProperty(node)`

mikesherov:
I see 6 rules changed with only three tests updated. Please add a test to all 6 to prevent coverage regression. If you add the utils method, you'll need to test it in test/utils.js as well.

mikesherov:
LGTM otherwise.

lib/rules/disallow-space-after-object-keys.js
zxqfox: Yep. `Property`?

mrjoelkemp: We should maybe think about tree iteration yielding custom classes like an ObjectExpression class that has a list of Property instances. Each Property instance will have the isKeylessProperty method. This avoids the bag of utils. Probably an issue worthy comment?

zxqfox: I very like that idea but realization can be painful. Anyway, we can discuss about it and decide in that issue ;-) So, yep!

zxqfox:
About method in utils: I've started to add this but it looks uglier ;-( I suggest to add it as property or method to `ObjectExpression` node.

mikesherov:
@zxqfox it's a dangerous path modifying nodes. Let's just stick with what you have for now then, and we can revisit this in the future.

mikesherov:
LGTM

mikesherov:
@zxqfox, can you please make sure future commit messages match our documented format from contributing.md:

```
<rulename>: short message
<emptyline>
Long description (if useful)
<emptyline>
Closes gh-<pullRequestNumber>
Fixes #<issueNumber>
```

PR #1022, opened 2015-02-08
Auto-generate: Handle esnext option for es6 codebases

### LINKED ISSUE #1001: AutoConfigure: Make it work for an es6 codebase
From https://github.com/jscs-dev/node-jscs/pull/1022:

> ask the user which dialect?: es(3), es(5), es(6)

Then modify the generator's internal checker configuration appropriately.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8320547-autoconfigure-make-it-work-for-an-es6-codebase?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mrjoelkemp:
Fixes #1001

Manually verified that it works.

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -55,10 +55,12 @@ Presents a walkthrough that allows you to generate a JSCS configuration by
 choosing a preset and handling violated rules.
 
 ```
-jscs --auto-configure path
+jscs --auto-configure path [--esnext]
 ```
 
-`path` can be a file or directory to check the presets against
+`path` can be a file or directory to check the presets against.
+
+Include the `--esnext` flag to auto-configure an es6 codebase.
 
 ### `--config`
 Allows to define path to the config file.
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -98,7 +98,9 @@ module.exports = function(program) {
         return returnArgs;
     }
     if (program.autoConfigure) {
-        var generator = new ConfigGenerator();
+        var generator = new ConfigGenerator({
+            esnext: program.esnext
+        });
 
         generator
         .generate(program.autoConfigure)
--- a/lib/config/generator.js
+++ b/lib/config/generator.js
@@ -37,9 +37,15 @@ var prompts = [
  * JSCS Configuration Generator
  *
  * @name Generator
+ * @param {Object} options
+ * @param {Boolean} options.esnext
  */
-function Generator() {
+function Generator(options) {
+    options = options || {};
+
     this._config = {};
+
+    this._esnext = options.esnext;
 }
 
 /**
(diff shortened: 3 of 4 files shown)


mikesherov:
It might be more difficult, but perhaps you can run the code without esnext true, than run it with esnext true if it encounters a parse error, and if it then doesn't encounter a parse error, you add `esnext: true` to the `.jscsrc`?

zxqfox:
:+1: It'd be great feature.

mrjoelkemp:
Are there downsides ()to just enabling esnext by default?

mikesherov:
At some point, esnext will be the default for JSCS. That'll happen when it happens for Esprima. Until then, es5 is our default because the es6 stuff is still somewhat buggy.

@mdevils, thoughts?

mrjoelkemp:
Understood, however, I'd assume es5 codebases wouldn't be affected at all by `esnext: true` being the default for autoconfiguration. A lot of the bugs that I see reported here are from folks actually using es6 features, not es5 folks getting hit with regressions from es6's introduction.

mikesherov:
@mrjoelkemp it's about what gets put into your `.jscsrc`. I'd like to avoid adding `esnext: true` to everyone's projects just because it doesn't break anything.

I'm willing to concede the point, but I'd like to hear from rest of team.

zxqfox:
I think that unconditionally added `esnest` will confuse users. At least we should tell them about it if they don't use es6 atm. But in that way we should know about their code and can skip adding this option. So -1 for default `esnext: true`.

mrjoelkemp:
@zxqfox The primary demographic for autogenerate users is newcomers. The confusion around esnext expands to all rules.

In the ideal sense of this feature, users will never touch their jscsrc, so it's our configuration to control.

mikesherov:
@mrjoelkemp what's important to note here though is that `esnext` comes with a few caveats. 

For example, we always assume modules instead of scripts for `esnext`, which means our users don't have to specify that they're using modules. This automatically opts them into strict mode, which changes the behavior of esprima. Also, lots of things that are illegal grammar in es5 are permitted in es6. 

It's our duty to respect the fact that most JS runtime environments are not es6 by default yet, and that means, IMO, that we should follow suit and default to es5. That situation changes once master esprima changes and browsers change.

Perhaps a different way forward is to simply *ask* the user which dialect?: `es(3), es(5), es(6)`?

mdevils:
> Perhaps a different way forward is to simply ask the user which dialect?: es(3), es(5), es(6)?

I think it is a good idea.

mdevils:
I believe right now we shouldn't put esnext by default. esprima's harmony branch is far from stability right now.

mrjoelkemp:
Yes, that would be the best option. I'll close this PR and update the issue so that someone from the community could give it a shot.

Issue #598, opened 2014-08-27
Require blank line in front of comments


jzaefferer:
Once available, this would go into the jQuery preset. See also [jQuery style guide](http://contribute.jquery.org/style-guide/js/#comments).

jzaefferer:
Yay! Is the autofixer going to deal with this as well?

hzoo:
Since it uses errors.assert then it should (and it does with `errors.assert.linesBetween`)!

mikesherov:
Yup, will be autofixed.

jzaefferer:
Nice, looking forward to start using both.

PR #1028, opened 2015-02-09
New Rule: Require padding newlines before line comments

TheSavior:
Very common rule in style guides that didn't have an implementation here. This will be able to be cleaned up when we can iterate over comments just as regular nodes.

closes: #598 

This PR now relies on #1082


TheSavior:
This is part of the Airbnb style guide: https://github.com/airbnb/javascript#comments

mikesherov:
@TheSavior since comments are now part of the token list, you should be able to iterate over all comments and get the prevToken to use assertions here. If you run into troubles with that, let me know. Thanks again as always for contributing!

TheSavior:
I haven't been able to figure out what the Type of Line comments are. I have tried iterating Tokens/Nodes by type with 'Comment', 'LineComment', 'Block', and 'Line' to no avail.

I ran this:
```
file.iterateTokensByType('Punctuator', function(token) {
            // console.log(token);
            var nextToken = file.getNextToken(token, {
                includeComments: true
            });

            console.log(nextToken);
        });```

And found some next tokens in this form:

```{ type: 'Block',
  value: ' comment ',
  range: [ 11, 24 ],
  loc: { start: { line: 2, column: 0 }, end: { line: 2, column: 13 } },
  isComment: true,
  whitespaceBefore: '\n',
  _tokenIndex: 5 }

{ type: 'Line',
  value: ' comment',
  range: [ 13, 23 ],
  loc: { start: { line: 4, column: 0 }, end: { line: 4, column: 10 } },
  isComment: true,
  whitespaceBefore: '\n\n\n',
  _tokenIndex: 5 }
```

But when I iterate over type being 'Line' or 'Block', I don't get anything.

Any ideas?

zxqfox:
@TheSavior I'd make a test (spec) for that case. These tests would be useful in any way even if it works as it should (and it's not a bug).

TheSavior:
Agreed, there should be a test for it. Once we know it works we can write a test.

TheSavior:
Currently, everything appears to be iterating on `getTokens()`. There looks like there is both `getTokens` and `getComments`.

```
/**
     * Returns token list, built using esprima.
     *
     * @returns {Object[]}
     */
    getTokens: function() {
        return this._tree.tokens;
    },

    /**
     * Returns comment token list, built using esprima.
     */
    getComments: function() {
        return this._tree.comments;
    },
```

I'd like us to consider having `getTokens` iterate over both tokens and comments, and remove the `getComments` call because everything that was using getComments should be using the iterators anyways.

TheSavior:
This has been rebased on top of #1082 to get access to the new iterators and assertions.

mikesherov:
If getTokens is not iterating over comments at this point, it's a mistake. @mdevils did some work to merge the two, perhaps this missed it?

mikesherov:
@TheSavior, yeah, the intent is for getTokens to iterate over comments now too. This is, in fact, how we `render()`

TheSavior:
@mikesherov I went into a bit more detail on this (with a fix) in #1082, that is probably a better place to have this conversation. It still does need help though.

Issue #1026, opened 2015-02-09
[Feature Request] (Disallow|Require)YodaConditions for specific operators


Slayer95:
I'd like to forbid YodaConditions for <code>===</code>, but allow them for numeric comparisons, so 
that the following constructs can be <i>simultaneously</i> made invalid/valid respectively:

Invalid
``if (5 === foo) {
...
}``

Valid
``if (foo === 5) {
...
}``

Valid
``if (0 <= foo && foo < 5) {
...
}``

It's aesthetically pleasing to be able to chain comparison operators like this.
This would imply that both YodaConditions rules accept arrays of operators as legal values, like this

<code>"DisallowYodaConditions": ["==", "===", "!=", !=="]</code>

coolaj86:
Likewise I would like to require yoda conditions for `===` and `==`  because it prevents accidental assignment where comparison was intended.

However, in the case of `<`, `>`, `>=` and `<=` it's more difficult to understand and there's no safety advantage (the safety there comes from always writing it so that the condition is true to prevent unchecked vars like `null >= 3`)

masi:
If safety/securiity is a concern than IMHO Yoda is only a must only for == (beware type juggling!). But then you could always write === and !== (which linters may force upon you anyway).

markelog:
@Slayer95, i see you have a commit there, would you like to send us a PR?

hzoo:
I think it's https://github.com/jscs-dev/node-jscs/pull/1031

paulschreiber:
Is this ready to be integrated? It'd be very helpful for me.

kilianc:
+1 on this

PR #1031, opened 2015-02-09
(require|disallow)YodaConditions: add support for listing specific operators

Slayer95:
Fixes https://github.com/jscs-dev/node-jscs/issues/1026

### DIFF
--- a/lib/rules/disallow-yoda-conditions.js
+++ b/lib/rules/disallow-yoda-conditions.js
@@ -1,14 +1,23 @@
 /**
  * Requires the variable to be the left hand operator when doing a boolean comparison
  *
- * Type: `Boolean`
+ * Type: `Array` or `Boolean`
  *
- * Value: `true`
+ * Values: Array of quoted operators or `true` to disallow yoda conditions for all possible comparison operators
  *
  * #### Example
  *
  * ```js
- * "disallowYodaConditions": true
+ * "disallowYodaConditions": [
+ *     "==",
+ *     "===",
+ *     "!=",
+ *     "!==",
+ *     ">",
+ *     "<",
+ *     ">=",
+ *     "<="
+ * ]
  * ```
  *
  * ##### Valid
@@ -34,21 +43,22 @@ module.exports = function() {};
 
 module.exports.prototype = {
 
-    configure: function(options) {
+    configure: function(operators) {
+        var isTrue = operators === true;
+
         assert(
-            options === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
+            Array.isArray(operators) || isTrue,
+            this.getOptionName() + ' option requires array or true value'
         );
-        this._operatorIndex = {
-            '==': true,
-            '===': true,
-            '!=': true,
-            '!==': true,
-            '>': true,
-            '<': true,
-            '>=': true,
-            '<=': true
-        };
+
+        if (isTrue) {
+            operators = ['==', '===', '!=', '!==', '>', '<', '>=', '<='];
+        }
+
+        this._operatorIndex = {};
+        for (var i = 0, l = operators.length; i < l; i++) {
+            this._operatorIndex[operators[i]] = true;
+        }
     },
 
     getOptionName: function() {
--- a/test/specs/rules/require-yoda-conditions.js
+++ b/test/specs/rules/require-yoda-conditions.js
@@ -3,160 +3,118 @@ var assert = require('assert');
 
 describe('rules/disallow-yoda-conditions', function() {
     var checker;
+    var operators = ['==', '===', '!=', '!==', '>', '<', '>=', '<='];
 
     beforeEach(function() {
         checker = new Checker();
         checker.registerDefaultRules();
-        checker.configure({ requireYodaConditions: true });
     });
 
-    it('should report yoda condition for strict equality', function() {
-        assert(
-            checker.checkString(
-                'if (x === 1) {\n' +
-                    'x++;\n' +
-                '}'
-            ).getErrorCount() === 1
-        );
+    operators.forEach(function(operator) {
+        var yodaCondition = 'var test; 1' + operator + 'test';
+        var notYodaCondition = 'var test; test' + operator + '1';
+
+        [[operator], true].forEach(function(value) {
+
+            describe(value + ' option', function() {
+                beforeEach(function() {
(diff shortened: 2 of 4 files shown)


lib/rules/disallow-yoda-conditions.js
```diff
@@ -1,14 +1,24 @@
...
+ * "disallowYodaConditions": [
+ *     "==",
+ *     "===",
+ *     "!=",
+ *     "!==",
+ *     ">",
+ *     "<",
+ *     ">=",
+ *     "<=",
+ *     "!="
```
hzoo: there's an extra `!=` here

hzoo:
looks good!

lib/rules/disallow-yoda-conditions.js
Slayer95: Removed it.

zxqfox:
I thought about it. Look like we lost a reason of denying yoda conditions.

Afaik it was a bad code style because there is a difference between `a = 5` and `a == 5`. And mistype is not a syntax error. But it's impossible to mistype `a < 5`. Do we really need this fix or we should just rethink this rule?

Slayer95:
I agree that yoda conditions are only useful to guard against that typo (i.e. <code>requireYodaConditions</code> can be limited to it. However, that's more easily dealt with other sort of rules, like JSHint's <code>"Expected a conditional expression and instead saw an assignment".</code>

Regarding <code>disallowYodaConditions</code>, an argument based on aesthetics and backwards compatibility can be made to keep supporting the current behavior, but I wouldn't mind its getting restricted to <code>==</code>.

Interestingly, ESLint provides the option to forbid Yoda Conditions [except for ranges](http://eslint.org/docs/rules/yoda.html), which is my use case.
<code>requireYodaConditions</code> only needs to distinguish between <code>=</code> and <code>==</code>

test/rules/disallow-yoda-conditions.js
```diff
@@ -3,84 +3,37 @@ var assert = require('assert');
...
+                function() {
+                    checker.configure({ disallowYodaConditions: value });
+                    assert(checker.checkString(notYodaCondition).isEmpty());
+                }
+            );
+        });
     });
 
     it('should report yoda condition for numeric', function() {
+        checker.configure({ disallowYodaConditions: true });
```
mrjoelkemp: Why was this removed from the `beforeEach`?

test/rules/require-yoda-conditions.js
```diff
@@ -3,84 +3,37 @@ var assert = require('assert');
...
+                function() {
+                    checker.configure({ requireYodaConditions: value });
+                    assert(checker.checkString(notYodaCondition).getErrorCount() === 1);
+                }
+            );
+        });
     });
 
     it('should report yoda condition for numeric', function() {
+        checker.configure({ requireYodaConditions: true });
```
mrjoelkemp: Should go back in the `beforeEach`.

mrjoelkemp:
Is this PR mergeable (once rebased)?

cc @zxqfox @hzoo

zxqfox:
@mrjoelkemp I think it's too complicated and can be simplified. I'm not sure that `5 <= value` should be interpreted as yoda condition.

markelog:
Is that still WIP?

Slayer95:
Waiting for feedback / replies to @zxqof

markelog:
ping @zxqfox

lib/rules/disallow-yoda-conditions.js
```diff
@@ -34,21 +43,22 @@ module.exports = function() {};
...
-            '!=': true,
-            '!==': true,
-            '>': true,
-            '<': true,
-            '>=': true,
-            '<=': true
-        };
+
+        if (isTrue) {
+            operators = ['==', '===', '!=', '!==', '>', '<', '>=', '<='];
```
zxqfox: What if make ['==', '===', '!='] by default (`true`)?

test/specs/rules/require-yoda-conditions.js
```diff
@@ -3,84 +3,37 @@ var assert = require('assert');
...
+                function() {
+                    checker.configure({ requireYodaConditions: value });
+                    assert(checker.checkString(notYodaCondition).getErrorCount() === 1);
+                }
+            );
+        });
     });
 
     it('should report yoda condition for numeric', function() {
+        checker.configure({ requireYodaConditions: true });
```
zxqfox: Usually, we making several `describe` blocks with different checker configurations. See https://github.com/jscs-dev/node-jscs/blob/master/test/specs/rules/disallow-space-after-object-keys.js

zxqfox:
Sorry for delay, somehow missed this PR. ;-(

lib/rules/disallow-yoda-conditions.js
mdevils: Why?

zxqfox: Even just '===' and '=='.
Historical reasons.

mdevils: I'm not sure about this. As to me they are all Yoda-style. And still I can't see enough reasons to break backwards compatibility.

Slayer95: It wouldn't actually break anything though. That is, no failing builds for no reason.

test/specs/rules/require-yoda-conditions.js
Slayer95: Will look at it.

lib/rules/disallow-yoda-conditions.js
mdevils: I'm not talking about the builds. I'm about the unexpected behaviour change. For instance, I know several projects which rely on the current behaviour.

zxqfox: Actually I'm fine if we left it as is.

Slayer95:
Updated tests as requested.

zxqfox:
@Slayer95 Thanks!

LGTM

mdevils:
LGTM! Good job!

markelog:
Just a remainder, if pr has two "LGTM" you can merge it, @mdevils, @zxqfox could one of you do that?

mdevils:
@markelog Yes, sure :)

mdevils:
Merged into `2.0`. Thank you!

PR #1036, opened 2015-02-09
Add esprimaOptions in BUILTIN_OPTIONS

yannickcr:
`esprimaOptions` was missing in the `BUILTIN_OPTIONS` object.

This could trigger a `Unsupported rule: esprimaOptions` error if we specify an `esprimaOptions` in the `.jscsrc` file.


mikesherov:
@yannickcr good catch! Can you please add a unit test for this behavior?

yannickcr:
@mikesherov I added a test to check that the rule is not added to the `unsupportedRuleNames` array

Issue #994, opened 2015-02-04
Add jscs-jsdoc and cover the core with comments


zxqfox:
Ref https://github.com/jscs-dev/node-jscs/pull/908#discussion_r24061948

zxqfox:
Houston, `jscs-jsdoc` has a `peerDependency` of `jscs` itself. Should I make a symlink in `node_modules`? Is there a more elegant solution?

mrjoelkemp:
Not sure, honestly. @sindresorhus might know.

sindresorhus:
Don't use `peerDependencies`. They're to be deprecated.

zxqfox:
@sindresorhus Yes sir! ;-( But what to use for plugins? (Please don't tell me that "plugins are deprecated" ;-)

zxqfox:
https://docs.npmjs.com/cli/dedupe ? Not sure that it's a right way to do it.

sindresorhus:
I don't know what the officially recommended solution is, just that `peerDependencies` is not the solution to anything. Dedupe is a good candidate. npm 3.0, which I think is shipping in the near time, is going to dedupe by default on install, which will make deduping a lot more useful.

zxqfox:
@sindresorhus Great news everyone! Thanks, it can really help with plugins. Looks like a real solution.

markelog:
So what's with this one?

zxqfox:
Looking for a time to get it done. ;-(

PR #1038, opened 2015-02-10
Extract JavaScript to validate from HTML files (fixes #764)

### LINKED ISSUE #764: Extract JavaScript to validate from HTML files
I saw an `--extract` option on JsHint (http://www.jshint.com/docs/cli/) that allow to extract JavaScript (what comes between `<script>` tags) from HTML pages. Would be great if JSCS would have similar option.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5953712-extract-javascript-to-validate-from-html-files?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lahmatiy:
Implementation of `extract` option, that behave similar to jshint `extract` option (see #764 for details).
Solution tries avoid undesirable errors by `disallowTrailingWhitespace`, `disallowMixedSpacesAndTabs ` and `validateIndentation` rules.

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -149,6 +149,19 @@ jscs path[ path[...]] --preset=jquery
 
 In order to add/remove preset rules you will need to create a `.jscsrc` config file.
 
+### `--extract`
+With this option you can set glob pattern for files which embeded JavaScript should be checked.
+```
+jscs path[ path[...]] --extract *.html 
+```
+
+You can set several patterns if necessary.
+```
+jscs path[ path[...]] --extract *.html --extract *.htm
+```
+
+Currently, only the `html` format is supported (JavaScript inside `<script>` will be checked) and extracted code cannot be auto fixed.
+
 ### `--reporter` (`-r`)
 `jscs` itself provides eight reporters: `checkstyle`, `console`, `inline`, `inlinesingle`, `junit`, `text`, `unix` and `json`.
 ```
@@ -281,6 +294,26 @@ Values: A single file extension or an Array of file extensions, beginning with a
 #### Default
 
 `.js` files are processed by default
+
+### extract
+
+Set list of glob patterns for files which embeded JavaScript should be checked.
+
+Type: `Array` or `false`
+
+Values: Array of file matching patterns
+
+JavaScript extracting from files that doesn't match to [fileExtensions](#fileExtensions) or [excludeFiles](#excludefiles), but match to patterns in this list. Currenly, only `html` format is supported.
+
+#### Example
+
+```js
+"extract": ["*.htm", "*.html"]
+```
+
+#### Default
+
+Default value is `["**/*.+(htm|html|xhtml)"]`. So JavaScript is extracted from files with `.htm`, `.html` or `.xhtml` extension.
 
 ### maxErrors
 Set the maximum number of errors to report (pass -1 or null to report all errors).
--- a/bin/jscs
+++ b/bin/jscs
@@ -17,6 +17,9 @@ program
     .option('-c, --config [path]', 'configuration file path')
     .option('--auto-configure [path]', 'auto-generate a JSCS configuration file')
     .option('-x, --fix', 'fix code style violations (applies to fixable violations)')
+    .option('--extract <mask>', 'set file masks from which to extract JavaScript', function(value, memo) {
+        return memo ? memo.concat(value) : [value];
+    })
     .option('-e, --esnext', 'attempts to parse esnext code (currently es6)')
     .option('--es3', 'validates code as es3')
     .option('-s, --esprima <path>', 'attempts to use a custom version of Esprima')
--- a/lib/checker.js
+++ b/lib/checker.js
@@ -1,6 +1,7 @@
 var vowFs = require('vow-fs');
 var Vow = require('vow');
 var StringChecker = require('./string-checker');
+var extractJs = require('./extract-js');
 var utils = require('util');
 var path = require('path');
 
@@ -65,6 +66,36 @@ Checker.prototype.fixFile = function(path) {
 };
 
 /**
+ * Extract JavaScript from file.
+ *
+ * @param {String} path
+ * @returns {Promise.<Errors>}
+ */
+Checker.prototype.extractFile = function(path) {
+    if (this._configuration.isFileExcluded(path)) {
+        return Vow.resolve(null);
+    }
+
+    if (!this._configuration.shouldExtractFile(path)) {
+        return Vow.resolve(null);
+    }
+
+    return vowFs.read(path, 'utf8').then(function(data) {
+        var result = extractJs(path, data);
+
+        result.sources.forEach(function(script) {
+            this.checkString(script.source, path).getErrorList().forEach(function(error) {
+                error.line += script.line;
+                error.column += script.offset;
+                result.addError(error);
+            });
+        }, this);
+
+        return result.errors;
+    }, this);
+};
+
+/**
  * Checks directory recursively.
  *
  * @param {String} path
@@ -116,12 +147,14 @@ Checker.prototype._processDirectory = function(path, fileHandler) {
                 }
 
                 if (!this._hasCorrectExtension(fullname)) {
-                    return [];
+                    if (!this._configuration.shouldExtractFile(fullname)) {
+                        return [];
+                    }
+
+                    return this.extractFile(fullname);
                 }
 
-                return Vow.when(fileHandler(fullname)).then(function(errors) {
-                    return errors;
-                });
+                return fileHandler(fullname);
             }, this);
         }, this)).then(function(results) {
             return [].concat.apply([], results);
@@ -138,7 +171,6 @@ Checker.prototype._processDirectory = function(path, fileHandler) {
  */
 Checker.prototype._processPath = function(path, fileHandler) {
     path = path.replace(/\/$/, '');
-    var _this = this;
 
     return vowFs.exists(path).then(function(exists) {
         if (!exists) {
@@ -147,14 +179,14 @@ Checker.prototype._processPath = function(path, fileHandler) {
 
         return vowFs.stat(path).then(function(stat) {
             if (stat.isDirectory()) {
-                return _this._processDirectory(path, fileHandler);
+                return this._processDirectory(path, fileHandler);
             }
 
             return fileHandler(path).then(function(errors) {
                 return [errors];
             });
-        });
-    });
+        }, this);
+    }, this);
 };
 
 /**
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -10,6 +10,7 @@ var BUILTIN_OPTIONS = {
     excludeFiles: true,
     additionalRules: true,
     fileExtensions: true,
+    extract: true,
     maxErrors: true,
     configPath: true,
     esnext: true,
@@ -124,6 +125,30 @@ function Configuration() {
     this._excludedFileMatchers = [];
 
     /**
+     * Extraction masks.
+     *
+     * @protected
+     * @type {Array}
+     */
+    this._extractFileMasks = [];
+
+    /**
+     * Default extractions masks.
+     *
+     * @protected
+     * @type {Array}
+     */
+    this._defaultExtractFileMasks = ['**/*.+(htm|html|xhtml)'];
+
+    /**
+     * List of file matchers from which to extract JavaScript.
+     *
+     * @protected
+     * @type {Array}
+     */
+    this._extractFileMatchers = [];
+
+    /**
      * Maxixum amount of error that would be reportered.
      *
      * @protected
@@ -222,6 +247,7 @@ Configuration.prototype.getProcessedConfig = function() {
     }, this);
     result.excludeFiles = this._excludedFileMasks;
     result.fileExtensions = this._fileExtensions;
+    result.extract = this._extractFileMasks;
     result.maxErrors = this._maxErrors;
     result.preset = this._presetName;
     result.esnext = this._esnextEnabled;
@@ -290,6 +316,27 @@ Configuration.prototype.isFileExcluded = function(filePath) {
  */
 Configuration.prototype.getFileExtensions = function() {
     return this._fileExtensions;
+};
+
+/**
+ * Returns extract file masks.
+ *
+ * @returns {String[]}
+ */
+Configuration.prototype.getExtractFileMasks = function() {
+    return this._extractFileMasks;
+};
+
+/**
+ * Should filePath to be extracted?
+ *
+ * @returns {Boolean}
+ */
+Configuration.prototype.shouldExtractFile = function(filePath) {
+    filePath = path.resolve(filePath);
+    return this._extractFileMatchers.some(function(matcher) {
+        return matcher.match(filePath);
+    });
 };
 
 /**
@@ -439,6 +486,14 @@ Configuration.prototype._processConfig = function(config) {
         options.additionalRules.forEach(function(rule) {
             this._loadAdditionalRule(rule, options.configPath);
         }, this);
+    }
+
+    if (options.hasOwnProperty('extract')) {
+        this._loadExtract(options.extract);
+
+    // Set default masks if there is no presets that could define their own
+    } else if (!options.hasOwnProperty('preset')) {
+        this._loadExtract(this._defaultExtractFileMasks);
     }
 
     if (options.hasOwnProperty('fileExtensions')) {
@@ -677,6 +732,27 @@ Configuration.prototype._loadExcludedFiles = function(masks) {
 
     this._excludedFileMasks = this._excludedFileMasks.concat(masks);
     this._excludedFileMatchers = this._excludedFileMasks.map(function(fileMask) {
+        return new minimatch.Minimatch(path.resolve(this._basePath, fileMask), {
+            dot: true
+        });
+    }, this);
+};
+
+/**
+ * Load paths for extract.
+ *
+ * @param {Array} masks
+ * @protected
+ */
+Configuration.prototype._loadExtract = function(masks) {
+    if (masks === false) {
+        masks = [];
+    }
+
+    assert(Array.isArray(masks), '`extract` option should be array of strings or false');
+
+    this._extractFileMasks = masks.slice();
+    this._extractFileMatchers = this._extractFileMasks.map(function(fileMask) {
         return new minimatch.Minimatch(path.resolve(this._basePath, fileMask), {
             dot: true
         });
--- a/lib/extract-js.js
+++ b/lib/extract-js.js
@@ -1,0 +1,230 @@
+var htmlparser = require('htmlparser2');
+var Errors = require('./errors');
+var rLineSplit = /\r\n|\r|\n/;
+var rHasNonWhitespace = /\S/;
+
+/**
+ * Html file representation (needed for errors output).
+ *
+ * @name HtmlFile
+ * @param {Object} params
+ * @param {String} params.filename
+ * @param {String} params.source
+ */
+var HtmlFile = function(params) {
+    this._filename = params.filename;
+    this._lines = params.source.split(rLineSplit);
+};
+
+HtmlFile.prototype = {
+    /**
+     * Returns source filename for this object representation.
+     *
+     * @returns {String}
+     */
+    getFilename: function() {
+        return this._filename;
+    },
+
+    /**
+     * Returns array of source lines for the file.
+     *
+     * @returns {String[]}
+     */
+    getLines: function() {
+        return this._lines;
+    }
+};
+
+/**
+ * Parse html and retrieve script sources.
+ *
+ * @param {String} html
+ * @returns {Object[]}
+ */
+function getScripts(html) {
+    function onopen(name, attrs) {
+        // tag should be a <script>
+        if (name !== 'script' ||
+            // ignore scripts with src attribute
+            attrs.src ||
+            // script tag should has no type attribute or attribute should be equal to text/javascript
+            (attrs.type && attrs.type.toLowerCase() !== 'text/javascript')) {
+            return;
+        }
+
+        // store script content start pos
+        scriptStartPos = parser.endIndex + 1;
+    }
+
+    function onclose() {
+        if (!scriptStartPos) {
+            return;
+        }
+
+        // get script content
+        var scriptEndPos = parser.startIndex;
+        var source = html.substring(scriptStartPos, scriptEndPos);
+
+        // store script content only if it contains non-whitespace characters
+        if (rHasNonWhitespace.test(source)) {
+            scripts.push({
+                source: source,
+                start: scriptStartPos,
+                end: scriptEndPos
+            });
+        }
+
+        // reset script start position
+        scriptStartPos = 0;
+    }
+
+    var scriptStartPos = 0;
+    var scripts = [];
+    var parser = new htmlparser.Parser({
+        onopentag: onopen,
+        onclosetag: onclose
+    });
+
+    parser.parseComplete(html);
+
+    return scripts;
+}
+
+/**
+ * JavaScript in HTML usually shifted based on first JS line. For example
+ * if first line of fragment is offset by 4 spaces, each line in this
+ * fragment will have offset 4 to restore the original column.
+ * This function trim script source and normalize lines offset.
+ *
+ * @param {String} source
+ * @returns {Object[]}
+ */
+function normalizeSource(source) {
+    var lines = source.split(rLineSplit);
+    var lineCount = lines.length;
+    var tabOnlyOffset = false;
+    var spaceOnlyOffset = false;
+    var offset;
+
+    // remove first list if it's an empty string
+    // usually <script> starts with new line
+    if (!rHasNonWhitespace.test(lines[0])) {
+        lines.shift();
+    }
+
+    // replace last line by empty string if it contains only whitespaces
+    // it helps avoid disallowTrailingWhitespace errors on last line
+    if (!rHasNonWhitespace.test(lines[lines.length - 1])) {
+        lines[lines.length - 1] = '';
+    }
+
+    // calculate min line offset
+    offset = Math.min.apply(null, lines.map(function(line) {
+        // skip empty lines
+        if (!line) {
+            return Infinity;
+        }
+
+        // fetch whitespaces at the line beginning
+        var offsetStr = line.match(/^\s*/)[0];
+        var tabCount = offsetStr.match(/\t*/)[0].length;
+
+        if (offsetStr.length === line.length) {
+            return 0;
+        }
+
+        // mixed spaces and tabs in one offset -> don't remove offsets
+        if (tabCount && tabCount !== offsetStr.length) {
+            return 0;
+        }
+
+        if (tabCount) {
+            if (spaceOnlyOffset) {
+                // no spaces, but previous offset has ony spaces -> mixed spaces and tabs
+                return 0;
+            } else {
+                // remember offset contains only tabs
+                tabOnlyOffset = true;
+            }
+        } else {
+            if (tabOnlyOffset) {
+                // no tabs, but previous offset has only tabs -> mixed spaces and tabs
+                return 0;
+            } else {
+                // remember offset contains only spaces
+                spaceOnlyOffset = true;
+            }
+        }
+
+        return offsetStr.length;
+    }));
+
+    // remove common offsets if possible
+    if (offset) {
+        lines = lines.map(function(line) {
+            return line.substr(offset);
+        });
+    }
+
+    return {
+        source: lines.join('\n'),
+        offset: offset,
+        lineCount: lineCount
+    };
+}
+
+/**
+ * Parse HTML and search for <script> sources. Each script source also normalize
+ * by line offset. Result contains script sources with information about line
+ * offset (that was removed for each line) and lines count before script source.
+ * This information helps restore absolute positions in html file for errors.
+ *
+ * @param {String} filename
+ * @param {String} data
+ * @returns {Object[]}
+ */
+function extractJs(filename, data) {
+    var errors = new Errors(new HtmlFile({
+        filename: filename,
+        source: data
+    }));
+    var scripts = getScripts(data);
+    var sources = [];
+    var line = 1;
+    var lastHtmlPos = 0;
+
+    scripts.forEach(function(scriptInfo) {
+        // fetch script source and normalize it
+        var normalized = normalizeSource(scriptInfo.source);
+
+        // add line offset before script
+        line += data.substring(lastHtmlPos, scriptInfo.start).split(rLineSplit).length - 1;
+
+        sources.push({
+            source: normalized.source,
+            offset: normalized.offset,
+            line: line
+        });
+
+        // save offsets for next fragment
+        line += normalized.lineCount - 1;
+        lastHtmlPos = scriptInfo.end;
+    });
+
+    return {
+        sources: sources,
+        errors: errors,
+        addError: function(error) {
+            errors._errorList.push({
+                filename: filename,
+                rule: error.rule,
+                message: error.message,
+                line: error.line,
+                column: error.column
+            });
+        }
+    };
+}
+
+module.exports = extractJs;
--- a/test/specs/checker.js
+++ b/test/specs/checker.js
@@ -64,6 +64,14 @@ describe('checker', function() {
                 assert.equal(errors[0].getErrorList().length, 0);
             });
         });
+
+        it('should throw an exception when path doesn\'t exists', function() {
+            return checker.checkPath('./test/non-exists').then(function() {
+                assert(false);
+            }, function() {
+                assert(true);
+            });
+        });
     });
 
     describe('checkStdin', function() {
@@ -93,6 +101,92 @@ describe('checker', function() {
 
             process.stdin.emit('data', 'foo');
             process.stdin.emit('end');
+        });
+    });
+
+    describe('extract', function() {
+        it('should extract from *.htm, *.html and *.xhtml by default', function() {
+            return checker.checkPath('./test/data/extract').then(function(errors) {
+                assert.equal(errors.length, 3);
+                assert.equal(errors[0].getErrorList().length, 2);
+                assert.equal(errors[1].getErrorList().length, 2);
+                assert.equal(errors[2].getErrorList().length, 0);
+            });
+        });
+
+        it('should not extract js when set to false', function() {
+            checker.configure({
+                extract: false
+            });
+
+            return checker.checkPath('./test/data/extract').then(function(errors) {
+                assert.equal(errors.length, 0);
+            });
+        });
+
+        it('should try extract js from any file', function() {
+            checker.configure({
+                extract: ['**']
+            });
+
+            return checker.checkPath('./test/data/extract').then(function(errors) {
+                assert.equal(errors.length, 4);
+                assert.equal(errors[0].getErrorList().length, 2);
+                assert.equal(errors[1].getErrorList().length, 2);
+                assert.equal(errors[2].getErrorList().length, 1);
+                assert.equal(errors[3].getErrorList().length, 0);
+            });
+        });
+
+        it('extractFile should return empty error list if no errors', function() {
+            var checker = new Checker();
+
+            // no check rules no any errors
+            checker.configure({
+                extract: ['**']
+            });
+
+            return checker.extractFile('./test/data/extract/always.htm').then(function(errors) {
+                assert.equal(errors.getErrorList().length, 0);
+            });
+        });
+
+        it('extractFile should return null if file doesn\'t check', function() {
+            checker.configure({
+                extract: false
+            });
+
+            return checker.extractFile('./test/data/extract/always.htm').then(function(errors) {
+                assert.equal(errors, null);
+            });
+        });
+
+        describe('should take in account excludeFiles', function() {
+            it('on checkPath', function() {
+                checker.configure({
+                    extract: ['**'],
+                    excludeFiles: [
+                        './test/**'
+                    ]
+                });
+
+                return checker.checkPath('./test/data/extract').then(function(errors) {
+                    assert.equal(errors.length, 0);
+                });
+            });
+
+            it('on extractFile', function() {
+                checker.configure({
+                    extract: ['**'],
+                    excludeFiles: [
+                        './test/**'
+                    ]
+                });
+
+                return checker.extractFile('./test/data/extract/always.htm').then(function(errors) {
+                    assert.equal(errors, null);
+                });
+            });
         });
     });
 
--- a/lib/config/node-configuration.js
+++ b/lib/config/node-configuration.js
@@ -14,7 +14,8 @@ var OVERRIDE_OPTIONS = [
     'esprima',
     'es3',
     'verbose',
-    'esnext'
+    'esnext',
+    'extract'
 ];
 
 function req(entity, dir) {
--- a/package.json
+++ b/package.json
@@ -86,7 +86,8 @@
     "to-single-quotes": "^1.0.2",
     "vow": "~0.4.8",
     "vow-fs": "~0.3.4",
-    "xmlbuilder": "^2.6.1"
+    "xmlbuilder": "^2.6.1",
+    "htmlparser2": "3.8.3"
   },
   "devDependencies": {
     "browserify": "^11.0.0",
--- a/test/data/extract/always.htm
+++ b/test/data/extract/always.htm
@@ -1,0 +1,18 @@
+should be ignored when not always
+<!doctype html>
+
+<html>
+<head>
+  <title>Test file</title>
+  <script type="ignore/me"></script>
+  <script>
+    with (x) { y++; }
+  </script>
+</head>
+<body>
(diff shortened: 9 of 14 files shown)


aik099:
:+1: Please also do the squash.

zxqfox:
I'm not sure but it's very close to inversion of jsx checking. And since that check uses custom esprima should do the extraction right in checker?

zxqfox:
btw. One more similar and common case is checking js inside md-files.

lib/checker.js
```diff
@@ -45,6 +46,38 @@ Checker.prototype.checkFile = function(path) {
 };
 
 /**
+ * Extract JavaScript from file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
+ */
+Checker.prototype.extractFile = function(path) {
+    if (!this._configuration.isFileExcluded(path)) {
```
mrjoelkemp:
```
if (!this._configuration.isFileExcluded(path)) {
  return Vow.resolve(null);
}

return vowFs.read(path, 'utf8').then(function(data) {
...
```

lib/config/configuration.js
```diff
@@ -148,6 +151,15 @@ Configuration.prototype.getFileExtensions = function() {
 };
 
 /**
+ * Returns extract mode.
+ *
+ * @returns {String[]}
```
mrjoelkemp: Seems to only return a `String` not a `String[]`.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
+var htmlparser = require('htmlparser2');
+var Errors = require('./errors');
+
+/**
+ * Html file representation (needs for errors output).
```
mrjoelkemp: `needed`?

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * @name HtmlFile
+ */
+var HtmlFile = function(filename, source) {
```
mrjoelkemp: Document params please

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+    this._lines = source.split(/\r\n|\r|\n/);
+};
+
+HtmlFile.prototype = {
+    /**
+     * Returns source filename for this object representation.
+     *
+     * @returns {String}
+     */
+    getFilename: function() {
```
mrjoelkemp: Opt for `HtmlFile.prototype.getFilename`

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+};
+
+/**
+ * Parse html and retrieve script sources.
+ * @param {String} html
+ * @returns {Object[]}
+ */
+function getScripts(html) {
+    function onopen(name, attrs) {
```
mrjoelkemp: Consider extracting outside of `getScripts` with documentation of the function

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+        // tag should be a <script>
+        if (name !== 'script') {
```
mrjoelkemp: Consider:

```js
// tag should be a <script>
if (name !== 'script' || 
    // ignore scripts with src attribute  
    attrs.src ||
    // script tag should has no type attribute or attribute should be equal to text/javascript
    (attrs.type && attrs.type.toLowerCase() !== 'text/javascript')) {
    return;
}
```

lib/extract-js.js
mrjoelkemp: Ah, I see that this depends on closed over vars later on. Tiny touch confusing, but no harm done. A class might be clearer, but not a blocker.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+            return;
+        }
+
+        // store script content start pos
+        scriptStartPos = parser.endIndex + 1;
+    }
+
+    function onclose() {
+        // if there is a script start pos
+        if (scriptStartPos) {
```
mrjoelkemp:
```
if (!scriptStartPos) {
    return;
}
```

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+        if (attrs.type && attrs.type.toLowerCase() !== 'text/javascript') {
+            return;
+        }
+
+        // store script content start pos
+        scriptStartPos = parser.endIndex + 1;
+    }
+
+    function onclose() {
+        // if there is a script start pos
```
mrjoelkemp: Unhelpful comment. Please remove

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+        onclosetag: onclose
+    });
+
+    parser.parseComplete(html);
+
+    return scripts;
+}
+
+/**
+ * Trim script source and normalize lines offset.
```
mrjoelkemp: @zxqfox Is there a jscs-jsdoc rule to enforce a newline after function description? Might be useful to add to #994

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+ * @returns {Object[]}
+ */
+function extractJs(filename, data) {
+    var errors = new Errors(new HtmlFile(filename, data));
+    var scripts = getScripts(data);
+    var sources = [];
+    var line = 1;
+    var lastHtmlPos = 0;
+
+    for (var i = 0; i < scripts.length; i++) {
```
mrjoelkemp: Opt for `scripts.forEach(function(scriptInfo) {...`

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+        lineCount: lineCount
+    };
+}
+
+/**
+ * Parse HTML and search for <script> sources. Each script source also normalize
+ * by line offset. Result contains script sources with information about line
+ * offset (that was removed for each line) and lines count before script source.
+ * This information helps restore absolute positions in html file for errors.
+ * @param {String}
```
mrjoelkemp: Finish the params documentation please.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+ * @returns {Object[]}
+ */
+function extractJs(filename, data) {
```
mrjoelkemp: Consider renaming `data` to `source` as it's more descriptive and matches the constructor's signature.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+var Errors = require('./errors');
+
+/**
+ * Html file representation (needs for errors output).
+ *
+ * @name HtmlFile
+ */
+var HtmlFile = function(filename, source) {
+    this._filename = filename;
+    this._lines = source.split(/\r\n|\r|\n/);
```
mrjoelkemp: Can this regex be substituted with `/\r\n?|\n\r?/`? If so, extract this regex to a closed over var near the top of the file. It's used in 3 places.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+
+    for (var i = 0; i < scripts.length; i++) {
+        // fetch script source and normalize it
+        var scriptInfo = scripts[i];
+        var normalized = normalizeSource(scriptInfo.source);
+
+        // add line offset before script
+        line += data.substring(lastHtmlPos, scriptInfo.start).split(/\r\n?|\n\r?/).length - 1;
+
+        // store result
```
mrjoelkemp: Unnecessary comment. Please remove

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+        // save offsets for next fragment
+        line += normalized.lineCount - 1;
+        lastHtmlPos = scriptInfo.end;
+    }
+
+    return {
+        sources: sources,
+        errors: errors,
+        addError: function(error) {
+            errors._errorList.push({
```
mrjoelkemp: `errors.add`?

mrjoelkemp: Consider:

```js
addError: function(error) {
    error.filename = filename;
    errors.add(error);
}
```

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+function normalizeSource(source) {
+    var lines = source.split(/\r\n?|\n\r?/);
+    var lineCount = lines.length;
+    var lastLine = lines.pop();
+    var tabOnlyOffset = false;
+    var spaceOnlyOffset = false;
+    var offset;
+
+    // remove first list if it's an empty string
+    // usualy <script> starts with new line
```
mrjoelkemp: `usually`

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+
+    function onclose() {
+        // if there is a script start pos
+        if (scriptStartPos) {
+            // get script content
+            var scriptEndPos = parser.startIndex;
+            var source = html.substring(scriptStartPos, scriptEndPos);
+
+            // store script content only if it contains non-whitespace characters
+            if (/\S/.test(source)) {
```
mrjoelkemp: Consider extracting `/\S/` to a var for reuse. The var name should explain what the regex represents. The comment above could likely be removed in that case.

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+
+    // remove first list if it's an empty string
+    // usualy <script> starts with new line
+    if (!lines[0]) {
+        lines.shift();
+    }
+
+    // replace last line by empty string if it contains only whitespaces
+    // it helps avoid disallowTrailingWhitespace errors on last line
+    lines.push(/\S/.test(lastLine) ? lastLine : '');
```
mrjoelkemp: Why pop the last line only to push it back on? Why not just do something like:

```js
var lastLine = lines[lines.length - 1];
...
if (!/\S/.test(lastLine)) {
  lines[lines.length - 1] = '';
}
```

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+
+    // calculate min line offset
+    offset = Math.min.apply(null, lines.map(function(line) {
+        // skip empty lines
+        if (!line) {
+            return Infinity;
+        }
+
+        // fetch whitespaces at the line begining
```
mrjoelkemp: `beginning`

lib/extract-js.js
```diff
@@ -0,0 +1,216 @@
...
+
+    return scripts;
+}
+
+/**
+ * Trim script source and normalize lines offset.
+ * @param {String} source
+ * @returns {Object[]}
+ */
+function normalizeSource(source) {
```
mrjoelkemp: I'm not understanding why this function is necessary. Mind expanding on this function's docblock explaining why?

lib/checker.js
```diff
@@ -60,15 +93,22 @@ Checker.prototype.checkDirectory = function(path) {
...
                 if (stat.isDirectory()) {
                     return this.checkDirectory(fullname);
                 }
 
                 if (!this._hasCorrectExtension(fullname)) {
-                    return [];
+                    if (this._configuration.getExtract() === 'never') {
+                        return [];
+                    }
+                    promise = Vow.when(this.extractFile(fullname));
```
mrjoelkemp: Wouldn't you need to check that `fullname` contains the `.html` extension? Is that check done somewhere else? Seems like this would attempt to parse *everything* but `.js` files.

lib/checker.js
```diff
@@ -45,6 +46,38 @@ Checker.prototype.checkFile = function(path) {
...
 /**
+ * Extract JavaScript from file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
+ */
+Checker.prototype.extractFile = function(path) {
+    if (!this._configuration.isFileExcluded(path)) {
+        return vowFs.read(path, 'utf8').then(function(data) {
+            if (this._configuration.getExtract() === 'always' || /^\s*</.test(data)) {
```
mrjoelkemp: Comment on what the regex is or extract to a var with a name that explains it.

lib/checker.js
```diff
@@ -45,6 +46,38 @@ Checker.prototype.checkFile = function(path) {
...
 
 /**
+ * Extract JavaScript from file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
+ */
+Checker.prototype.extractFile = function(path) {
+    if (!this._configuration.isFileExcluded(path)) {
+        return vowFs.read(path, 'utf8').then(function(data) {
```
mrjoelkemp: Consider renaming `data` to `source` for clarity

lib/checker.js
```diff
@@ -45,6 +46,38 @@ Checker.prototype.checkFile = function(path) {
...
+
+                content.sources.forEach(function(script) {
+                    this.checkString(script.source, path).getErrorList().forEach(function(error) {
+                        error.line += script.line;
+                        error.column += script.offset;
+                        content.addError(error);
+                    });
+                }, this);
+
+                if (!content.errors.isEmpty()) {
```
mrjoelkemp: Need to return something for the case when it's empty? Would `null` suffice?

lib/checker.js
```diff
@@ -45,6 +46,38 @@ Checker.prototype.checkFile = function(path) {
 };
 
 /**
+ * Extract JavaScript from file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
```
mrjoelkemp: I believe we typically do `@returns {Promise} descriptionHere` where the description can describe the resolved values. In this case it's an error list or null.

test/checker.js
```diff
@@ -47,6 +47,14 @@ describe('modules/checker', function() {
                 assert(errors.length === 1);
             });
         });
+
+        it('should throw an exception when path doesn\'t exists', function() {
+            return checker.checkPath('./test/non-exists').then(function() {
```
mrjoelkemp: An alternative might be:

```js
return checker.checkPath('./test/non-exists').then(null, done);
```

This requires `done` as a parameter to the `it`, of course

test/checker.js
```diff
@@ -78,4 +86,107 @@ describe('modules/checker', function() {
...
+
+        it('extractFile should return null if no errors', function() {
+            var checker = new Checker();
+            checker.registerDefaultRules();
+            checker.configure({
+                extract: 'always'
+            });
+
+            return checker.extractFile('./test/data/extract/always.html').then(function(errors) {
+                assert(errors === null);
```
mrjoelkemp: Why return `null` instead of an empty list?

test/extract-js.js
```diff
@@ -0,0 +1,71 @@
+var assert = require('assert');
+var Checker = require('../lib/checker');
+
+describe('modules/extract', function() {
```
mrjoelkemp: `modules/extract-js` although `lib/extract-js` would be more accurate (though, I'm aware we have the discrepancy for other test suites).

test/extract-js.js
```diff
@@ -0,0 +1,71 @@
...
+            });
+
+            return checker.checkPath('./test/data/extract').then(function(errorsCollection) {
+                var errorCount = 0;
+
+                assert.doesNotThrow(function() {
+                    errorsCollection.forEach(function(errors) {
+                        if (!errors.isEmpty()) {
+                            errors.getErrorList().forEach(function(error) {
+                                errors.explainError(error, true);
```
mrjoelkemp: Does this simply test that the error format added within extract-js is correct?

test/extract-js.js
```diff
@@ -0,0 +1,71 @@
...
+
+        it('should be errors on disallowMixedSpacesAndTabs', function() {
+            var checker = new Checker();
+            checker.registerDefaultRules();
+            checker.configure({
+                disallowMixedSpacesAndTabs: true
+            });
+            return checker.extractFile('./test/data/extract/index.html').then(function(errors) {
+                // disallowMixedSpacesAndTabs doesn't warn when some line contains only and some
+                // only spaces (i.e. '\s\svar foo;\t\tvar bar;'). If this behavior will change
```
mrjoelkemp: Sounds like an issue should be filed.

mrjoelkemp: Would you mind adding an issue for this with what you expect to see from the rule?

mrjoelkemp:
@lahmatiy This is really impressive work! Great job and thanks for contributing!

Please rebase off of master. 

I added a bunch of comments, most of which are suggestions to increase the clarity of the code. Feel free to ask questions about any of the comments.

After the comments are addressed, we'll need another maintainer to review the code before merging. It's a big PR, so we need more than one review on this one.

You're also missing changes to Overview.md: https://github.com/jscs-dev/node-jscs/blob/master/OVERVIEW.md#options to document the `extract` option and its possible values.

I noticed (and you stated) that the algorithm is similar to jshint's approach but the code is original, so I think we're okay with licensing.

Awesome work!

lib/extract-js.js
aik099: I'm not sure that we should even compare `type` attribute of `script` tag, because in HTML5 it defaults to `text/javascript` when missing.

lib/extract-js.js
zxqfox: Nope, but it's easy to make ;-) https://github.com/jscs-dev/jscs-jsdoc/issues/85

test/checker.js
zxqfox: I'm not sure but looks like mocha has built-in promise support.

test/extract-js.js
zxqfox: :+1:

lahmatiy:
@mrjoelkemp Thank you. I'll fix PR asap.

I read your comments. And I would like to notice that I tried write the code as close as possible to original codebase. I prefer code consistency even when I see that some code could be written better. Otherwise other similar code fragments should be changed too, but it's out of bounds of this PR.

I also worrying about default value for `extract`. JSHint has `auto` by default. But the same setting for `jscs` might break build of projects that using `jscs`, as `jscs` will start checking something that didn't check before. Probably `never` should be better for backward capability. In this case `extract` will be opt-in.

aik099:
Maybe you should also rebase, because GitHub showing that PR can't be merged due conflicts.

lahmatiy:
@aik099 jscs codebase changes too fast, that PRs becomes broken all the time :( Sure I'll rebase, when return to work on feature.

@mdevils I believe your should take a look on this PR. As I know you're refactoring `JsFile` class. For this feature we need some common `File` abstraction. Because we need to parse fragments of some file and it isn't a JavaScript file. I've made a simple mock class `HtmlFile` for this purposes (temporary solution). But I'm sure there should be better solution.

zxqfox:
@lahmatiy Better solution is to use custom esprima imo.

aik099:
> jscs codebase changes too fast

That is something each open source project can wish for. As long as not BC breaks are introduced.

lahmatiy:
@zxqfox How it could help to parse `html`, find out `js` fragments and aggregate fragment errors in one error list mapped on original file?

zxqfox:
@lahmatiy It can skip all html nodes as comments (or anything concrete) and returns just Parser API AST nodes to for checking. But I agree that it will be a problem with indentation probably.

lahmatiy:
@zxqfox It's not a working solution. We need text of correct `<script>` tags only. And we should know where fragment starts and ends. It requires for normalization, as rules check indentation, empty lines, trailing whitespaces, trailing new line, `var` position and count etc.

mikesherov:
> @zxqfox It's not a working solution. We need text of correct <script> tags only. And we should know where fragment starts and ends. It requires for normalization, as rules check indentation, empty lines, trailing whitespaces, trailing new line, var position and count etc.

@lahmatiy the correct way to handle this then seems to make the rules that check this stuff aware of that fact. For example, the indentation problem is easily fixed by making the rule not assume that 0 indents mean column 1. It can easily check the indentation of the first token it encounters and set that as the offset of the code. 

For the rest of the issues, it seems like a possible solution is remove all non script tags, replace it all with whitespace, and add `/*jscs:disable*/` to the top of the document, replace `<script>` with `/*jscs:enable*/`, and `</script>` with `/*jscs:disable*/`.
 
For example:

```
<html>
....
<body>
yada
  <script>
  var a=1;
  </script>
  dada
  <script>
  var b=2;
  </script>
```

becomes:
```
/*jscs:disable*/



  /*jscs:enable*/
  var a=1;
  /*jscs:disable*/

  /*jscs:enable*/
  var b=2;
  /*jscs:disable*/
```

this will automatically ignore errors anywhere that isn't actual scripts tags, and automatically retain position information. All that needs to be adjusted is again, the offset for indentation.

zxqfox:
@lahmatiy Em. Why not?
For simple html:
```

for 
you'll have AST like that:
```js
comments (or something):
[
  { type: 'HTMLPart', value: '<html>...<script></script><script>\n</script>...<script>', range: ... },
  { type: 'HTMLPart', value: '</script>etc...</html>' }
]
nodes:
[
  { type: 'SomeStandardNode', value: 'var', etc... }, ...
]
```

The only thing to fix here is lines/columns normalization for indentation and reporters. For errors

lib/extract-js.js
zxqfox: I think `/\r\n?|\n/` is enough. What is `\n\r`?

aik099: See there is `|` in there, so we're catching any wild `\n` or `\r`.

zxqfox: @aik099 Line end is a `\r` (mac) or `\n` (unix) or `\r\n` (win). We can tune for gracefully parse mixes of them, but it really can be substituted with `/\r\n?|\n/`. Right?

aik099: `\r` isn't used on Mac since ages :wink: To deal with files, where mixed line endings exists I usually do this:

1. replace `\r\n` with `\n`
2. replace `\r` with `\n`
3. split by `\n`

lahmatiy:
@mikesherov @zxqfox
I'm not sure that custom parser could strip markup correctly. There are many cases for wrong matches if you don't process source as html, but js with possible markup inside.
We can't merge all scripts in single one. We should process scripts independantly (as separate files, as browsers do). Otherwise one broken script may broke others, or fix it, or change sematic of next script, for example:
```html
<script>
  (function(){ /* I'm IIFE */ })()
</script>
<script>
  (function(){ /* I'm IIFE */ })()
</script>
```

Indentations may vary from `<script>` to `<script>`, so fixing `validateIndentation` rule couldn't help here.
And there are some rules that could find an errors, because of script merge and no normalization:
- disallowPaddingNewlinesBeforeKeywords
- disallowSemicolons
- disallowTrailingWhitespace
- requireMultipleVarDecl
- requireParenthesesAroundIIFE
- requireSemicolons (PR)
- validateIndentation
- validateQuoteMarks: true

etc.

mikesherov:
@lahmatiy those are good points. Still would like to see this code somehow simplified. Adds a lot of complexity at the moment.

mdevils:
@lahmatiy I will take a look soon.

mdevils:
We discussed this with @lahmatiy, this implementation looks like a hack right now. We need better abstraction for this case (which will also support autofixing).

I'm going to work on this later, but it might require to make a `2.0` as API rebuild might be necessary.

What do you think?

zxqfox:
:+1: Go ahead!

lahmatiy:
@zxqfox @mdevils Checking code inside jsdocs in @example section would be useful too ;)

markelog:
How about we settle on this version right now and rewrite it if needed?

zxqfox:
@markelog It would be great to have an abstraction for these purposes...

markelog:
There is stuff to improve, but i would rather do it later, get response from the community and wait three months

nicolasgarnier:
Any news on this feature? We want to add JSCS checks as part of the Polymer Starter Kit's build pipeline so we're trying to assess if we should go for a workaround or wait for this.

Cheers!

lahmatiy:
@mdevils any news about new abstractions for code fragments? Maybe we could merge this feature for html only with future improvements (js fragments in any content type) when ready?
/cc @mikesherov @markelog

markelog:
We should prioritize this and schedule for the new release

lahmatiy:
Ok, I will rebase pr and fix review notes then.
When do you plan next release?

markelog:
Whenever it's ready :-), it seems it could be soon though, there is couple prs we could land

nicolasgarnier:
Awesome thanks guys. We'll keep an eye on this PR and hope for the best :)

lib/checker.js
lahmatiy: I'm not sure about this. jshint does the same as I remember

lib/extract-js.js
lahmatiy: You mean `HtmlFile.prototype.getFilename = function() { .. }` should be here?
I wrote this code as close as possible to [JsFile style](https://github.com/jscs-dev/node-jscs/blob/master/lib/js-file.js#L60).

lib/extract-js.js
lahmatiy: It doesn't work as you expected https://github.com/jscs-dev/node-jscs/blob/master/lib/errors.js#L34
I'll be happy if it does

test/extract-js.js
lahmatiy: I had an issue when reporter crashes for somehow. So I'd add this test.
But I don't remember details. Probably this could be removed...

lahmatiy: Ah... I just remembered! Reporter crashes as we use custom file class (`HtmlFile`) in errors. So error details should be compatible with console reporter (other reporters ok, but console reporter not, because do additional formatting). Something like that.

test/extract-js.js
lahmatiy: Well, I expect that if some lines has tabs offset but others spaces offset – warning should be shown. For now nothing happen. So, here is we will see 2 warnings:

```
\s\t
\t\s
```

But in this case no warnings:

```
\s\s
\t\t
```

But probably it's a feature, not a bug ;)

lahmatiy:
Mostly done.
If everything is ok I'm squash commits then.

lib/extract-js.js
```diff
@@ -0,0 +1,231 @@
...
+ * @returns {Object[]}
+ */
+function getScripts(html) {
+    function onopen(name, attrs) {
+        // tag should be a <script>
+        if (name !== 'script' ||
+            // ignore scripts with src attribute
+            attrs.src ||
+            // script tag should has no type attribute or attribute should be equal to text/javascript
+            (attrs.type && attrs.type.toLowerCase() !== 'text/javascript')) {
```
markelog: `text/javascript`  is obsolete, check this out - http://tools.ietf.org/html/rfc4329

lib/extract-js.js
markelog: Hm, it seems we need to create a public push method

test/specs/checker.js
```diff
@@ -80,6 +88,109 @@ describe('checker', function() {
...
+
+        it('should not extract js when set `never`', function() {
+            var checker = new Checker();
+            checker.registerDefaultRules();
+            checker.configure({
+                disallowKeywords: ['with'],
+                extract: 'never'
+            });
+
+            return checker.checkPath('./test/data/extract').then(function(errors) {
```
markelog: What if pass a full path?

markelog:
`extract`
So this is not a cli flag? What about documentation? Do we really needed? JSHint has it, but i wonder if it is redundant?

lib/checker.js
markelog: Wouldn't that slows us down? If user does something like `jscs .`? Then read we would read every perky out there.

lib/extract-js.js
lahmatiy: I know, but still valid and may be used in sources.
This condition checks that `<script>` has no `type` other than `text/javascript`.

```html
<script></script>  <!-- process -->
<script type="text/javascript"></script>  <!-- process -->
<script type="some/template"></script>    <!-- ignore -->
```

markelog: Yes, sure, but i provided the link not for nothing, what about - 

```
application/ecmascript
application/javascript
application/x-ecmascript
application/x-javascript
text/ecmascript
text/javascript
text/javascript1.0
text/javascript1.1
text/javascript1.2
text/javascript1.3
text/javascript1.4
text/javascript1.5
```

lib/checker.js
markelog: How about extending defaults of `fileExtensions` option instead of introducing the `extract` and putting every possible html extension?

It has obvious downside, like if user redefine it, their would need to add those extension again,

lib/checker.js
```diff
@@ -1,6 +1,7 @@
 var vowFs = require('vow-fs');
 var Vow = require('vow');
 var StringChecker = require('./string-checker');
+var extractJs = require('./extract-js');
```
markelog: So this class called as `HtmlFile` analogues to `JsFile`, but file named as `extract-js`?

lahmatiy: `HtmlFile` is just a helper for `extract` option, because we can't use `JsFile`, but errors must has a reference to some `file`.

lib/checker.js
lahmatiy: We could use `never` as default value for `extract` option as `JSHint` has. I believe we should do initial solution as simple as possible. Additional options (and other improvement) could be added later.

If somebody runs `jscs .` it's not a problem I guess. `never` for `extract` by default should does no any performance impact.

lib/extract-js.js
lahmatiy: I suppose values other than `text/javascript` is really rare used. If you insist, I can add check for all of those values. But I thought it's good idea to start with `text/javascript` and add new values if needed depending on users feedback.

test/specs/checker.js
lahmatiy: Should I add `checkPath` and `checkFile` versions for every single test?

lahmatiy:
@markelog 

> So this is not a cli flag? What about documentation?

No requests before, but I could add both if needed. But we need to determine your interest first. PR is 7 months old already, but opinion about feature is still not clear to me.

> Do we really needed? JSHint has it, but i wonder if it is redundant?

Yes, it's needed. I saw feature requests before. And one day I needed this feature too. That's why I implemented it and thought it might be helpful for others too. You can see some attached issues here. I don't understand why you think it might be redundant?

Most strange part to me, that you ask those questions after your own proposition recently to continue work on it (and proposition about merge in june) and when I fixed known problems once again :-/

lib/checker.js
markelog: If `JSHint` did it, it doesn't mean it is a good solution, with `extract = never` by default, user would need to know about this option, with `extract = auto` by default, it would hurt performance.

Whereas if we would extend `fileExtensions` we would support most use-cases performance downside.

> I believe we should do initial solution as simple as possible.

If we would add `extract` option we need to support it for long time.

lib/checker.js
markelog: Fair enough

test/specs/checker.js
markelog: I was asking would happen if we would pass a full path with extension

lib/extract-js.js
markelog: Would be good, but yeah, we could do it later

markelog:
> No requests before, but I could add both if needed. But we need to determine your interest first. PR is 7 months old already, but owners opinion about feature is still not clear to me.

No-no, that was about replacing it with `fileExtensions` property.

> PR is 7 months old already, but owners opinion about feature is still not clear to me.

I would landed it it seven months ago, but couldn't, since some of us thought we should wait, in other words, sometimes, perfectionism wins when it shouldn't :-(.

> Most strange part to me, that you ask those questions after your own proposition recently to continue work on it (and proposition about merge in june) and when I fixed known problems once again :-/

Sorry for misunderstanding, i should have better express my thoughts, i'm very excited about this feature,  one way or another, it will be in the next release.

lib/checker.js
lahmatiy: Maybe add support for array of minimatch masks in additional of `never`, `auto` and `always`? And set `['*.html', '*.htm', '*.xhtml']` or `['*.(html?|xhtml)']` by default?

```
{
  "extract": ['*.html', '*.htm', '*.xhtml']
}
```

```
jscs . --extract *.html *.xhtml
```

markelog: Sounds good!

lahmatiy:
Oh, I see. Thank you for your clarifications.

I implementing cli option and array of masks support then. But I'm not sure about documentation, should I add it here https://github.com/jscs-dev/jscs-dev.github.io/blob/master/overview.html?

markelog:
[Here](https://github.com/jscs-dev/node-jscs/blob/master/OVERVIEW.md) would be good

lahmatiy:
OK. Will be done.

markelog:
Would like to release new version this week, will you have time to finish this up?

lahmatiy:
I have some problems with merge now. Even after merge conflict resolving, some tests are failed because of changes in master :( I tried to fix it today but no success yet.
I afraid I have no time till Sunday. But on Sunday I would continue to work on it (maybe earlier but I'm not sure).

markelog:
> I afraid I have no time till Sunday. But on Sunday I would continue to work on it (maybe earlier but I'm not sure).

Friendly ping

lahmatiy:
Sorry for delay.
Planning to continue work later today.

lahmatiy:
Looks done.
Awaiting for review.

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
...
+```
+jscs path[ path[...]] --extract *.html 
+```
+
+You could set several masks if needed.
+```
+jscs path[ path[...]] --extract *.html --extract *.htm
+```
+
+Or you could you one of shortcut value: `always`, `auto` or `never`. Note that this values couldn't be comined with file masks.
```
markelog: `comined`

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
...
+
+```
+jscs path[ path[...]] --extract always
+```
+
+Currently only `html` format is supported (JavaScript inside `<script>` will be checked). Auto fixing is not apply.
```
markelog: ` Auto fixing is not apply` - > `Auto fixing does not apply`?

OVERVIEW.md
```diff
@@ -281,6 +300,30 @@ Values: A single file extension or an Array of file extensions, beginning with a
 
 `.js` files are processed by default
 
+### extract
+
+Set list of file masks from which JavaScript should be extracted.
+
+Type: `Array` or `"always"` or `"auto"` or `"never"`
+
+JavaScript extracting from files that doesn't match to `fileExtensions` or `excludeFiles`. Only `html` format supported.
```
markelog: Maybe other options could be links?

OVERVIEW.md
```diff
@@ -281,6 +300,30 @@ Values: A single file extension or an Array of file extensions, beginning with a
...
+
+Special values meaning:
```
markelog: Would remove `meaning`

lahmatiy:
For now I'm not sure we need for special values, i.e. `always`, `auto` and `never`. It looks odd. Maybe I should remove it?

OVERVIEW.md
```diff
@@ -281,6 +300,30 @@ Values: A single file extension or an Array of file extensions, beginning with a
...
+- `always` - JavaScript extracts from any file
```
markelog: `JavaScript extracts from any file` -> `extract from any file`, since you use imperative mood down the road

OVERVIEW.md
```diff
@@ -281,6 +300,30 @@ Values: A single file extension or an Array of file extensions, beginning with a
...
+#### Example
+
+```js
+"extract": ["*.htm", "*.html"]
+"extract": "auto"
+```
+
+#### Default
+
+Default value is `["**/*.+(htm|html|xhtml)"]`. So JavaScript extracts from files with `.htm`, `.html` or `.xhtml` extension.
```
markelog: `So JavaScript extracts` -> `So JavaScript is extracted`?

bin/jscs
```diff
@@ -36,3 +38,25 @@ program
     .parse(process.argv);
 
 cli(program);
+
+/**
+ * Collect array of file mask. It also allow single special value
+ * for this option.
+ */
+function extract(value, memo) {
```
markelog: It would be better to not have any logic in bin file, would you mind moving it?

lib/checker.js
```diff
@@ -116,12 +151,19 @@ Checker.prototype._processDirectory = function(path, fileHandler) {
                 }
 
                 if (!this._hasCorrectExtension(fullname)) {
-                    return [];
+                    if (!this._configuration.shouldExtractFile(fullname)) {
+                        return [];
+                    }
+
+                    return this.extractFile(fullname).then(function(errors) {
+                        // extractFile could don't extract js from file
```
markelog: `could don't`?

lib/checker.js
```diff
@@ -147,14 +188,14 @@ Checker.prototype._processPath = function(path, fileHandler) {
 
         return vowFs.stat(path).then(function(stat) {
             if (stat.isDirectory()) {
-                return _this._processDirectory(path, fileHandler);
+                return this._processDirectory(path, fileHandler);
```
markelog: You really like `this`, huh? :-)

OVERVIEW.md
markelog: `Currently only `html` format...` what else is there? `xml`? or...?

markelog:
> For now I'm not sure we need for special values, i.e. always, auto and never.

:thumbsup:

lib/checker.js
lahmatiy: Just align to style in other methods.

markelog:
Would be cool if anyone else would check docs, /cc @hzoo, @mikesherov

OVERVIEW.md
lahmatiy: We discussed about extracting js from markdown or jsdocs, as example.

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
 
 In order to add/remove preset rules you will need to create a `.jscsrc` config file.
 
+### `--extract`
+With this option you could set file mask from which JavaScript should be extracted.
```
hzoo: ` you could set file ` -> ` you can set the file `

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
...
+```
+jscs path[ path[...]] --extract *.html 
+```
+
+You could set several masks if needed.
```
hzoo: `could set` -> `can set` 
`if needed` -> `if necessary`

lahmatiy:
Ok, I'm going to remove always/auto/never. It's simplify implementation.

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
...
+```
+jscs path[ path[...]] --extract *.html --extract *.htm
+```
+
+Or you could you one of shortcut value: `always`, `auto` or `never`. Note that those values couldn't be combined with file masks.
```
hzoo: `you could you one of` -> `you can use a`.

`Shortcut values cannot be combined with file masks`

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
...
+
+```
+jscs path[ path[...]] --extract always
+```
+
+Currently only `html` format is supported (JavaScript inside `<script>` will be checked). Auto fixing does not apply.
```
hzoo: Currently, only the `html` format is supported (JavaScript inside `<script>` will be checked) and extracted code cannot be auto fixed.

hzoo:
Yeah I'l look at the docs.

I guess I'm not familiar with the term - is it normally `file masks`? Does file extension work as well?

OVERVIEW.md
```diff
@@ -148,6 +148,25 @@ jscs path[ path[...]] --preset=jquery
 
 In order to add/remove preset rules you will need to create a `.jscsrc` config file.
 
+### `--extract`
+With this option you can set file mask from which JavaScript should be extracted.
```
hzoo: Sorry I updated my comment - `set file mask` -> `set the file mask`. Actually the use of `from which` should probably be changed. Maybe `This option checks javascript style in the specified file extensions (currently only html)` or something like that.

lahmatiy:
Well, I may be wrong about `file masks`. Actually I mean `glob patterns` as I can see in `excludeFiles`.

OVERVIEW.md
lahmatiy: I don't how to describe it right :-/
It's not about file extensions, but glob patterns, I guess.

lahmatiy:
Done! No more always/auto/never :)

lahmatiy:
Feature looks completed.
Should I do anything else?
 /cc @markelog @mikesherov

markelog:
Okay, i see there is couple artefacts of the previous logic and room for improvement, but we can deal with that later if that would be an issue

markelog:
Would you mind squashing your commits?

lahmatiy:
Done

markelog:
Hooray! 

Many thanks @lahmatiy, this was truly epic patience test!

hzoo:
Truly :tada:

lahmatiy:
\o/ it's finally merged :)

Issue #779, opened 2014-11-19
Problems with require-padding-newlines-before-keyword and function keyword


jdlrobson:
There are various cases where this rule when used with function doesn't make sense to me. To capture the issue I've written a bunch of patches https://github.com/jdlrobson/node-jscs/tree/upstream/require-padding-newlines-before-keyword

Could you let me know if I am right to modify the existing config (it doesn't introduce any issues with the existing tests) or if I should write this as a new config option...

Issue #511, opened 2014-07-07
Disallow space before/after dot operator


cowwoc:
I'd like to require dots to show up at the end of lines, as opposed to the beginning of new lines.

```
// Good
a.b.
  c()

// Bad
a.b
  .c();
```

And I'm guessing other people will want the opposite. I scanned the existing rules but couldn't find anything that would allow this. Did I miss something or do we need a new rule?

mrjoelkemp:
Looks like the original issue was fixed but now there's a lingering bug. Safe to open a new issue with the bug report and close this feature request?

zxqfox:
:+1: for making a separate issue about the bug.

Also, looks like rule can be simplified now with new tokens API.

mrjoelkemp:
Issue opened https://github.com/jscs-dev/node-jscs/issues/1437

PR #1048, opened 2015-02-12
New Rule: requireDollarBeforejQueryAssignment

### LINKED ISSUE #1049: Rule to require a $ prefix on jQuery variables
Many style guides have this rule:

```javascript
/**
 * Require a $ before variable names that are jquery assignments.
 *
 * Type: `Boolean`
 *
 * Values: `true`
 *
 *
 * #### Example
 *
 * ```js
 * "requireDollarBeforejQueryAssignment": true
 * ```
 *
 * ##### Valid
 *
 * ```js
 * var $x = $(".foo");
 * ```
 *
 * ##### Invalid
 *
 * ```js
 * var x = $(".foo");
 * ```
 */
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8507829-rule-to-require-a-prefix-on-jquery-variables?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
I was wondering if anyone had any recommendations on better ways to implement this. I'm looking for both stylistic / iterator optimizations to just improve the code that is here, as well as larger questions to actually match more cases that would cause a jQuery object to be returned.

For example, this only catches a subset of the times you could be creating a jQuery object since is stops validating if you ever chain off a jQuery object during assignment For example `$('.foo').first()` still should fail this rule because `.first` returns a jQuery object, but `$('.foo').val()` wouldn't because `.val` returns a value.

Fixes #1049


lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -0,0 +1,100 @@
...
+                var props = token.properties[0];
+                varName = props.key.name;
+                right = props.value;
+            }
+
+            if (varName[0] === '$') {
+                return;
+            }
+
+            if (right.type === 'ObjectExpression') {
```
mrjoelkemp: Join condition with previous condition and comment explaining why this condition matters. i.e., what's an example case that this catches?

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -0,0 +1,100 @@
...
+                return;
+            }
+
+            var nextToken = file.getTokenByRangeStart(right.callee.range[0]);
+            if (nextToken.value !== '$') {
+                return;
+            }
+
+            nextToken = file.getNextToken(nextToken);
+            if (nextToken.type !== 'Punctuator' && nextToken.value !== '(') {
```
mrjoelkemp: Using tokens instead of CallExpression nodes is what makes this implementation seem off, however, we've had a similar discussion in https://github.com/jscs-dev/node-jscs/issues/779 and settled on tokens.

mrjoelkemp:
@TheSavior Awesome rule! Nice work.

How many jquery functions would you have to add that return jquery objects? 

We'd typically maintain an internal list and then your rule would just make sure that the last function in the chain was a member of that list. To get the "last function in the chain" would be a nice helper that could also help with https://github.com/jscs-dev/node-jscs/issues/511.

mikesherov:
I'll echo @mrjoelkemp here and say great work, however, I think this rule should be using nodes rather than tokens.

I can comment more fully in the morning but just wanted to jot down that note.

TheSavior:
It's been a while since I started this, but I remember having extreme trouble doing this with nodes and started going down this path. It has since become way more pain. I'm running this through a really large codebase and there are way more things that I missed as tests.

TheSavior:
Any ideas why coveralls hasn't run for this?

TheSavior:
Guessing that coveralls only runs on things that have attached issues.

Fixes #1049

mrjoelkemp:
@TheSavior Not sure. Mind rebasing your branch off of master, squashing commits, and then force pushing up to trigger a rebuild?

TheSavior:
Yeah, I'm about to push up a new version with a ton more tests

TheSavior:
Hmm. Still no Coveralls.

mrjoelkemp:
@TheSavior For now, you can do `npm run coverage` or `npm run coverage-html` for coverage data.

mrjoelkemp:
Coveralls obviously doesn't like this rule. Closing the PR :) :)

TheSavior:
O hai coveralls. Pleasure to see you.

mrjoelkemp:
@TheSavior Landed! Sorry this took so long. Nice work!

PR #1050, opened 2015-02-12
crockford.json: Enable requireSpaceBeforeObjectValues

jacksonrayhamilton:
Hey there,

I believe this rule is missing. Pretty sure Crockford always puts at least one space between the colon and the value.

### DIFF
--- a/presets/crockford.json
+++ b/presets/crockford.json
@@ -62,5 +62,6 @@
     "requireCapitalizedConstructors": true,
     "requireDotNotation": true,
     "disallowNewlineBeforeBlockStatements": true,
-    "disallowMultipleLineStrings": true
+    "disallowMultipleLineStrings": true,
+    "requireSpaceBeforeObjectValues": true
 }
--- a/test/data/options/preset/crockford.js
+++ b/test/data/options/preset/crockford.js
@@ -48,7 +48,7 @@ div.onclick = function (e) {
 };
 
 obj = {
-    method: function () {
+    method: function () { // requireSpaceBeforeObjectValues
         return this.datum; // requireDotNotation
     },
     datum: 0


mrjoelkemp:
Hey @jacksonrayhamilton. Thanks for contributing! Can you provide a few references to prove that this should be added to the crockford preset? Folks are pretty touchy about this preset, so we need to make sure the addition won't start a war :)

alawatthe:
Crockford uses this style in a code example on his [Code Conventions page](http://javascript.crockford.com/code.html). However, he does not mention whether one should use a space or not. 

In his projects he uses the style throughout, e.g.:
https://github.com/douglascrockford/JSLint/blob/master/jslint.js#L287-L537
https://github.com/douglascrockford/JSON-js/blob/master/json.js#L388-L396
https://github.com/douglascrockford/JSCheck/blob/master/jscheck.js#L455-L467

I haven't found a single place in his code, where he is not using it. (But of course, I might have missed some.)

Therefore I think we can include the rule.

mrjoelkemp:
Thanks for the references @alawatthe. 

We also add examples for each rule to our fixture file associated with the preset. Namely, this file https://github.com/jscs-dev/node-jscs/blob/master/test/data/options/preset/crockford.js should have additions for the new rule.

I'm also interested to see if https://github.com/douglascrockford/JSLint/blob/394bf291bfa3881bb9827b9fc7b7d1112d83f313/jslint.js#L307 breaks the tests.

Please update the fixture.

jacksonrayhamilton:
http://jslint.com/:

Try this snippet:

```js
var a = {
    a:5
};
```

And you get this:

```
Missing space between ':' and '5'.
a:5
```

mrjoelkemp:
LGTM

markelog:
Since next release should be 1.12 it's safe to landed

markelog:
Thank you

Issue #986, opened 2015-02-01
Execution without options


markelog:
If project doesn't have jscs config and If jscs cli tool is executed without options but with path argument - `jscs <files>`. What should we do?

Two possible ways:
1. Apply rules from "default" preset, like with `go fmt`, `standart` npm <a href="https://github.com/feross/standard">package</a> or with python pep8 <a href="http://pep8.readthedocs.org/en/latest/intro.html">tool</a>. While proposing to use auto-config option at the end of error list.
2. Enable auto-config option by default.

zxqfox:
I think auto-config by default is not a good idea. You can easily mistype files or forget to put paths and you'll get wrong config.

3. Tell user that there are no rules to check and give him a note how to configure with preset or autoconfigure jscs.

But (1) is pretty fine too.

mdevils:
I like the current behaviour.

hzoo:
Also :-1: for the default preset for me. We could make the error message with no config better like explaining autoconfigure or using the preset flag.

ESLint has many default rules setup and they are actually planning on moving to have no rules turned on by default (many complaints about what should be in by default, etc). They just added a feature to extend other people's configs (can publish `eslint-config-x`) http://eslint.org/docs/developer-guide/shareable-configs which is pretty cool (like our presets, except it's by the user rather than ESLint itself).

markelog:
That is because eslint added and removed default rules with each new version, which is irritating, we good here, because we have strict semver policy, so in our case, if we would add or remove rules that cause more errors, we would do that only in major.

markelog:
Defining your own config is hard even if you define only online like `"preset": ...`, since you would need to know the rules of that preset or at least have basic idea.

But if we would use default preset, people wouldn't need to think about that stuff, they just would start working, this is main advantage of the https://github.com/feross/standard package, this is main reason why people used it, i think we should listen the community and plan accordingly

mrjoelkemp:
> Defining your own config is hard, even if you define only online like "preset": ..., since you would need to know the rules of that preset or at least have basic idea of that guide.

+1, though the barrier to entry is high only for preset authors. We have the presets at the top of the Overview section on the website, so we've at least attempted to make it easy to get started. 

What would the default preset be? I'd vote for `airbnb` as it seems to be the most mainstream.

mikesherov:
[...]

PR #1053, opened 2015-02-12
Config: Add "ignore-missing-config" option

jacksonrayhamilton:
Currently, `jscs` exits with a non-zero status code when it can't find a config file.

This is great for certain use cases, e.g. as a reminder for when you have forgotten to add a `.jscsrc` file to a project directory.

But for other situations, e.g. when you have configured your editor to send any opened "*.js" file to `jscs`, and your editor warns you when the program exits with a non-zero status code, then it can be quite annoying to try and edit files from projects which do not use JSCS, as you will be bombarded with "No configuration found." messages.

Trivially, this is the particular case for a [GNU Emacs](http://www.gnu.org/software/emacs/) + [Flycheck](https://github.com/flycheck/flycheck) setup defined as follows in a user's `~/.emacs` file:

```lisp
(flycheck-define-checker javascript-jscs
  "JSCS - JavaScript Code Style"
  :command ("jscs" "--reporter=checkstyle" source-inplace)
  :error-parser flycheck-parse-checkstyle
  :modes (js-mode js2-mode js3-mode))

(add-to-list 'flycheck-checkers 'javascript-jscs)
```

One workaround is to `echo '{}' >> ~/.jscsrc`, but it's not a very good solution, especially if a user doesn't know to do that. This option allows you to no-op your way through the program if a config file is not found.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -47,7 +47,7 @@ module.exports = function(program) {
      * Trying to load config.
      * Custom config path can be specified using '-c' option.
      */
-    if (!config && !program.preset && !program.autoConfigure) {
+    if (!config && !program.preset && !program.autoConfigure && !program.ignoreMissingConfig) {
         if (program.config) {
             console.error('Configuration source', program.config, 'was not found.');
         } else {
--- a/test/cli.js
+++ b/test/cli.js
@@ -121,6 +121,18 @@ describe('modules/cli', function() {
         console.error.restore();
     });
 
+    it('should not exit if missing custom configs are ignored', function() {
+        var result = cli({
+            args: ['test/data/cli/success.js'],
+            config: 'config.js',
+            ignoreMissingConfig: true
+        });
+
+        assert(typeof result === 'object');
+
+        return assertNoCliErrors(result);
+    });
+
     it('should set presets', function() {
         var Checker = require('../lib/checker');
         var originalCheckPath = Checker.prototype.checkPath;
--- a/test/cli.js
+++ b/test/cli.js
@@ -121,7 +121,7 @@ describe('modules/cli', function() {
         console.error.restore();
     });
 
-    it('should not exit if missing custom configs are ignored', function() {
+    it('should not exit with error status if missing config is ignored', function() {
         var result = cli({
             args: ['test/data/cli/success.js'],
             config: 'config.js',
--- a/bin/jscs
+++ b/bin/jscs
@@ -16,6 +16,7 @@ program
     .description('A code style linter for programmatically enforcing your style guide.')
     .option('-c, --config [path]', 'configuration file path')
     .option('--auto-configure [path]', 'auto-generate a JSCS configuration file')
+    .option('-i, --ignore-missing-config', 'if configuration file is not found, don\'t complain')
     .option('-e, --esnext', 'attempts to parse esnext code (currently es6)')
     .option('--es3', 'validates code as es3')
     .option('-s, --esprima <path>', 'attempts to use a custom version of Esprima')


test/cli.js
```diff
@@ -121,6 +121,18 @@ describe('modules/cli', function() {
         console.error.restore();
     });
 
+    it('should not exit if missing custom configs are ignored', function() {
```
zxqfox: custom configs → config, exit → exit with error status?

lib/cli.js
```diff
@@ -47,7 +47,7 @@ module.exports = function(program) {
      * Trying to load config.
      * Custom config path can be specified using '-c' option.
      */
-    if (!config && !program.preset && !program.autoConfigure) {
+    if (!config && !program.preset && !program.autoConfigure && !program.ignoreMissingConfig) {
```
zxqfox: Shouldn't we just exit(0) on `program.ignoreMissingConfig` without a config? Like...
```js
if (program.ignoreMissingConfig) {
  defer.resolve();
  return returnArgs;
}
```

zxqfox:
Thanks for contribution! Please, make an issue for PR next time ;-)

lib/cli.js
jacksonrayhamilton: It really depends on how we want this option to behave. What it does currently is true to the option's name: It just ignores the fact that a config is missing, and everything else proceeds as normal.

- This has the advantage that if you always expect an output (especially if you specified a reporter), you will still get an output, as opposed to getting nothing.
- This has the disadvantage of being "a waste of time," and arguably "incorrect." As there are no rules, there is nothing to check, so why claim that we checked anything? (Though [line 81](https://github.com/jacksonrayhamilton/node-jscs/blob/79ef83206549b902f3e2ba25352c0a8675342812/lib/cli.js#L81) doesn't exactly support that argument.)

With `exit(0)`, the behavior would be more consistent with what normally happens when you don't have a config file: The program doesn't do any checking. And with the option enabled, then it wouldn't print an error message either.

In that case, a more correct name would be `--suppress-missing-config-error`. That implies that there is no message, and that there is no erroneous exit code, but it does not imply that we "skip over" missing configs as "ignore" would suggest (rather, we still "stop on them").

zxqfox: Yeah, you're right. Thin difference.

jacksonrayhamilton: How should we move forward? I am fine with either route.

zxqfox: /cc @mikesherov @mrjoelkemp

jacksonrayhamilton: (bump)

markelog:
I'd say is pretty much an edge use-case - to much for addining a new option, it seems such problem should be resolved on plugin side.

jacksonrayhamilton:
Such errors could certainly could be resolved by plugins, but in this particular case (and perhaps other cases, if they exist) it would be much easier to make JSCS more flexible.

Flycheck does offer an API for parsing shell command output, so you'd imagine I could regex for your error message and ignore it as necessary. But even if I return an empty list of errors in the case of matching `"^No configuration found\\. Add a \\.jscsrc file to your project root or use the -c option\\.$"`, Flycheck always finds it suspicious when a checker program exits non-zero without errors (as it probably should). [(Source.)](https://github.com/flycheck/flycheck/blob/e67c502d381dd0ca5c82f0c2d86a26fc2afeee26/flycheck.el#L4361)

So I actually have to add a new feature to Flycheck to get it to ignore the JSCS error:

```lisp
(defconst flycheck-javascript-jscs-no-configuration-error-regexp
  (concat "^No configuration found\\. Add a \\.jscsrc file to your project "
          "root or use the -c option\\.$")
  "Match the dreaded jscs error.")

(defvar flycheck-ignorable-outputs
  `((javascript-jscs . (,flycheck-javascript-jscs-no-configuration-error-regexp)))
  "Map checkers to regexps specifying ignorable outputs.")

(defun flycheck-determine-output-ignorability (output checker)
  "Determine if OUTPUT for CHECKER can be ignored."
  (let ((regexps (cdr (assq checker flycheck-ignorable-outputs)))
        regexp
        ignorable)
    (while (and regexps (not ignorable))
      (setq regexp (car regexps))
      (when (string-match regexp output)
        (setq ignorable t))
      (setq regexps (cdr regexps)))
    ignorable))

(defun flycheck-finish-checker-process
    (checker exit-status files output callback)
  "Finish a checker process from CHECKER with EXIT-STATUS.

FILES is a list of files given as input to the checker.  OUTPUT
is the output of the syntax checker.  CALLBACK is the status
callback to use for reporting.

Parse the OUTPUT and report an appropriate error status."
  (let ((errors (flycheck-parse-output output checker (current-buffer)))
        (ignorable (flycheck-determine-output-ignorability output checker)))
    (when (and (not ignorable) (/= exit-status 0) (not errors))
      ;; Warn about a suspicious result from the syntax checker.  We do right
      ;; after parsing the errors, before filtering, because a syntax checker
      ;; might report errors from other files (e.g. includes) even if there
      ;; are no errors in the file being checked.
      (funcall callback 'suspicious
               (format "Checker %S returned non-zero exit code %s, but no errors from \
output: %s\nChecker definition probably flawed." checker exit-status output)))
    (funcall callback 'finished
             ;; Fix error file names, by substituting them backwards from the
             ;; temporaries
             (mapcar (lambda (e) (flycheck-fix-error-filename e files))
                     errors))))
```

I could try to convince the author of Flycheck to add the above code, plus an API for specifying ignorable errors for checkers. But now I'm hopping between library authors, trying to get one person to make his program compatible with someone else's. Making JSCS more configurable in this regard seems to require a lot less JS than the alternative amount of LISP.

markelog:
> than the alternative amount of LISP

This is why js better! jk. Don't do regexes, orient on the exit code

jacksonrayhamilton:
It's not as simple as just ignoring a non-zero exit code (which also requires modification of Flycheck). At least, it's not desirable, as I'd still like to know about other errors. e.g.:

```bash
$ echo '{"a":,}' >> .jscsrc
$ jscs foo.js
Config source is corrupted - SyntaxError: Unexpected token ,
```

So I'd still want my checker plugin to alert me, as it does:

<p align="center"></p>

So I do need to match some strings to determine which errors to ignore.

Being able to suppress the missing config error would eliminate this necessity.

This feature is probably useful outside of my particular case. The Atom JSCS plugin has actually added a feature [to do nothing when a config file isn't found](https://github.com/AtomLinter/linter-jscs/blob/ff783059af77e79338354e88e95e7a05b67ca3dc/lib/linter-jscs.coffee#L66). (Though this doesn't do much more than suppress `console.warn`'s in debugging mode, as the linter package normally silences all errors. Even important ones, like the malformed JSON one.)

The Sublime Text JSCS plugin also seems to swallow errors, though it does have [a fast track for empty output](https://github.com/SublimeLinter/SublimeLinter3/blob/b87d3e41b8680e4d5c072b2840e6983d26126622/lint/linter.py#L1313), so you could speed things up for free by taking the fast track.

Even though some linter plugins are *generous* enough to swallow all errors (even important errors), luckily some plugins are more suspicious. You should assist suspicious plugins in white-listing unimportant errors (unimportant in the context of blindly checking files with a ".js" extension, that is) by providing this flag.

This addition is a meager and worthwhile price to pay to add another program to your "Friendly packages" list and to simplify plugin integration.

markelog:
The problem with adding additional options with very limited use that it might clutter existing API. 

When some API provides a lot of options is starts perceived as complex, when some API provides big docs (like with `jscs -h`) people stop reading it, especially, when good amount of those options do not fit with most of use cases, especially, when similar tools don't know about that user story or when users might not know that this is theirs user story.

If you don't implement your own config search (which is, i think, is good way to go), then currently, you might have to navigate through existing exit codes. 

Yes, that is not a perfect solution, since, as you mentioned, user might not immediately notice that their config is faulty. But i think it's best thing to do, at the moment. 

However, we can improve on it by providing additional exit code for every faulty situation, so user user can differentiate by them, unix way and all that, but this change only can happen in major version.

Could you settle on hack now and after 2.0 will come out issue will disappear by itself?

jacksonrayhamilton:
I have already hacked my way through this. Personally I could continue npm linking my fork and not have to deal with this error. I just thought other Flycheck users would find it valuable if they didn't need to google their way to my fork, or have to discover some blog post that instructs them to copy and paste that gigantic chunk of Elisp, just to get JSCS to work. Or to avoid a hack like "sudo echo '{}' >> /.jcscrc". IMO, we shouldn't have to resort to hacks, or have to *reimplement* features like config searching. The program should just work out of the box without giving us a hard time.

I really think it would be a good thing to add this, because if you don't, and if I can't convince the Flycheck author to "clutter" *his* API with an "error message whitelist", then this problem will remain a barrier to entry for Emacs users.

markelog:
I'm not sure if understand you, differentiation with exit codes would solve you problem?

mikesherov:
Right, if we exited with code `127` for example when no config was found,
then that's a common way all consumers of JSCS can choose to ignore that
error. This is a fairly standard unix principle.

On Fri, Feb 20, 2015 at 2:26 PM, Oleg Gaidarenko <notifications@github.com>
wrote:

> I'm not sure if understand you, differentiation with exit codes would
> solve you problem?
>
> --
> <https://github.com/jscs-dev/node-jscs/pull/1053#issuecomment-75303012>.
>



-- 
Mike Sherov

markelog:
In any case it would be to have it - https://github.com/jscs-dev/node-jscs/issues/1095

jacksonrayhamilton:
I don't think exit codes would make things any better. I'd still need to add support to Flycheck to check for particular exit codes. All it does now is check for non-zero exit codes.

I don't think 127 is a correct or good code to emit. "command not found" is different from "config not found". "Command not found" implies that you forgot to do `npm install -g jscs`.

If we had to choose *some* exit code, the closest would probably be 78, "configuration error." But that is still ambiguous: Is it a "configuration not found" error, or a "malformed JSON" error? You are still going to need error messages when the standard set of error codes runs out.

markelog:
> I'd still need to add support to Flycheck to check for particular exit codes

This is seems as most promising and correct way.

> I don't think 127 is a correct 

We can debate on code numbers in relevant issue, the main thing is that we can assign different exit codes on different situations.

mikesherov:
> I don't think exit codes would make things any better. I'd still need to add support to Flycheck to check for particular exit codes. All it does now is check for non-zero exit codes.

Flycheck would need to be updated to correctly support JSCS in either situation. In any case, either you or your users would have to know to implement the config option you propose. So, you're doing work one way or the other!

> I don't think 127 is a correct or good code to emit. "command not found" is different from "config not found". "Command not found" implies that you forgot to do npm install -g jscs.

I was using 127 as a "for instance". 78 is better.

> But that is still ambiguous: Is it a "configuration not found" error, or a "malformed JSON" error? You are still going to need error messages when the standard set of error codes runs out.

No one is suggesting to ditch the error messages. We're saying add and document them so that tools can choose to ignore certain errors at their leisure. This is the purpose of exit codes.

jacksonrayhamilton:
Exit codes won't make things any better.

All linter plugins will need to be updated to support particular exit codes. None of the various linters, including Atom, Sublime, Webstorm and Emacs make any distinction between the type of errors thrown by a linter program. Most of them just try, catch, and ignore. Additionally, that error checking is usually handled at a lower level, and configuring new linters is often accomplished via inheritance or configuration. This increases the difficulty of adding particular exit code handling support across the board.

78 is not sufficiently specific, especially in this particular case, because I don't want to be alerted about missing configs, but I do want to be alerted about malformed configs, and using 78 would seem to imply the code should be used for either both, or neither case.

You could start making up new exit codes, and you could certainly document them, but general-purpose linter programs are not going to support your arbitrarily-defined set of exit codes at a low level, because those arbitrarily-defined codes could vary from linter to linter. So they have to handled by the JSCS integration plugins, if that is even possible with the base linter plugin's API. If it was possible, that'd probably be a good thing, because the user could get more detailed feedback on issues, but it is also going to require more code.

On the other hand, Flycheck does not need to be updated to support a new CLI flag. In fact, no linter plugins need to be updated to support this. All linter plugins are already capable of adding support for a new option. Without modifying the source code of flycheck.el, I can already do this to get the behavior I want:

```lisp
(flycheck-define-checker javascript-jscs
  "JSCS - JavaScript Code Style"
  :command ("jscs" "--reporter=checkstyle" "--ignore-missing-config" source-inplace)
  :error-parser flycheck-parse-checkstyle
  :modes (js-mode js2-mode js3-mode))
```

So it is not true that "Flycheck would need to be updated to correctly support JSCS in either situation." Although it certainly might be convenient if the author of Flycheck added the above snippet so that checker support was built-in, any user can set up JSCS without any additional hacks with the above code.

mikesherov:
> So it is not true that "Flycheck would need to be updated to correctly support JSCS in either situation." Although it certainly might be convenient if the author of Flycheck added the above snippet so that JSCS support was built-in, any user can set up JSCS without any additional hacks with the above code.

That's true, and a good point. I think we've sufficiently beaten this horse dead. I'm fine with adding the new option to make it easier on our integrations.

@mdevils, @zxqfox thoughts?

mikesherov:
@jacksonrayhamilton thanks for sticking with this as we talked it through, by the way.

markelog:
> those arbitrarily-defined codes could vary from linter to linter

I could say the same about your option, there is already established way for handling these kind of situations and that is an exit codes, it's been like this for years and years before us and we should stick to it then making up our own ways to workaround that.

You see only how it would benefit to you, but some reason ignoring my points above. I think we discuss this long enough, so i'm closing this.

mikesherov:
@markelog, I also know my coworkers have been asking for this as an option as well, and we had other tickets about this.

I literally just told @jacksonrayhamilton that I wanted to here from @zxqfox and @mdevils here and you close it? What gives?

markelog:
> I also know my coworkers have been asking for this as an option as well, and we had other tickets about this

If that so, then this is a duplicate so it should be closed, also could you provide the links?

> I literally just told @jacksonrayhamilton that I wanted to here from @zxqfox and @mdevils 

How issue status preventing them from sharing their opinions? If we decide to go another way, we can always re-open.

> What gives?

I can redirect the same question to you, i just close this, what gives? Please leave this closed.

mikesherov:
@markelog, because you also closed the other one from my coworker without a chance for discussion!

Anyway, I'll reclose this, but we need to an issue that doesn't presuppose that exit codes is the right way.

There are 4 options:

1. Make JSCS not complain by default (bad).
2. Make JSCS configurable to not complain about configuration. (ironic)
3. Use fine grained exit codes (most of Unix behaves this way)
4. Separate out warnings from errors and consider no config as a warning and introduce an option to suppress warnings.
5. Do nothing and have packages do the heavy lifting.

I still think we need @mdevils and @zxqfox opinion here, but going to reclose this PR for now.

zxqfox:
@mikesherov This issue forces me to search `sysexits` codes. Is it time to cleanup? We can freely implement and document exit codes but there are some issues like #962 for v2.

- Sysexists #1095
- Obey of silence #962
- Execution without options #986
- etc.

Is #962 the same as `--ignore-missing-config`? If so we can close this issue.

Probably would be great to collect all of this cli issues into another one and talk about cli behaviour there.

markelog:
> @markelog, because you also closed the other one from my coworker without a chance for discussion!

Could you post a ticket number? My apologies if i did, if you find my action too hasty don't hesitate to ping me in IRC or somewhere else next time.

> Is #962 the same as --ignore-missing-config

Doesn't look like a duplicate, point of that issue is to do be quite only if config is there and there is no errors to show.

I also like 3, not sure about the 4.

zxqfox:
> Doesn't look like a duplicate, point of that issue is to do be quite only if config is there and there is no errors to show.

Flag `-s` makes grep silent. Probably we need something similar.

```
$ sh -c 'grep asd .jscsrc 2>/dev/null'; echo $?
1
$ sh -c 'grep asd asd 2>/dev/null'; echo $?
2
$ sh -c 'grep -s qwe asd'; echo $? # nice.
2
$ sh -c 'jscs -c asd lib 2>/dev/null'; echo $?
1
$ sh -c 'jscs -c .jscsrc zxc 2>/dev/null'; echo $?
1
```

jacksonrayhamilton:
I do think that exit codes would be good in principle, but currently they would only benefit a single linter out of the set of linter packages you support.

Here's how they each handle errors:

- Atom: [console.warns all errors](https://github.com/AtomLinter/Linter/blob/bb877d3cce4080d787123dc685e361049a1ae745/lib/linter.coffee#L137)
- Brackets: [Doesn't use CLI](https://github.com/globexdesigns/brackets-jscs/blob/b6cc9ff827dc511b0e15c23ad4a14b77ab1b8836/main.js#L99)
- Grunt: [Doesn't use CLI](https://github.com/jscs-dev/grunt-jscs/blob/8e17966c67239a82700713a1dd7225313ca47e5f/tasks/jscs.js#L21)
- Gulp: [Doesn't use CLI](https://github.com/jscs-dev/gulp-jscs/blob/5f1c5837ca95dba8ea40dce83606190a68ab9345/index.js#L53)
- Sublime: [Doesn't handle exit codes](https://github.com/SublimeLinter/SublimeLinter3/blob/b87d3e41b8680e4d5c072b2840e6983d26126622/lint/util.py#L1356)
- Vim: [Does handle exit codes](https://github.com/scrooloose/syntastic/blob/1d2f8d514a3a03fba28d46a9072817e2777bec00/syntax_checkers/javascript/jscs.vim#L34)
- Visual Studio: [Doesn't use CLI](https://github.com/madskristensen/WebEssentials2013/blob/30d34214bb2ca25550d5d5987ddb70a757d41791/EditorExtensions/Resources/server/services/srv-jscs.js#L63)
- IntelliJ et al: [Exit code handling is commented-out](https://github.com/idok/jscs-plugin/blob/43bca7e945f6d781567dac5fe55eb3ef297abcf0/src/com/jscs/utils/JscsRunner.java#L43)
- and Flycheck: [Doesn't differentiate exit codes](https://github.com/flycheck/flycheck/blob/e67c502d381dd0ca5c82f0c2d86a26fc2afeee26/flycheck.el#L4371)

So your Vim users might appreciate (3), but it won't have any impact on the other linter packages without significant additions to those packages' APIs. I probably sound like a broken record at this point, but unless you plan on submitting a lot of PRs to improve the above situation, I recommend against creating an exit code lexicon.

I would be fine with (1), (2) or (4), since any of those solve my particular problem.

PR #1055, opened 2015-02-13
Docs: add reporter

sanemat:
document says "six reporters", but shown only five.

$ ls lib/reporters/
checkstyle.js console.js inline.js inlinesingle.js junit.js
text.js


markelog:
Thank you

sanemat:
:bowtie:

PR #1056, opened 2015-02-13
Disallowing do..while on new line

### LINKED ISSUE #885: Disallowing do..while on new line
I would like to have this restriction:

```javascript
do {
  ...
} while (...);
```

The `while` used with `do` should be placed in the same line of the closing curly brace. However, this cannot be defined with the option `disallowKeywordsOnNewLine` because a single `while` would be marked as invalid.

```javascript
// This gives an error: Keyword 'while' should not be placed on a new line
while (...);
```

How can I require a do..while on the same line but allow single whiles on a new line?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7544391-disallowing-do-while-on-new-line?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oredi:
1.	Allow while that is not part of a do to succeed even though it contradicts rule meaning
2.	Allow while that is part of a do to succeed for do whiles with no braces

Fixes #885 - issue

### DIFF
--- a/lib/rules/disallow-keywords-on-new-line.js
+++ b/lib/rules/disallow-keywords-on-new-line.js
@@ -53,8 +53,14 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         var keywordIndex = this._keywordIndex;
+        // do counter to handle #885
+        var openDoLoop = 0;
 
         file.iterateTokensByType('Keyword', function(token) {
+            if (token.value === 'do') {
+                openDoLoop++;
+            }
+
             if (keywordIndex[token.value]) {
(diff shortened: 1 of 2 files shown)


oredi:
Sorry I found an issue with the usage of an int counter for the Do Loop opened reference. I am testing a different implementation. I will squash the commits into a single commit and create a new PR when I am done.

PR #1057, opened 2015-02-13
disallowSpacesInsideArrayBrackets: fix error messages

hzoo:
@zxqfox

### DIFF
--- a/lib/rules/disallow-spaces-inside-array-brackets.js
+++ b/lib/rules/disallow-spaces-inside-array-brackets.js
@@ -109,8 +109,7 @@ module.exports.prototype = {
                 errors.assert.noWhitespaceBetween({
                     token: openBracket,
                     nextToken: afterOpen,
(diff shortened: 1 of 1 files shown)


lib/rules/disallow-spaces-inside-array-brackets.js
```diff
@@ -110,7 +110,7 @@ module.exports.prototype = {
                     token: openBracket,
                     nextToken: afterOpen,
                     spaces: 1,
```
zxqfox: Does this `spaces` number mean something?

hzoo: No, yeah it doesn't mean anything for `noWhitespaceBetween`, only for `whitespaceBetween`.

hzoo:
Ok went ahead and removed the extraneous `spaces` option as well

PR #1058, opened 2015-02-14
Autofixing: initial implementation

mdevils:
**Work in progress**

In this pull-request I have implemented autofixing mechanism:

* Flag `-x, --fix` does autofixing.
* In case of `stdin` input, autofixed content is written to the console.
* I tried to keep everything as simple as possible.
* I will write tests later.
* I will cover other assertion methods soon.
* All unfixed errors are printed.

@mikesherov @markelog @mrjoelkemp @zxqfox please take a look!

### DIFF
--- a/lib/checker.js
+++ b/lib/checker.js
@@ -35,13 +35,32 @@ Checker.prototype.configure = function(/*config*/) {
  * @returns {Promise * Errors}
  */
 Checker.prototype.checkFile = function(path) {
-    if (!this._configuration.isFileExcluded(path)) {
-        return vowFs.read(path, 'utf8').then(function(data) {
-            return this.checkString(data, path);
-        }, this);
+    if (this._configuration.isFileExcluded(path)) {
+        return Vow.resolve(null);
     }
 
-    return Vow.resolve(null);
+    return vowFs.read(path, 'utf8').then(function(data) {
+        return this.checkString(data, path);
+    }, this);
+};
+
+/**
+ * Fixes single file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
+ */
+Checker.prototype.fixFile = function(path) {
+    if (this._configuration.isFileExcluded(path)) {
+        return Vow.resolve(null);
+    }
+
+    return vowFs.read(path, 'utf8').then(function(data) {
+        var result = this.fixString(data, path);
+        return vowFs.write(path, result.output).then(function() {
+            return result.errors;
+        });
+    }, this);
 };
 
 /**
@@ -51,6 +70,37 @@ Checker.prototype.checkFile = function(path) {
  * @returns {Promise * Error[]}
  */
 Checker.prototype.checkDirectory = function(path) {
+    return this._processDirectory(path, this.checkFile.bind(this));
+};
+
+/**
+ * Checks directory or file.
+ *
+ * @param {String} path
+ * @returns {Promise * Error[]}
+ */
+Checker.prototype.checkPath = function(path) {
+    return this._processPath(path, this.checkFile.bind(this));
+};
+
+/**
+ * Fixes directory or file.
+ *
+ * @param {String} path
+ * @returns {Promise * Error[]}
+ */
+Checker.prototype.fixPath = function(path) {
+    return this._processPath(path, this.fixFile.bind(this));
+};
+
+/**
+ * Processes directory recursively.
+ *
+ * @param {String} path
+ * @param {Function} fileHandler
+ * @returns {Promise * Error[]}
+ */
+Checker.prototype._processDirectory = function(path, fileHandler) {
     return vowFs.listDir(path).then(function(filenames) {
         return Vow.all(filenames.map(function(filename) {
             var fullname = path + '/' + filename;
@@ -61,14 +111,14 @@ Checker.prototype.checkDirectory = function(path) {
 
             return vowFs.stat(fullname).then(function(stat) {
                 if (stat.isDirectory()) {
-                    return this.checkDirectory(fullname);
+                    return this._processDirectory(fullname, fileHandler);
                 }
 
                 if (!this._hasCorrectExtension(fullname)) {
                     return [];
                 }
 
-                return Vow.when(this.checkFile(fullname)).then(function(errors) {
+                return Vow.when(fileHandler(fullname)).then(function(errors) {
                     if (errors) {
                         return errors;
                     }
@@ -83,12 +133,13 @@ Checker.prototype.checkDirectory = function(path) {
 };
 
 /**
- * Checks directory or file.
+ * Processes directory or file.
  *
  * @param {String} path
- * @returns {Error[]}
+ * @param {Function} fileHandler
+ * @returns {Promise * Error[]}
  */
-Checker.prototype.checkPath = function(path) {
+Checker.prototype._processPath = function(path, fileHandler) {
     path = path.replace(/\/$/, '');
     var _this = this;
 
@@ -99,10 +150,10 @@ Checker.prototype.checkPath = function(path) {
 
         return vowFs.stat(path).then(function(stat) {
             if (stat.isDirectory()) {
-                return _this.checkDirectory(path);
+                return _this._processDirectory(path, fileHandler);
             }
 
-            return _this.checkFile(path).then(function(errors) {
+            return fileHandler(path).then(function(errors) {
                 if (errors) {
                     return [errors];
                 }
@@ -119,9 +170,26 @@ Checker.prototype.checkPath = function(path) {
  * @returns {Promise}
  */
 Checker.prototype.checkStdin = function() {
+    return this._processStdin(this.checkString.bind(this));
+};
+
+/**
+ * Fixes stdin input
+ *
+ * @returns {Promise}
+ */
+Checker.prototype.fixStdin = function() {
+    return this._processStdin(this.fixString.bind(this));
+};
+
+/**
+ *
+ * @param {Function} stdinHandler
+ * @returns {Promise}
+ */
+Checker.prototype._processStdin = function(stdinHandler) {
     var stdInput = [];
     var deferred = Vow.defer();
-    var _this = this;
 
     process.stdin.setEncoding('utf8');
 
@@ -130,9 +198,7 @@ Checker.prototype.checkStdin = function() {
     });
 
     process.stdin.on('end', function() {
-        var errors = _this.checkString(stdInput.join(''));
-
-        deferred.resolve(errors);
+        deferred.resolve(stdinHandler(stdInput.join('')));
     });
 
     return deferred.promise();
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -114,6 +114,15 @@ module.exports = function(program) {
         // So the dash doesn't register as a file
         if (usedDash) { args.length--; }
 
+        if (program.fix) {
+            return {
+                promise: checker.fixStdin().then(function(result) {
+                    process.stdout.write(result.output);
+                }),
+                checker: checker
+            };
+        }
+
         checkerPromise = checker.checkStdin().then(function(errors) {
             return [errors];
         });
@@ -121,7 +130,8 @@ module.exports = function(program) {
 
     // Processing specified files and dirs.
     if (args.length) {
-        checkerPromise = Vow.all(args.map(checker.checkPath, checker)).then(function(results) {
+        var method = program.fix ? checker.fixPath : checker.checkPath;
+        checkerPromise = Vow.all(args.map(method, checker)).then(function(results) {
             return [].concat.apply([], results);
         });
     }
--- a/lib/string-checker.js
+++ b/lib/string-checker.js
@@ -4,6 +4,8 @@ var Errors = require('./errors');
 var JsFile = require('./js-file');
 var Configuration = require('./config/configuration');
 
+var MAX_FIX_ATTEMPTS = 5;
+
 /**
  * Starts Code Style checking process.
  *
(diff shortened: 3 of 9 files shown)


hzoo:
Wow this is awesome! I just tried a simple test

```js
// var test = function(options){
//     for (var i=0;i<asdf.length;i++)
//         asdf();
//     x+ y;
// }

var test = function(options) {
    for (var i = 0; i < asdf.length; i++)
        asdf();
    x + y;
}
```

Fixes all whitespace errors and prints unfixed ones (indentation, curly braces)

lib/cli.js
```diff
@@ -114,14 +114,24 @@ module.exports = function(program) {
...
-            return [errors];
-        });
+        if (program.fix) {
+            return {
+                promise: checker.fixStdin().then(function(result) {
+                    process.stdout.write(result.output);
+                }),
+                checker: checker
+            };
+        } else {
```
zxqfox: If you returning in `if`, why to use `else`?

lib/token-assert.js
```diff
@@ -31,19 +31,23 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
         if (nextToken.loc.start.line === token.loc.end.line &&
             (nextToken.loc.start.column - token.loc.end.column) !== spaces
         ) {
+            nextToken.whitespaceBefore = new Array(spaces + 1).join(' ');
```
zxqfox: Is it right that we fixing code right in assert methods?

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+                        break;
+                    }
+                }
+
+                if (!hasFixes) {
+                    break;
+                }
+
+                source = file.render();
+                sourceTree = JsFile.parse(source, this._esprima, this._configuration.getEsprimaOptions());
```
zxqfox: What is the case to render/parsing multiple time?

zxqfox:
Looks awesome!

lib/string-checker.js
mikesherov: That the fix for one rule causes a violation of another.

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+
+                attempt++;
+            } while (attempt < MAX_FIX_ATTEMPTS);
```
mikesherov: Doing 5 attempts is suboptimal. We can likely do the following instead:

1. Instead of counting the number of attempts we made, count the number of violations remaining.

2. Keep looping as long as `fixes > 0` and `violations < violationsOfPreviousRun`. 

This ensures that we keep fixing until it's no longer helping, which will break a cycle after 1 iteration, not 5, and in the rare case where you'd have to do more than 5 iterations, it'll keep going until it's not helping.

lib/token-assert.js
mikesherov: Yes, but this modification should likely only happen if fixing is turned on.

Eventually, rules might actually be checking the whitespaceBefore property, and if we unconditionally change this whitespace, than overlapping rules will fail to report an error at the same location if we're not in fix mode.

lib/cli.js
mdevils: Correct structures are easy to read.

lib/token-assert.js
mdevils: I believe rules should not access `whitespaceBefore` property.

lib/string-checker.js
mdevils: > Keep looping as long as fixes > 0 and violations < violationsOfPreviousRun.

This might not work. For instance one rule does 1 fix and then there are 2 or more fixes needs to be done also.

lib/string-checker.js
zxqfox: But... I believe that we shouldn't allow checking with conflicting rules like requireSomething and disallowSomething at once. And at that POV this line looks like overkill.

lib/string-checker.js
mikesherov: You mean 1 fix increases violations for next run? I see.

What about the case where more than 5 runs are needed? I doubt such a case exists, so this is fine for now I suppose.

lib/string-checker.js
mikesherov: Not conflicting rules, complimentary rules. For example, our spacing rules typically ignore tokens that are on the next line.

If a "disallowNewLine" fix moves the token onto the same line as the previous token, it may now be in violation of a "disallowSpace" rule.

lib/cli.js
zxqfox: Incorrect structure are hard to parse ;-) Use of `return` makes an usual `if-else` harder to read. That's why I've asked ;-)

lib/token-assert.js
zxqfox: Yeah, +1 @mdevils, +1 @mikesherov for only if fixing is on.

mikesherov: Well, think of it this way. If we both modify whitespaceBefore and use it for checking, then we avoid a lot of cases where two full runs would be needed.

Imagine a "requireNewline" rule and a "requireSpace" rule that both report on a `{` at the end of a line...

If we had rules check the whitespaceBefore property, then a fix for requireNewline would make requireSpace ignore.

If we don't, then 2 passes must happen, one fixing the spacing, while the second pass finally moved it to the right spot on the next line.

mikesherov: Anyway, we can defer this discussion till later, especially consider this is all consolidated to one spot considering we have this nice assertion layer.

lib/string-checker.js
zxqfox: Feels like we need some specs for this ;-) for objective discussing and to get rid of "might".

lib/string-checker.js
zxqfox: Oh, right. Thanks. But should we really reparse? Probably there is a way to just recheck fixed tree. It's much harder, yes. Ok, nvm ;-(

mdevils: Fixing token without reparse it way harder.

lib/token-assert.js
mdevils: Yeah, I vote for deferring this discussion till later. I believe good assertion layer must cover 100% of cases.

lib/cli.js
mdevils: OK, I will fix this.

lib/token-assert.js
zxqfox: Agreed totally.

lib/string-checker.js
mdevils: I can't say I like `5` runs approach, but I think we can leave this for now and watch for user feedback.

mikesherov: @mdevils can you file an issue that we should revisit this later? I don't want it holding up the PR, but I don't want to lose track of the action item.

mdevils:
I think this can be merged now :)

mrjoelkemp:
Taking a look

lib/string-checker.js
mdevils: Here we go: #1062

lib/checker.js
```diff
@@ -45,12 +45,62 @@ Checker.prototype.checkFile = function(path) {
 };
 
 /**
+ * Fixes single file.
+ *
+ * @param {String} path
+ * @returns {Promise * Errors}
+ */
+Checker.prototype.fixFile = function(path) {
+    if (!this._configuration.isFileExcluded(path)) {
```
mrjoelkemp: Non-blocker: This style becomes pervasive throughout community contributions. What about opting for an early return?

```js
if (this._configuration.isFileExcluded(path)) {
  return Vow.resolve(null);
}

return vowFs.read(path, ...
```

The idea is along the lines of: http://sourcemaking.com/refactoring/replace-nested-conditional-with-guard-clauses

lib/string-checker.js
```diff
@@ -77,38 +79,50 @@ StringChecker.prototype = {
...
+     * Fills Errors instance with validation errors.
+     *
+     * @param {JsFile} file
+     * @param {Errors} errors
+     * @private
+     */
+    _checkJsFile: function(file, errors) {
+        var errorFilter = this._configuration.getErrorFilter();
+
+        if (!this._maxErrorsExceeded) {
```
mikesherov: perhaps make this an early return now that it's an `if` enclosing almost the entire body of the function?

lib/checker.js
```diff
@@ -45,12 +45,62 @@ Checker.prototype.checkFile = function(path) {
...
  */
 Checker.prototype.checkDirectory = function(path) {
+    return this._processDirectory(path, this.checkFile.bind(this));
+};
+
+/**
+ * Checks directory or file.
+ *
+ * @param {String} path
+ * @returns {Promise * Error[]}
```
mrjoelkemp: Non-blocker: I was wondering where this came from in https://github.com/jscs-dev/node-jscs/pull/1038/files#diff-605f23c7e3628b173a428e6a469e6ff6R52. 

Is that valid jsdoc?

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+            var errors;
+            var file;
+
+            do {
+                file = this._createJsFileInstance(filename, source, sourceTree);
+                errors = new Errors(file, this._verbose);
+                this._checkJsFile(file, errors);
+                var hasFixes = false;
+                var errorList = errors.getErrorList();
+                for (var i = 0; i < errorList.length; i++) {
```
mikesherov: I would encapsulate this as a method in error object: `hasFixes`. The `break` inside the code here is confusing because it breaks out of the `for` but not the `do`.

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+            es3: this._configuration.isES3Enabled(),
+            es6: this._configuration.isESNextEnabled()
+        });
+    },
+
+    /**
+     * Checks file provided with a string.
+     * @param {String} source
+     * @param {String} [filename]
+     * @returns {{output: String, errors: Errors}}
```
mrjoelkemp: I thought the jsdoc way to do this was:

```js
@returns {Object} data
@returns {String} data.output
@returns {Errors} data.errors
```

cc @zxqfox

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+        return new JsFile(filename, source, sourceTree, {
+            es3: this._configuration.isES3Enabled(),
+            es6: this._configuration.isESNextEnabled()
+        });
+    },
+
+    /**
+     * Checks file provided with a string.
+     * @param {String} source
+     * @param {String} [filename]
```
mrjoelkemp: `[filename='input']`

lib/string-checker.js
mdevils: This is a valid JsDoc: http://usejsdoc.org/tags-type.html

mrjoelkemp: :+1:

lib/checker.js
mdevils: This case is partially covered in JsDoc, so different projects use different notations: `{Promise ~ Type}`, others — `{Promise * Type}`. Following jsdoc documentation, the only close notation is `{Promise<Type>}`.

mrjoelkemp: :+1:

lib/string-checker.js
mrjoelkemp:
```js
var hasFixes = errorList.some(function(err) {
  return err.fixed;
})
```

lib/checker.js
zxqfox: Nope. https://github.com/jsdoc3/jsdoc/issues/509
The most logical for me is to use something like `{Promise.<Object,Error>}` or `{Promise}`

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+        if (parseError) {
+            var parseErrors = new Errors(this._createJsFileInstance(filename, source, sourceTree), this._verbose);
+            this._addParseError(parseErrors, parseError);
+            return {output: source, errors: parseErrors};
+        } else {
+            var attempt = 0;
+            var errors;
+            var file;
+
+            do {
```
mrjoelkemp: What in this `do` actually fixes the source? I see no explicit call to any fix-related function.

lib/string-checker.js
mdevils: This looks easier, thanks

lib/string-checker.js
mdevils: Assertions.

mikesherov: @mrjoelkemp checking the file automatically fixes the whitespace, which then is realized by rendering the file.

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+                        hasFixes = true;
+                        break;
+                    }
+                }
+
+                if (!hasFixes) {
+                    break;
+                }
+
+                source = file.render();
```
mikesherov: @mrjoelkemp here is where the fixes are reapplied back to the source, as a result of a parse on the next line.

lib/string-checker.js
zxqfox: Btw you can look at https://developers.google.com/closure/compiler/docs/js-for-compiler there are a lot of different examples for many cases.

lib/string-checker.js
mrjoelkemp: Thanks.

lib/string-checker.js
```diff
@@ -136,8 +150,87 @@ StringChecker.prototype = {
...
+            return {output: source, errors: parseErrors};
+        } else {
+            var attempt = 0;
+            var errors;
+            var file;
+
+            do {
+                file = this._createJsFileInstance(filename, source, sourceTree);
+                errors = new Errors(file, this._verbose);
+                this._checkJsFile(file, errors);
```
mrjoelkemp: Non-blocker: consider leaving a comment above that indicates the implicit assertions fix. Something like Mike's explanation would be helpful to newcomers to the source who are not familiar with the inner assertions.

test/string-checker.js
```diff
@@ -297,29 +339,20 @@ describe('modules/string-checker', function() {
...
-            checker = new Checker();
+            checker = new StringChecker();
             checker.registerDefaultRules();
         });
 
         it('should accept a path to a filter function to filter out errors', function() {
             checker.configure({
                 disallowQuotedKeysInObjects: true,
-                errorFilter: __dirname + '/data/error-filter.js'
+                errorFilter: require(__dirname + '/data/error-filter.js')
```
mrjoelkemp: Why?

mrjoelkemp: This is the only change that blocks this PR for me. Does this require a change in the way users use this feature?

mdevils: This is `StringChecker` test and for some reason `StringChecker` was not actually tested here. `Checker` was. I have changed `new Checker(...)` to `new StringChecker(...)` and it required to change some tests for `StringChecker` compatibility.

> Does this require a change in the way users use this feature?

Nope.

lib/string-checker.js
mdevils: I agree.

test/string-checker.js
mrjoelkemp: :+1:

mrjoelkemp:
LGTM. Great work Marat! This is really exciting.

mdevils:
Thank you guys! I will make fixes you have suggested and will merge this soon.

lib/string-checker.js
gero3: 5 is just about right I think. I already had runs that took 3 times to fix it and that took 3 different rules in my autofixer. But mine doesn't fix the string at the same time of checking it, so this will need a lot of less runs.

lib/string-checker.js
mdevils: I agree.

lib/checker.js
mdevils: I agree.

lib/string-checker.js
mikesherov: @gero3, thanks for the insight. And thanks for the initial nudge. Even though we didn't end up going the "format function" route, your initial PR was invaluable. Thanks again!

lib/checker.js
```diff
@@ -35,13 +35,32 @@ Checker.prototype.configure = function(/*config*/) {
...
+ */
+Checker.prototype.fixFile = function(path) {
+    if (this._configuration.isFileExcluded(path)) {
+        return Vow.resolve(null);
+    }
+
+    return vowFs.read(path, 'utf8').then(function(data) {
+        var result = this.fixString(data, path);
+        return vowFs.write(path, result.output).then(function() {
+            return result.errors;
```
mikesherov: Ah, @mdevils, here is where it writes files. Sorry I missed it!

PR #1023, opened 2015-02-09
Make rules use iterators and whitespace assertions

TheSavior:
Should complete step one and two of #516.

Branched from #1014 to get the new differentLine assertion. That branch should be merged first (which will automatically remove a bunch of the commits from this PR)

PR #1059, opened 2015-02-14
Added `disallow-multiple-spaces` rule. Fixes #435

### LINKED ISSUE #435: [feature request] disallow-multiple-spaces
It would be great if we could have a rule to disallow multiple spaces, such as:

    var x  = 3;

instead of

    var x = 3;

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2616308-feature-request-disallow-multiple-spaces?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

twolfson:
This PR adds a rule for disallowing multiple spaces between any tokens (e.g. `var` and `x`). I decided to add to `token-assert` for reusability purposes. In this PR:

- Added tests for `disallow-multiple-spaces`
- Added rule for `disallow-multiple-spaces`
- Added `maximumSpaces` option for `TokenAssert.whitespaceBetween`

**Missing:**

- Negation rule for `disallow-multiple-spaces` (e.g. `require-multiple-spaces`)
    - I started building this but realized that it was unlikely to be used and none of the other whitespace `disallow` rules had negations
        - https://github.com/twolfson/node-jscs/compare/496c8911bb1f6d9ccadd98e201558e4b6c344ae4

### DIFF
--- a/lib/rules/disallow-multiple-spaces.js
+++ b/lib/rules/disallow-multiple-spaces.js
@@ -1,0 +1,74 @@
+/**
+ * Disallows multiple spaces between identifiers, keywords, and any other token
+ *
+ * Type: `Boolean`
+ *
+ * Values: `true`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowMultipleSpaces": true
+ * ```
+ *
+ * ##### Valid
+ * ```js
+ * var x = "hello";
+ * function y() {}
+ * ```
+ *
+ * ##### Invalid
+ * ```js
+ * var x  = "hello";
+ * function y() {}
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true ||
+            typeof options === 'object' &&
+            options.allowEOLComments === true,
+            'options option requires true value ' +
+            'or an object with `allowEOLComments` property'
+        );
+
+        this.allowEOLComments = options.allowEOLComments;
+    },
+
+    getOptionName: function() {
+        return 'disallowMultipleSpaces';
+    },
+
+    check: function(file, errors) {
+        // Iterate over all tokens (including comments)
+        var _this = this;
+        file.getTokens().forEach(function(token, index, tokens) {
+            // If there are no trailing tokens, exit early
+            var nextToken = tokens[index + 1];
+            if (!nextToken) {
+                return;
+            }
+
+            // If we are allowing EOL comments and the next token is an EOL comment skip it
+            // We don't need to check the current token since EOL comments must be on separate lines from the next one
+            if (_this.allowEOLComments && nextToken.type === 'Line') {
+                return;
+            }
+
+            // Verify we have at most 1 space between this token and the next (won't fail for different lines)
+            errors.assert.whitespaceBetween({
+                token: token,
+                nextToken: nextToken,
+                atMost: 1,
+            });
+        });
+    }
+
+};
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -40,6 +40,21 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
                 fixed: true
             });
         }
+    } else if (options.hasOwnProperty('atMost')) {
+        var atMost = options.atMost;
+        if (nextToken.loc.start.line === token.loc.end.line &&
+            (nextToken.loc.start.column - token.loc.end.column) > atMost
+        ) {
+            this.emit('error', {
+                message: options.message ||
+                    (atMost === 1 ?
+                        'At most 1 space is allowed between ' + token.value + ' and ' + nextToken.value :
+                        'At most ' + atMost + ' spaces are allowed between ' +
+                            token.value + ' and ' + nextToken.value),
+                line: token.loc.end.line,
+                column: token.loc.end.column
+            });
+        }
     } else {
         if (nextToken.range[0] === token.range[1]) {
             nextToken.whitespaceBefore = ' ';
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -564,6 +564,7 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-keywords'));
     this.registerRule(require('../rules/disallow-multiple-line-breaks'));
     this.registerRule(require('../rules/disallow-multiple-line-strings'));
+    this.registerRule(require('../rules/disallow-multiple-spaces'));
     this.registerRule(require('../rules/validate-line-breaks'));
     this.registerRule(require('../rules/validate-quote-marks'));
     this.registerRule(require('../rules/validate-indentation'));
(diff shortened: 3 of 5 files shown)


twolfson:
Ah, I forgot to test multiple lines. I will add that shortly.

test/rules/disallow-multiple-spaces.js
```diff
@@ -0,0 +1,38 @@
...
+
+    it('should not report no spaces', function() {
+        assert(checker.checkString('var x="oops";').isEmpty());
+        assert(checker.checkString('function x(){}').isEmpty());
+        assert(checker.checkString('1+2').isEmpty());
+    });
+
+    it('should not report multiple spaces in strings/regular expressions', function() {
+        assert(checker.checkString('"hello  world";').isEmpty());
+        assert(checker.checkString('/hello  world/').isEmpty());
```
zxqfox: Can you add some tests with comments? Like:
```
var X =  /** @lends X */  {  /** @type {String} */  name: "some" };
```

twolfson: Sure thing, more tests coming right up.

lib/rules/disallow-multiple-spaces.js
```diff
@@ -0,0 +1,65 @@
...
+            'disallowMultipleSpaces option requires true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'disallowMultipleSpaces';
+    },
+
+    check: function(file, errors) {
+        file.getTokens().forEach(function(token, index, tokens) {
```
zxqfox: I'm not sure but `getTokens` returns token list without comments and tests suggested by me will fail.
https://github.com/jscs-dev/node-jscs/blob/master/lib/js-file.js#L492

I suggest to use here `while (token = file.nextToken({includeComments: true}) { ...` or something similar to iterate through tokens and comments.

zxqfox: @mikesherov @mdevils What you think about integrating this into `JSFile.interateTokens*` methods?

twolfson: Ah, didn't realize we were missing those tokens. Good catch. Do you think there would be a reuse value in an `iterateTokens` with options for including comment tokens?

twolfson:
@zxqfox I have added in the comment tests and fix for iterating with comments. The `file.getNextToken` method requires inputting a seed token. Unfortunately, there is no way to get the first token of a file without accessing `_tokens`. As a result, I I chose to replace our `getTokens` with a direct iterator on `_tokens`.

test/rules/disallow-multiple-spaces.js
```diff
@@ -0,0 +1,56 @@
...
+        assert(checker.checkString('var x  = "oops";').getErrorCount() === 1);
+        assert(checker.checkString('function x  () {}').getErrorCount() === 1);
+        assert(checker.checkString('function x()  {}').getErrorCount() === 1);
+        assert(checker.checkString('1  + 2').getErrorCount() === 1);
+        assert(checker.checkString('1 +  2').getErrorCount() === 1);
+    });
+
+    it('should report multiple spaces between comments', function() {
+        assert(checker.checkString('var X = {  /** @type {String} */ name: "some" };').getErrorCount() === 1);
+        assert(checker.checkString('var x = "oops";  // Multiple spaces before comment').getErrorCount() === 1);
```
zxqfox: I think this case should be an exception. Maybe we can make an option?
```
var CONST1 = "one"; // makes me sad
var C2 = "1";       // makes me happy
```

lib/token-assert.js
```diff
@@ -38,6 +38,21 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
                 column: token.loc.end.column
             });
         }
+    } else if (options.hasOwnProperty('maximumSpaces')) {
```
zxqfox: This is very close with https://github.com/jscs-dev/node-jscs/pull/1023/files#diff-526f88acb4a66447fc67eaf20ca6acadR104

We need to decide what naming we should use.
/cc @markelog @mdevils @mikesherov

zxqfox:
@twolfson Thanks for contribution! Great work here. Here we have a conflict with another PR so I suggest to wait someone else to discuss with.
Personally, I like atLeast/atMost (just because it's shorter). ;-)

lib/token-assert.js
twolfson: We should probably clarify that those are about lines. It will be confusing to have them and `maximumSpaces`. We can rename to `atMostSpaces` for consistency.

twolfson: Ah, nvm. Fell into the same trap. That is a different function entirely. I will still rename to `atLeastSpaces` for semi-consistency but whitespace isn't necessarily spaces only.

zxqfox: `atMost` option is quietly fine if method will be about spaces (if we decide to use `atMost` for lines). I just wanted some consistency in options naming.

test/rules/disallow-multiple-spaces.js
twolfson: Yep, definitely. I will add a new option:

```js
disallowMultipleSpace: {
  allowEOLComments: true
}
```

lib/token-assert.js
twolfson: There is a possible trap if we ever start to use `tabs` inside of `whitespaceBetween` but I will move to `atMost` for now. I am going to add a comment on the other PR about renaming `exactly` to `lines` for consistency with `whitespaceBetween`.

zxqfox: Good idea ;-)

twolfson:
I have added the case for EOL comments and renamed to `atMost`. After the rename, I did a `git grep` and saw that we are using `maximumLineLength` as a rule and not using `atMost` anywhere else. That might have a factor on naming.

mrjoelkemp:
@mikesherov cool with this?

@twolfson Mind rebasing?

mikesherov:
I'm not really a fan of this rule because I think it's solved by adding `atLeast`, `atMost`, and `exactly` to the spacing assertions and passing those down through the rule config themselves. This is a fairly decent stop gap measure but I predict will be rife with requests for people asking for leniency to this rule in certain cases, which highlights the problem of it being too overarching.

vkramskikh:
I'd really love to see this in master. Double spaces is the most frequent error in my project which is still not covered by JSCS.

twolfson:
@mikesherov I am not sure I follow. Do you mean adding `atLeast`/`atMost`/`exactly` support to every `requireSpace*` rule?

twolfson:
@mrjoelkemp I will take a shot at rebasing by the end of this weekend. It looks like there are some merge conflicts as well =/

mikesherov:
I'm fine merging it in, I was just saying that the customizable route will be not to address it by making this rule have tons of options, but rather by enabling more fine grained control of the other options.

twolfson:
I have merged/squashed on top of the latest `master`. Notable changes:

- Relocated test files/changes for consistency
- Renamed `configure` parameter to `options` for consistency
- Moved to `file.getTokens` over `file._tokens` to stop relying on internals

mrjoelkemp:
Nice work here @twolfson. Always happy to see a super old issue closed. Thanks for contributing.

PR #1061, opened 2015-02-14
requireSpaceAfterKeywords: use token assert

hzoo:
I modified the tests since we don't need multiple error messages anymore? This rule enforces 1 space.
I wasn't sure about the `nextToken.loc.end.line === token.loc.start.line` check - we didn't have any tests for that either.

### DIFF
--- a/lib/rules/require-space-after-keywords.js
+++ b/lib/rules/require-space-after-keywords.js
@@ -42,11 +42,6 @@
  */
 
 var assert = require('assert');
-var util = require('util');
-var texts = [
-    'Missing space after "%s" keyword',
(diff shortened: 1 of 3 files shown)


hzoo:
Looks like an npm install issue with `esvalidate.cmd`?

TheSavior:
I believe this is a duplicate of part of #1023

hzoo:
Oh okay there's a lot of rules there! - but I don't see require-space-after-keywords in the files changed?

hzoo:
Hmm this is failing 2 tests: `modules/checker fixing fixFile should fix specified file:` @mdevils?

PR #1063, opened 2015-02-14
Disallowing do..while on new line

### LINKED ISSUE #885: Disallowing do..while on new line

oredi:
Make option disallowKeywordsOnNewLine perform special scenarios for the keyword "while" when checking "DoWhileStatements". The behavior contradicts the original idea of keyword "while" but it is done to reuse existing code in a efficient and less complex manner
    1.  Allow "while" that is part of a "do while" with no braces to succeed
    2.  Allow "while" that is not part of a "do while" to succeed even though it contradicts rule meaning

Fixes #885 - issue

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -36,7 +36,10 @@ var JsFile = function(filename, source, tree, options) {
     this._tokenRangeEndIndex = tokenIndexes.tokenRangeEndIndex;
     this._tokensByLineIndex = tokenIndexes.tokensByLineIndex;
 
-    this._index = this._buildNodeIndex();
+    var nodeIndexes = this._buildNodeIndex();
+    this._index = nodeIndexes.nodesByType;
+    this._nodesByStartRange = nodeIndexes.nodesByStartRange;
+
     this._fixEsprimaIdentifiers();
 
     this._buildDisabledRuleIndex();
@@ -371,6 +374,23 @@ JsFile.prototype = {
                 return false;
             }
         });
+        return result;
+    },
+
+    /**
+     * Returns nodes by range start index from earlier built index.
+     *
+     * @param {Object} token
+     * @returns {Object[]}
+     */
+    getNodesByFirstToken: function(token) {
+        var result = [];
+        if (token && token.range && token.range[0] >= 0) {
+            var nodes = this._nodesByStartRange[token.range[0]];
+            if (nodes) {
+                result = result.concat(nodes);
+            }
+        }
         return result;
     },
 
@@ -720,21 +740,30 @@ JsFile.prototype = {
     },
 
     /**
-     * Builds node index using node type as the key.
-     *
-     * @returns {Object}
+     * Builds node indexes using
+     *  i. node type as the key
+     *  ii. node start range as the key
+     * @returns {{nodesByType: {}, nodesByStartRange: {}}}
      * @private
      */
     _buildNodeIndex: function() {
-        var index = {};
+        var nodesByType = {};
+        var nodesByStartRange = {};
         this.iterate(function(node, parentNode, parentCollection) {
             var type = node.type;
 
             node.parentNode = parentNode;
             node.parentCollection = parentCollection;
-            (index[type] || (index[type] = [])).push(node);
+            (nodesByType[type] || (nodesByType[type] = [])).push(node);
+
+            // this part builds a node index that uses node start ranges as the key
+            var startRange = node.range[0];
+            (nodesByStartRange[startRange] || (nodesByStartRange[startRange] = [])).push(node);
         });
-        return index;
+        return {
+            nodesByType: nodesByType,
+            nodesByStartRange: nodesByStartRange
+        };
     },
 
     /**
--- a/lib/rules/disallow-keywords-on-new-line.js
+++ b/lib/rules/disallow-keywords-on-new-line.js
@@ -58,6 +58,24 @@ module.exports.prototype = {
                 return;
             }
 
+            // Special cases for #885, using while as the keyword contradicts rule meaning
+            // but it is more efficient and reduces complexity of the code in this rule
+            if (token.value === 'while') {
+                var nodes = file.getNodesByFirstToken(token);
+
+                if (nodes.length === 0) {
+                    // "while" that is part of a do will not return nodes as it is not a start token
+                    if (prevToken.value !== '}') {
+                        // allow "while" that is part of a "do while" with no braces to succeed
+                        return;
+                    }
+                } else {
+                    // it is a "while" statement that is not part of a "do while"
+                    // , allow it to succeed even though it contradicts rule meaning
+                    return;
+                }
+            }
+
             errors.assert.sameLine({
                 token: prevToken,
                 nextToken: token
--- a/test/specs/js-file.js
+++ b/test/specs/js-file.js
@@ -589,6 +589,68 @@ describe('modules/js-file', function() {
         });
     });
 
+    describe('getNodesByFirstToken', function() {
+        describe('invalid arguments', function() {
+            var file;
+            beforeEach(function() {
+                file = createJsFile('x++;y++;');
+            });
+
+            it('should return empty array if the argument is invalid', function() {
+                var nodes = file.getNodesByFirstToken();
+                assert.equal(nodes.length, 0);
+            });
+
+            it('should return empty array if the argument.range is invalid', function() {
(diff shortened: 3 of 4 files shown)


sdepold:
Will this change also change the behavior of `requirePaddingNewlinesBeforeKeywords: ["while"]` when using a do-while loop?

oredi:
This will not have any impact on requirePaddingNewlinesBeforeKeywords: ["while"], this is the nice thing I like about node-jscs. At least from my initial look into sections of the code all the rules are create using SOLID principles. So changes in 1 rule will not impact other rules.

The change occurred on the disallowKeywordsOnNewLine rule so it does not impact other rules. Saying this I took a look requirePaddingNewlinesBeforeKeywords I do think there is something incorrect when you have do-while loop. It does not handle the situation while you have a } before while. So the following code will fail for requirePaddingNewlinesBeforeKeywords 

```javascript
var i = 0;
do {
        i++;

} while (i < 10);
```

I hope I am answering you correctly. If there is a particular piece of code you use to run against this PR. Do let me know I can try it out for you.

sdepold:
So I had this code:

```
var conflict, something;

do {
  something = calculateSomethingRandom();
  conflict = isThereAConflict(something);
} while (conflict)
```

And it would tell me that there is not newline before `while` (which is true). Ideally it would be possible to treat do while in a special way

oredi:
I think this will be a special case for requirePaddingNewlinesBeforeKeywords. To keep this thread clean you might want to create a new issue for it. Then we will see if the core developers agree to implement this special case. I can take up the job since I think it will be similar to what I did for disallowKeywordsOnNewLine. But I did to see if what I have done here for disallowKeywordsOnNewLinefits into the direction of the core developers.

sdepold:
Done https://github.com/jscs-dev/node-jscs/issues/1092

oredi:
Thanks let's wait for the feedback of the others on  this.

mikesherov:
Hi @sdepold and @oredi, sorry for taking a while to get back to you guys. I was heads down in "autofixing" code. This looks pretty good, but using getNodeByRange is a bit unsafe, and we should probably deprecate it. What would be better is if you could create a `getNodesByFirstToken` which returned all nodes that start at that index, considering there could be many. It would take in a token, and not just a range. You would then array search instead of strict compare the result.

Let me know if you can tackle that, and we can discuss fixes this in the generic sense for the different rules that exist.

oredi:
Hi @mikesherov,

I will take a further look at this in my morning (JST) and let you know.

oredi:
Hi @mikesherov, 

I was exploring the js-file object today and I notice there is a loc.start property on both the node and token objects. I am thinking that these properties could be the key building the array for the search that you are mentioning.  

As I am pretty new to this I am not sure if I am hitting in the correct direction of what you require for getNodesByFirstToken. I will continue exploring the code to try to figure what is required here but if you do have someone who can implement this logic faster. I will be willing to volunteer looking into the existing rules referencing an example implementation if you need help working on it.

oredi:
Hi @mikesherov,

Sorry for the repeated comments, I got an idea on how to write getNodesByFirstToken but I wish to check with you about my thoughts to see if what I am doing fits your idea. As I feel there are a lot of changes to js-file object although not in a breaking way. I would like to see if the changes listed below fits into what you envision for getNodesByFirstToken. Thanks

Changes
1. _buildNodeIndex will be modified to be similar to _buildTokenIndex, instead of _buildNodeIndex  returning just a class with the current index by node type will be modified from 
```javascript
_buildNodeIndex: function() {
  var index = {};
  this.iterate(function(node, parentNode, parentCollection) {
    var type = node.type;

    node.parentNode = parentNode;
    node.parentCollection = parentCollection;
    (index[type] || (index[type] = [])).push(node);
  });
  return index;
}
```
to
```javascript
_buildNodeIndex: function() {
  var nodesByType = {};
  var nodesByStartRange = {};
  this.iterate(function(node, parentNode, parentCollection) {
    var type = node.type;

    node.parentNode = parentNode;
    node.parentCollection = parentCollection;
    (nodesByType [type] || (nodesByType [type] = [])).push(node);
    
    var startRange = node.range[0];
    (nodesByStartRange[startRange ] || nodesByStartRange[startRange] = [).push(node);
  });

  return {
    nodesByType: nodesByType,
    nodesByStartRange: nodesByStartRange
  };
}
```

2. line 39 of js-file.js will be modified from 
```javascript
this._index = this._buildNodeIndex();
```
to 
```javascript
var nodeIndexes = this._buildNodeIndex();
this._index = nodeIndexes.nodesByType;
this._nodesByStartRange= nodeIndexes.nodesByStartRange;
```

3. and the new function getNodesByFirstToken will be as follows

```javascript
getNodesByFirstToken : function(token) { 
  var result = [];
  if(token.range && token.range[0]) {
    var nodes = this._nodesByStartRange[token.range[0]];
    if(nodes) {
      result = result.concat(nodes);
    }
  }
  return result;
}
```

mikesherov:
cc @mdevils in regards to proposed implementation above?

mikesherov:
@oredi, seems like what I had in mind. Good work! I just want confirm from mdevils too though.

oredi:
@mikesherov Thanks cc @mdevils 
I have modified my PR to include this code and also change the implementation of disallowKeywordsOnNewLine to use it. The changes have been squash into a single commit after merging the latest from jscs-dev:master.

sdepold:
It broke the build :(

oredi:
Might be related to #1141. I think all the recent PRs will failing

zxqfox:
Nvm about appveyor, it uses iojs1.4.x (reason of broken build), we need to merge that PR first.

oredi:
@zxqfox Thanks. I see that PR merged and rebased this PR with the latest Master and the build is working now.

mrjoelkemp:
@mdevils ping https://github.com/jscs-dev/node-jscs/pull/1063#issuecomment-77146304.

oredi:
@mikesherov @mdevils 
I have rebase my PR based on the latest master. While working on the conflict I realized that the spec is starting to get a lot of repeated code. This is just a suggestion, but if this PR is suitable for merging. I could look into re-factoring the spec to make it DRY. I am not doing that now since I am introducing what I think is a fair amount of changes here. So I wanted to keep the tests as original as possible

mikesherov:
ping @mdevils

markelog:
@mikesherov 

>  but using getNodeByRange is a bit unsafe

Unsafe how?

markelog:
Will ping @mdevils too :-)

oredi:
@markelog just for your information reliance on getNodeByRange has been removed and a new method getNodesByFirstToken has been introduced in jsFile. The rule was also modified to use the new method. Thanks

mikesherov:
@markelog as discussed previously, `getNodeByRange` assumes that there is only one node that matches, which is incorrect.

markelog:
> @markelog as discussed previously, getNodeByRange assumes that there is only one node that matches, which is incorrect.

 `getNodeByRange` finds deepest node with it provided range number, so "algoritm" is pretty deterministic

mdevils:
Hello, guys! Sorry for a delay.

I think having this `getNodesByFirstToken` is fine. But what I'm not sure is adding these exceptions. I'd split while to two possible operator names `while` and `do-while` or something like that. This is more clear and strict. How do you think?

mdevils:
Don't listen to me :smile:

In this case I think this exception is reasonable. Who would ever disallow `while` which is not part of `do..while` to be on a new line...)

mdevils:
@markelog, @mikesherov, how do you think?

mikesherov:
I like this PR as is now. Up to @markelog to agree :)

Issue #1065, opened 2015-02-14
Add idiomatic preset [$15]


markelog:
It seems it's one of popular javascript "code styles" out there. 
It could be easily extended from jquery preset.

https://github.com/rwaldron/idiomatic.js

mdevils:
+1 to this preset

hzoo:
I'l give it a shot - although it doesn't seem to have a preference over some things like spaces/tabs - just that it should be consistent.

ehartford:
Please pull this!  Needed!

serg-io:
+1

markelog:
Since you dealt with the bounty ticket last time :-)

mrjoelkemp:
Henry can do it. I'm out of the country at the moment. Cc @hzoo

Issue #99, opened 2013-12-18
Add option to enforce variable declaration at the top of declaring scope


Druotic:
In short, the option would require that all variable declarations be made at the top of their declaring scope -- where a JS interpreter would "hoist" them.  

Previously, I had proposed this idea and implemented it for JSHint, but JSHint is moving away from styling entirely in V3.0, so it was not accepted.  Unlike JSHint, JSCS seems like the perfect place for this sort of feature!

Rather than duplicating the detailed explanation/examples, see the issue jshint/jshint#1263 and pull request jshint/jshint#1283 for additional info and previous discussions.

This will be my first contribution to jscs, so it might take me a bit to port it over.  Hopefully it won't be very different/difficult.

markelog:
This issue has been open more than a month without supported PR for it. 

Unfortunately, we don't have the resources to implement rules that could not be included in any of the supported [presets](https://github.com/jscs-dev/node-jscs/tree/master/presets) (if it could be, please point that out).

However, if you or anybody else submit PR for it, we will gladly consider it.

Krinkle:
@markelog This would be a rule to enable in the wikimedia preset. We currently enforce this rule by convention only. Would be great to have it be enforced by JSCS.

ava-dylang:
:+1: To quote my colleges "How is this not supported yet". I guess people who write multiple var declarations don't overlap that much with people who like to use linters, or write linters anyway. Anyways glad I found *something, anything* on this topic.

oredi:
Any preference for the rule name, I will try to take a look into this.
If anyone has any idea about an existing rule that can act as a start it will be a time saver.
Or else I would dig into the existing rules for implementation.

oredi:
Think I found the requireMultipleVarDecl that might be applicable for this. I am thinking about names like requireVarDeclAtTop with the only option value of true unless someone else has better suggestions.

mrjoelkemp:
requireVarDeclFirst sounds closer to the rule's logic.

markelog:
Check this out - https://github.com/jscs-dev/node-jscs/issues/807

oredi:
@mrjoelkemp  and @markelog  Thanks I will use that and and look into that sample implementation.

Issue #836, opened 2014-12-15
Ignore/Except Quotes in requireSpacesInsideParentheses


kelvinlouis:
Is it possible to ignore quotes from requireSpacesInsideParentheses?
I'd like to cover this case (https://github.com/rwaldron/idiomatic.js/#type 2.C.1.3) 
```javascript
// Single argument string literal, no space
foo("bar");
```

This doesn't work:
```javascript
"requireSpacesInsideParentheses": {
 "all": true,
 "except": [ "{", "}", "\"" ]
},
```

Any ideas?

markelog:
Reopening, since jquery code-style changed and doesn't need it anymore, but <a href="https://github.com/jscs-dev/node-jscs/issues/1065">incoming</a> `idiomatic` preset would require this feature.

/cc @hzoo

hzoo:
How would we go about checking this?
The rule is using `value = nextToken.value`.
```js
foo("bar"); // value = "bar"
foo('bar'); // value = 'bar'
```
quote is single quote, double quote, es6 template
and then something like `if (nextToken.type === 'Literal' && value[0] === quote) { return; }`

PR #1069, opened 2015-02-16
Preset: Idiomatic

hzoo:
#1065. WIP

Idiomatic | JSCS
-------------|---------
`indentation` | no preference but be consistent!
`quotes?` | use jquery preset: `"validateQuoteMarks": { "mark": "\"", "escape": true }`,
`foo("bar")`| **new rule option**: `requireSpacesInsideParentheses`: Single argument string literal, no space
`if ( !("foo" in obj) ) {` | **new rule option**: `requireSpacesInsideParentheses`: Inner grouping parens, no space
`foo([ "alpha", "beta" ]);` | `"requireSpacesInsideParentheses": { "except": [ "[", "]" ] }`
`(function( global ) {})()` | `"requireSpacesInsideParentheses": { "except": ["function"] }`
`foo({ a: "alpha" });` | `"requireSpacesInsideParentheses": { "except": [ "{", "}" ] }`
`square( 10 );` | `"disallowSpacesInCallExpression": true`
`// statements` | `"requireSpaceAfterLineComment": { "allExcept": ["#", "="] }`
 | `"disallowMultipleLineBreaks": null`
`var, const, let statements should always be in the beginning of their respective scope` | #99 **new rule** `requireVarDeclFirst`
`name of regex: rDesc` | **new rule**?
`End of line comments are prohibited!` | **new rule** `validateCommentPosition`
`var factorial = function factorial( number ) {` | similar to #850
`avoid switch if possible` | `disallowNodeTypes: ["SwitchStatement"]`
`early returns` | new rule https://github.com/jscs-dev/node-jscs/issues/1223 `requireEarlyReturn`

### DIFF
--- a/presets/idiomatic.json
+++ b/presets/idiomatic.json
@@ -1,0 +1,17 @@
+{
+    "preset": "jquery",
+    "requireSpacesInsideParentheses": {
+        "all": true,
+        "ignoreParenthesizedExpression": true,
+        "except": [ "{", "}", "[", "]", "function", "\"" ]
+    },
+    "disallowSpacesInCallExpression": true,
+    "requirePaddingNewLinesBeforeLineComments": null,
+    "requireSpaceAfterLineComment": {
+        "allExcept": ["#", "="]
+    },
+    "disallowMultipleLineBreaks": null,
+    "requireCapitalizedComments": null,
+    "validateIndentation": null,
+    "validateCommentPosition": { "position": "above" }
+}
--- a/test/data/options/preset/idiomatic.js
+++ b/test/data/options/preset/idiomatic.js
@@ -1,0 +1,223 @@
+// A. Parens, Braces, Linebreaks
+
+
+// if/else/for/while/try always have spaces, braces and span multiple lines
+// this encourages readability
+
+// 2.A.1.1
+// Use whitespace to promote readability
+
+if ( condition ) {
+
+}
+
+while ( condition ) {
+
+}
+
+for ( ; i < length; i++ ) {
+
+}
+
+for ( prop in object ) {
+
+}
+
+if ( true ) {
+
+} else {
+
+}
+
+// B. Assignments, Declarations, Functions ( Named, Expression, Constructor )
+
+// 2.B.1.2
+// Using only one `var` per scope (function) promotes readability
+// and keeps your declaration list free of clutter (also saves a few keystrokes)
+
+// or..
+var
+foo = "",
+bar = "",
+quux;
+
+// 2.B.1.3
+// var statements should always be in the beginning of their respective scope (function).
+
+// Good
+function foo() {
+  var bar = "",
+    qux;
+
+  // all statements after the variables declarations.
+}
+
+// 2.B.1.4
+// const and let, from ECMAScript 6, should likewise be at the top of their scope (block).
+
+// Good
+function foo() {
+  let foo;
+  if ( condition ) {
+    let bar = "";
+
+  }
+}
+
+// 2.B.2.1
+// Named Function Declaration
+function foo( arg1, argN ) {
+
+}
+
+// Usage
+foo( arg1, argN );
+
+// 2.B.2.2
+// Named Function Declaration
+function square( number ) {
+  return number * number;
+}
+
+// Usage
+square( 10 );
+
+// Really contrived continuation passing style
+function square( number, callback ) {
+  callback( number * number );
+}
+
+square( 10, function( square ) {
+  // callback statements
+});
+
+// 2.B.2.3
+// Function Expression
+function a() {
+    var square = function( number ) {
+      // Return something valuable and relevant
+      return number * number;
+    };
+}
+
+// Function Expression with Identifier
+// This preferred form has the added value of being
+// able to call itself and have an identity in stack traces:
+function b() {
+    var factorial = function factorial( number ) {
+      if ( number < 2 ) {
+        return 1;
+      }
+
+      return number * factorial( number - 1 );
+    };
+}
+
+// 2.B.2.4
+// Constructor Declaration
+function FooBar( options ) {
+  this.options = options;
+}
+
+// Usage
+function c() {
+    var fooBar = new FooBar({ a: "alpha" });
+    fooBar.options;
+    // { a: "alpha" }
+}
+
+// 2.C.1.1
+// Functions with callbacks
+foo(function() {
+  // Note there is no extra space between the first paren
+  // of the executing function call and the word "function"
+});
+
+// Function accepting an array, no space
+foo([ "alpha", "beta" ]);
+
+// 2.C.1.2
+// Function accepting an object, no space
+foo({
+  a: "alpha",
+  b: "beta"
+});
+
+// ** new rule **
+// Single argument string literal, no space
+foo("bar");
+
+// ** new rule **
+// Inner grouping parens, no space
+if ( !("foo" in obj) ) {
+
+}
+
+// 5.1.1
+// A Practical Module
+
+(function( global ) {
+  var Module = (function() {
+
+    var data = "secret";
+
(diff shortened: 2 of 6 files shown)


mdevils:
+ @markelog

presets/idiomatic.json
```diff
@@ -0,0 +1,7 @@
+{
+    "preset": "jquery",
+    "requireSpacesInsideParentheses": {
+        "all": true,
+        "except": [ "{", "}" ]
```
markelog: Correct one is 
```js
"except": [ "{", "}", "[", "]", "function" ]
```

And as a said feature, `requireSpacesInsideParentheses` should understand quotes, so case like this:
```js
foo("bar")
```
was acceptable.

test/data/options/preset/idiomatic.js
```diff
@@ -0,0 +1,221 @@
...
+
+// 2.C.1.1
+// Functions with callbacks
+foo(function() {
+  // Note there is no extra space between the first paren
+  // of the executing function call and the word "function"
+});
+
+// Function accepting an array, no space
+foo([ "alpha", "beta" ]);
```
markelog: If `requireSpacesInsideParentheses` doesn't throw on this, then it's a bug

test/data/options/preset/idiomatic.js
```diff
@@ -0,0 +1,221 @@
...
+
+// 2.C.1.2
+// Function accepting an object, no space
+foo({
```
markelog: same as above

test/data/options/preset/idiomatic.js
```diff
@@ -0,0 +1,221 @@
...
+  a: "alpha",
+  b: "beta"
+});
+
+// Single argument string literal, no space
+foo("bar");
+
+// Inner grouping parens, no space
```
markelog: Catch cases like this is currently impossible and i think, really hard to implement

test/data/options/preset/idiomatic.js
```diff
@@ -0,0 +1,221 @@
...
+if ( !("foo" in obj) ) {
+
+}
+
+// 5.1.1
+// A Practical Module
+
+(function( global ) {
+  var Module = (function() {
```
markelog: same as above

markelog:
@rwaldron you might like this :-)

test/data/options/preset/idiomatic.js
markelog: Maybe, i overestimated it, maybe not that hard :-), but still need to add additional feature for that rule

test/data/options/preset/idiomatic.js
hzoo: Ok it's good - I was testing on the wrong config or something at the time. Updated the OP

hzoo:
Oh failing because of `requirePaddingNewLinesBeforeLineComments` in jQuery

markelog:
We can't land this, until https://github.com/jscs-dev/node-jscs/pull/1069#discussion_r24805322 is all done, specifically until

```js
foo("bar")
```

 case is handled, would like to take a hit on this one? https://github.com/jscs-dev/node-jscs/issues/836

hzoo:
Yeah I was just updating the PRs since I forgot to update 2 files

hzoo:
Yeah I'l take a look

mrjoelkemp:
What's the status on this PR? Still held up?

hzoo:
Yeah on #836. Need an exception for `foo("bar")` - `requireSpacesInsideParentheses` but except string literals so I guess for
```js
", ', `
```

hzoo:
I guess I can finish this now lol

mrjoelkemp:
@hzoo Can we not just land this with that case in an open issue? Would it be reporting the wrong case or nothing at all? Can we just nullify that case?

Would love to land this finally.

hzoo:
Yeah we could land it now if we want (and make an issue for the other rules that need to be created) @markelog

ehartford:
Please pull?

markelog:
@ehartford 

> still need foo("bar") case

ehartford:
can not make a separate issue for that case?

markelog:
https://github.com/jscs-dev/node-jscs/issues/836

hzoo:
@markelog Do we want to release the new preset now and update it later when #836 is done?

markelog:
Well, we certainly can, but that feature is kind important

presets/idiomatic.json
```diff
@@ -0,0 +1,17 @@
+{
+    "preset": "jquery",
+    "requireSpacesInsideParentheses": {
+        "all": true,
+        "except": [ "{", "}", "[", "]", "function" ]
+    },
+    "requireSpacesInsideParenthesizedExpression": {
```
markelog: Why we need this?

markelog:
Can we include `requireVarDeclFirst`?

hzoo:
Ok I can add `requireVarDeclFirst`.

`JSDoc style is good, but requires a significant time investment` - don't think we need to make a rule to prevent jsdoc style unless the rule would be `disallowBlockComments`.

`End of line comments are prohibited!`

That would be https://github.com/jscs-dev/node-jscs/issues/651 - `validateCommentPosition: "above"` (disallow beside)

hzoo:
Getting an error with `requireVarDeclFirst` on 

```js
// Good
function foo() {
  let foo;
  if ( condition ) {
    let bar = "";

  }
}
```

presets/idiomatic.json
hzoo: Thought it was for

```
if ( !("foo" in obj) ) {

}
``` but I guess not

**Comment by markelog** [comment]
> JSDoc style is good, but requires a significant time investment - don't think we need to make a rule to prevent jsdoc style unless the rule would be disallowBlockComments.

I was thinking we should include jsdoc rules 

> Getting an error with requireVarDeclFirst on

https://github.com/jscs-dev/node-jscs/issues/1783

**Comment by hzoo** [comment]
Ok updating with

```js
   "requireSpacesInsideParentheses": {
        "all": true,
        "ignoreParenthesizedExpression": true,
        "except": [ "\"" ]
    },
```

markelog:
How about `disallowSpacesInsideParenthesizedExpression: true` and `validateCommentPosition:`?

hzoo:
Getting

```js
Illegal space after opening grouping parenthesis at 1.js :
   208 |
   209 |  Ctor.prototype.setFoo = function( val ) {
   210 |    return ( this.foo = val );
--------------------^
   211 |  };
   212 |

Illegal space before closing grouping parenthesis at 1.js :
   208 |
   209 |  Ctor.prototype.setFoo = function( val ) {
   210 |    return ( this.foo = val );
-----------------------------------^
   211 |  };
   212 |
```

for disallowSpacesInsideParenthesizedExpression - adding validatecomment

markelog:
:-(

markelog:
It seems we all good here?

hzoo:
Yeah only `requireVarDeclFirst` is left? (and we have an issue for that already)

hzoo:
It was kind of annoying to depend on a preset (jquery) though - since if you update the jquery preset, then the tests will fail (validateIdentation errors in that case) so I just removed that?

markelog:
> since if you update the jquery preset, then the tests will fail (validateIdentation errors in that case) so I just removed that?

Do what you think is right :-).

markelog:
How we doing here?

hzoo:
Can merge - there's only some rules we haven't implemented yet.

markelog:
I guess we did catch the spirit of this preset, we can take care the details afterwards

hzoo:
So :+1: for this?

markelog:
LGTM

hzoo:
Ok finally merged! :tada:

PR #1075, opened 2015-02-17
Autofixing: add more rules to use assertion framework

hzoo:
Refs #516
Closes gh-1075

I also removed some variable declarations and put it in the errors.assert object if we want that.


mikesherov:
Thanks as always!

PR #1082, opened 2015-02-19
Make rules use comment iterators

TheSavior:
This is necessary for the changes we want in #1028

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -499,7 +499,7 @@ JsFile.prototype = {
      * @returns {Object[]}
      */
     getTokens: function() {
-        return this._tree.tokens;
+        return this._tree.tokens.concat(this.getComments());
     },
 
     /**
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -499,7 +499,7 @@ JsFile.prototype = {
      * @returns {Object[]}
      */
     getTokens: function() {
-        return this._tree.tokens.concat(this.getComments());
+        return this._tokens;
     },
 
     /**
--- a/lib/rules/disallow-capitalized-comments.js
+++ b/lib/rules/disallow-capitalized-comments.js
@@ -57,7 +57,7 @@ module.exports.prototype = {
         var letterPattern = require('../../patterns/L');
         var lowerCasePattern = require('../../patterns/Ll');
 
-        file.getComments().forEach(function(comment) {
+        file.iterateTokensByType(['Line', 'Block'], function(comment) {
             var stripped = comment.value.replace(/[\n\s\*]/g, '');
             var firstChar = stripped[0];
 
--- a/lib/rules/disallow-keywords-in-comments.js
+++ b/lib/rules/disallow-keywords-in-comments.js
@@ -90,7 +90,7 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-        file.getComments().forEach(function(comment) {
+        file.iterateTokensByType(['Line', 'Block'], function(comment) {
             getCommentErrors(comment, this._keywordRegEx).forEach(function(errorObj) {
                 errors.add(this._message, errorObj);
             }.bind(this));
--- a/lib/rules/disallow-space-after-line-comment.js
+++ b/lib/rules/disallow-space-after-line-comment.js
@@ -43,14 +43,10 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-        var comments = file.getComments();
-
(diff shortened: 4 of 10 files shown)


lib/rules/disallow-capitalized-comments.js
```diff
@@ -57,7 +57,7 @@ module.exports.prototype = {
         var letterPattern = require('../../patterns/L');
         var lowerCasePattern = require('../../patterns/Ll');
 
-        file.getComments().forEach(function(comment) {
+        file.iterateTokensByType(['Line', 'Block'], function(comment) {
```
TheSavior: It definitely feels weird to iterate by the types Line and Block and expect only comments. This should be renamed to LineComments and BlockComments but it appears that the code that is doing that is deeper than our project. At least, I couldn't find where those were being set in JSCS.

TheSavior: I think that a rename change to how these tokens are typed should be a separate standalone PR.

TheSavior:
This actually has a bit more nuance to it.

@mdevils In js-file lines ~28 now has this._tree.tokens and this._tokens
```javascript
this._tree.tokens = this._fixEs6Tokens(this._tree.tokens);
this._tokens = this._buildTokenList(this._tree.tokens, this._tree.comments);
```

`getTokens` was returning only this._tree.tokens, but should probably now return this._tokens. Since we are now adding the EOF token as well, that means the tests for js-file need to be updated if we make this switch. It seems like the right switch to be making. Is there something I'm missing, or should I go ahead with this?

mikesherov:
we should now be using `this._tokens` everywhere, right @mdevils? That was my intent at least... and yeah, `getComments` should be deprecated.

lib/js-file.js
```diff
@@ -499,7 +499,7 @@ JsFile.prototype = {
      * @returns {Object[]}
      */
     getTokens: function() {
-        return this._tree.tokens;
```
mikesherov: no, the better thing here is `this._tokens`.

mikesherov:
Also, can you rebase this please?

lib/js-file.js
zxqfox: Yeah. `_tree.tokens.concat(_tree.comments)` ordered differently than `_tokens`.

mdevils:
@mikesherov Yes, looks like `getTokens` should return combined token list (with comments).

lib/rules/disallow-capitalized-comments.js
mdevils: Yeah, I thought so too... `LineComment` and `BlockComment` has clearer meaning.

zxqfox: :+1:

TheSavior:
@mdevils. Combined being `this._tokens` or `this._tree.tokens.concat(this._tree.comments)`? It seems like it should be `this._tokens`

zxqfox:
@TheSavior Difference just in order. `this._tokens` are better since the set is the same.

mikesherov:
@TheSavior, yes, just `this._tokens`

TheSavior:
There is actually one other difference, which is what breaks tests currently. `this._tokens` has EOF:

`this._tokens:`
```
['var', 'a', '// com1', 'var', 'b', '// com2', EOF]
```

`this._tree.tokens.concat(this._tree.comments):`

```
['var', 'a', 'var', 'b', '// com1', '// com2']
```

lib/rules/disallow-capitalized-comments.js
TheSavior: How do we actually do this? I don't see where the types for those tokens are being defined.

zxqfox: As I see it comes from esprima code base so the only thing we can do is redefining.

TheSavior: It's weird, but lets just be consistent with esprima.

TheSavior:
Alright, I think this is good for a final review.

lib/rules/disallow-capitalized-comments.js
zxqfox: Actually, we don't need to. It's not a part of Parser API because comments doesn't affect code execution and they are just skipped on parsing for execution. Since we merging comment tokens into code ASD we already using superset of Parser API and `Line` and `Block` types are confusing. I suggest to fix that and to describe it in JSFile docs.

mikesherov:
Thanks!

Issue #1084, opened 2015-02-19
Consider Errors and jsFile modules public interfaces


markelog:
Since it used in rules, which could be third party rules, we need to document and correctly semver them.

zxqfox:
Moreover these "modules" widely used in plugins. ;-)

hzoo:
Added to 3.0 - should be close or leave this open?

mrjoelkemp:
Can we reference this PR as being fixed from a commit in 3.0? Ideally, this
wouldn't be closed until 3.0 is merged to master

hzoo:
Yeah it doesn't get closed until a ref commit is merged to master. This is just something to keep in mind when doing 3.0 for semver/adding docs for APIs so people can make rules more easily.

PR #1087, opened 2015-02-19
validateIndentation: autofixing!

mikesherov:


gero3:
https://github.com/gero3/three.js/commit/20267dcac018384f35937ae54e0d120d4024058e?diff=split#diff-85fecdb9ce320722ba628969997a333bR233

There seems to be a problem here. This is code after run through the autofix.

mdevils:
Great job, Mike!

I would add one more assertion method `assert.indentation({token, whitespace, message})` or so. I don't like `whitespaceBefore` being changed in rules.

mikesherov:
@gero3, good catch. I'll whip up a test case to catch that.

@mdevils, the routine of splitting whiteSpaceBefore by linebreak, etc. needs to be used for multiple assertions anyway, so glad to move it there. Just wanted some feedback.

mikesherov:
@gero3 actually, it's due to the inconsistent use of `} else if...` with `}\nelse if`. I'll need to patch that, good catch. Leaving this here as my reference: https://github.com/gero3/three.js/blob/20267dcac018384f35937ae54e0d120d4024058e/examples/js/loaders/gltf/glTFLoader.js

mikesherov:
@gero3 can you test this against three.js again? That was incredibly useful.

gero3:
[Commit](https://github.com/gero3/three.js/commit/68299b4541789169385bc5286d84452e97849757) is run on the master of three.js with following config.
```javascript
{"validateIndentation": "\t",
    excludeFiles: [
        "build/*",
        "docs/*",
        "editor/js/libs/*",
        "examples/fonts/*",
        "examples/js/libs/*",
        "examples/models/*",
        "examples/obj/*",
        "examples/scenes/*",
        "test/*",
        "utils/*"]}
```

[Shouldn't this need to be the same Indentation as the line above it ](https://github.com/gero3/three.js/commit/68299b4541789169385bc5286d84452e97849757?diff=unified#diff-919bca0e6d0dc414aa91038a5993a69dR9)

[Should this curly brace be indented??](https://github.com/gero3/three.js/commit/68299b4541789169385bc5286d84452e97849757?diff=split#diff-71c40419dcb775cfdcfedb6e63883e80R620)

mikesherov:
@gero3, wow, the three.js codebase is all over the place with regards to style. Good to use it as a fuzzing tooling ;-)

gero3:
Not really, the library source itself is quite structured but the examples are a mess now and there are a lot of example files. Which is why we want a autofixer of course.

gero3:
There is a problem in the fact that whitespacebefore always gets deleted.

[File to be autofixed](https://gist.github.com/gero3/4569e239f984903a0ec3/d0aaa8fbeafbd9d174f719c93054097dba023710#file-test-js-L224-L226)
[Fixed file](https://gist.github.com/gero3/4569e239f984903a0ec3/7aa9f5bbe66c39faac22a96fc6d000d5ad0f27a2#file-test-js-L212)

As you can see, it removed the linefeeds.

mikesherov:
@gero3 give it a shot now. Looks like I caught all the edge cases you've reported.

mikesherov:
@mdevils can you rereview? @markelog @zxqfox ?

lib/js-file.js
```diff
@@ -43,6 +44,29 @@ var JsFile = function(filename, source, tree, options) {
 
 JsFile.prototype = {
     /**
+     * Returns the first line break character encountered in the file.
+     * Assumes LF if the file is only one line.
+     *
+     * @returns {String}
+     */
+    getLineBreak: function() {
```
zxqfox: Are we fine with this name? ;-) `newLine`? `firstLineBreak`?

lib/js-file.js
```diff
@@ -529,12 +553,10 @@ JsFile.prototype = {
 
     /**
      * Returns array of source lines for the file with comments removed.
-     * Will report erroneous trailing tokens in multiline comments if an error reporter is provided.
```
zxqfox: :+1:

lib/rules/validate-indentation.js
```diff
@@ -246,31 +245,54 @@ module.exports.prototype = {
...
+                        indentChar: indentChar
+                    });
+
+                    // for multiline statements, we need move subsequent lines over the correct
+                    // number of spaces to match the change made to the first line of the statement.
+                    lineAugment = expectedIndentation - actualIndentation;
+
+                    // correct the indentation so that future lines can be validated appropriately
+                    actualIndentation = expectedIndentation;
+                } else if (!line.empty) {
```
zxqfox: Non-blocker: empty line before `} else if {` ?

zxqfox:
@mikesherov :+1: Feels like you almost found all indentation edge cases ;-) Gratz!

lib/js-file.js
mikesherov: I called it `getLineBreak` because there *should* be consistent line breaks in a file.

gero3:
Next test:

[commit](https://github.com/gero3/three.js/commit/a57ddaf6d16724380b370c7713bdfceeb724b3f0#diff-ed8a31150cbb5b3a5052a4b11ddf6407)

I think there is only 1 problem left:
[shouldn't stop be on the same indetation as the brace](https://github.com/gero3/three.js/commit/a57ddaf6d16724380b370c7713bdfceeb724b3f0#diff-ed8a31150cbb5b3a5052a4b11ddf6407R53)

markelog:
FYI: https://github.com/jscs-dev/node-jscs/issues/1084.

I have talked about this with @mdevils recently and bring it up couple times via mail, this rule like many others feels very specific in autofixing terms. Like what other rule would use those assertions?

It feels we need something like `fix` field in rule declarations like in @gero3 fork otherwise our assertion module would became very big and hard to maintain.

Thoughts?

mikesherov:
the new line rules need indentation. The other indentation rules need indentation. The fact that they aren't there *yet* is a symptom of our iterative approach, not a design flaw, IMO.

all if the assertions will be used in multiple places. 

With that said, yes, we'll likely need a "fix" property for non trivial changes, as I've mentioned before.

But there doesn't seem to be many other needs for additional assertions, and I think the fear of an overblown is just that, fear.

markelog:
Oh yeah, even if we add assertions to every in our codebase it wouldn't mean any external rule could use. Which would mean auto-fixing wouldn't possible for the them.

On the other hand, given them api to change your code is dangerous. 

Opened another issue in order not to clutter this one - https://github.com/jscs-dev/node-jscs/issues/1113

lib/js-file.js
mdevils: I would vote for renaming to something like `getLineBreakStyle` or `getLineBreakType`

test/token-assert.js
```diff
@@ -940,4 +940,183 @@ describe('modules/token-assert', function() {
...
+                lineNumber: 1,
+                actual: 2,
+                expected: 0,
+                indentChar: ' '
+            });
+
+            assert(onError.called);
+        });
+
+        it('with silent option, should trigger on incorrect indentation', function() {
```
mdevils: should not?

lib/token-assert.js
```diff
@@ -78,6 +78,114 @@ TokenAssert.prototype.noWhitespaceBetween = function(options) {
 };
 
 /**
+ * Requires the specified line to have the expected indentation.
+ *
+ * @param {Number} options.lineNumber
+ * @param {Number} options.actual
+ * @param {Number} options.expected
+ * @param {String} options.indentChar
+ * @param {Boolean} [options.silent]
```
mdevils: Can you explain the purpose of this option?

mdevils:
Great job again, Mike!

TheSavior:
@mikesherov, Have you looked at the tests for esformatter to see if there are things that we should also be testing for? If we could pass their testsuite, that would be awesome. :)

mikesherov:
@TheSavior, feel like porting it over?

TheSavior:
I can take a look into this. Do we have an example test in our repo yet taking input and testing indentation? It seems like it won't be rule based like the rest of our fix tests are.

mikesherov:
@TheSavior I added an input-output test with this very PR :-) 

Thanks for looking into it!

PR #1090, opened 2015-02-19
Tests for Fixing Newlines

TheSavior:
Adding some tests to two rules that check that they are fixed properly.

I'm not super happy with these test being here. I originally thought these tests should exist in the rules, but now I believe I have a better idea.

Instead of testing all of the rules that they get fixed properly (which has too much overlap with testing `checker.fixString`) we should:

 - [ ] test that the line assertions state `fixed` and have the correct amount of `whitespaceBefore`
 - [ ] test ton `checker.fixString` when given a set of tokens with errors and `whitespaceBefore` outputs the right string

But we shouldn't be making all of the tests to `checker.fixString` use actual rules. We should only use rules with `checker.fixString` as a final integration test of sorts.

### DIFF
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -143,7 +143,7 @@ TokenAssert.prototype.differentLine = function(options) {
 TokenAssert.prototype.linesBetween = function(options) {
     var token = options.token;
     var nextToken = options.nextToken;
-    var whitespaceBefore = '';
+    var whitespaceBefore = ' ';
 
     if (!token || !nextToken) {
         return;
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -143,7 +143,7 @@ TokenAssert.prototype.differentLine = function(options) {
 TokenAssert.prototype.linesBetween = function(options) {
     var token = options.token;
     var nextToken = options.nextToken;
-    var whitespaceBefore = ' ';
+    var whitespaceBefore = '';
 
     if (!token || !nextToken) {
         return;
@@ -198,6 +198,7 @@ TokenAssert.prototype.linesBetween = function(options) {
         if (exactly > 0) {
             whitespaceBefore = new Array(exactly).join('\n');
         }
+
         emitError('exactly ' + exactly, whitespaceBefore);
     }
 };
--- a/test/string-checker.js
+++ b/test/string-checker.js
@@ -147,39 +147,41 @@ describe('modules/string-checker', function() {
     });
 
     describe('fixString', function() {
-        it('should apply fixes to the specified string', function() {
-            checker.configure({ requireSpaceBeforeBinaryOperators: true });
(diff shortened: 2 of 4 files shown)


mikesherov:
@TheSavior Thanks as always for contributing. 

Each pull request you send is getting more and more instances of merging master into your branch. Can you please rebase your PRs so that they don't contain merge commits?

lib/token-assert.js
```diff
@@ -143,7 +143,7 @@ TokenAssert.prototype.differentLine = function(options) {
 TokenAssert.prototype.linesBetween = function(options) {
     var token = options.token;
     var nextToken = options.nextToken;
-    var whitespaceBefore = ' ';
+    var whitespaceBefore = '';
```
mikesherov: this should remain a `' '` in the case where the two tokens joining would result in nonsense, e.g.:

```
typeof
a === 'function';
```
->
```
typeofa === 'function';
```

mikesherov:
I actually do think that individual rules have a test for fixing, even if it overlaps with `fixString`. That way, if we ever change out the implementation underneath, or modify how a rule behaves, we still have proof that it is autofixed.

lib/token-assert.js
TheSavior: That is definitely a valid case. It makes me sad though that we are adding a space on rules that only care about lines. :-/

I'll change this back.

TheSavior:
I've apparently been rebasing from origin/master instead of upstream/master. Since I've been keeping my fork up to date, I've ended up with a bunch of merges from upstream into origin. Perhaps there is a better way for me to be managing my fork.

I've rebased this from upstream/master.

mikesherov:
@TheSavior typically I keep my local master up to date with upstream master, and then rebase against my local master.

zxqfox:
@TheSavior @mikesherov My usual way is `git pull --rebase upstream master` ;-)

mdevils:
Great job!

test/string-checker.js
```diff
@@ -147,39 +147,41 @@ describe('modules/string-checker', function() {
...
+                assert.equal(result.output, 'x = 1 + 2;');
+            });
 
-        it('should return unfixable errors', function() {
-            checker.configure({ disallowImplicitTypeConversion: ['boolean'] });
-            var result = checker.fixString('x = !!x;');
-            assert.equal(result.errors.getErrorCount(), 1);
-            assert.equal(result.errors.getErrorList()[0].message, 'Implicit boolean conversion');
-        });
+            it('should return unfixable errors', function() {
```
mikesherov: Not sure why you nested this and the following two tests under `space rules`?

TheSavior: Great question. I didn't read the tests. :P

TheSavior:
This should be good to go now.

PR #1096, opened 2015-02-21
requireSpacesInsideArrayBrackets: comments should be also taken into account

gero3:
disallowSpaceBeforeBinaryOperators should include comments in the token list. This was found when using the autofixer with the mdcs prset against the three.js library.

### DIFF
--- a/lib/rules/require-spaces-inside-array-brackets.js
+++ b/lib/rules/require-spaces-inside-array-brackets.js
@@ -95,9 +95,9 @@ module.exports.prototype = {
 
         file.iterateNodesByType('ArrayExpression', function(node) {
             var openBracket = file.getFirstNodeToken(node);
-            var afterOpen = file.getNextToken(openBracket);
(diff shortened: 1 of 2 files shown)


mikesherov:
@gero3 what was the test case that was broken using the autofixer? Did it eat the comment token? If so, then we need to follow the "detect error, but don't fix" strategy from linesBetween when a rule itself doesn't account for comments.

gero3:
It kept saying there was an error while there was already 1 space between the comment and the bracket. So It fixed the error but kept reporting it as an error.

mrjoelkemp:
@mikesherov Safe to land once rebased? 

The merge conflict is basically this test being in the wrong folder (it's in `test/rules` instead of `test/spec/rules`). The rebase should resolve automatically.

mikesherov:
@mrjoelkemp yeah, but really we also need to solve this at the autofixer level, like I did for linesBetween.

mrjoelkemp:
Thanks @gero3!

PR #1097, opened 2015-02-21
token-assert:Make sure newlines get fixed

gero3:
The new lines rules should now be able to get fixed too.

### DIFF
--- a/test/token-assert.js
+++ b/test/token-assert.js
@@ -564,6 +564,38 @@ describe('modules/token-assert', function() {
 
                 assert(!onError.called);
             });
+
+            it('should edit the whitespaceBefore', function() {
+                var file = createJsFile('x\n=y;');
+
+                var tokenAssert = new TokenAssert(file);
+                tokenAssert.on('error', function() {});
+
+                var tokens = file.getTokens();
+                tokenAssert.linesBetween({
+                    token: tokens[0],
+                    nextToken: tokens[1],
+                    atLeast: 2
+                });
+
+                assert(tokens[1].whitespaceBefore.length = 2);
+            });
+
+            it('should edit the whitespaceBefore', function() {
+                var file = createJsFile('x\n\n\n=y;');
+
+                var tokenAssert = new TokenAssert(file);
+                tokenAssert.on('error', function() {});
+
+                var tokens = file.getTokens();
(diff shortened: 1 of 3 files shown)


mikesherov:
@gero3, good work! Just needs more tests, one for each of the conditions.

test/token-assert.js
```diff
@@ -564,6 +564,40 @@ describe('modules/token-assert', function() {
 
                 assert(!onError.called);
             });
+
+            it('should edit the whitespaceBefore', function() {
+                var file = createJsFile('x\n=y;');
+
+                var tokenAssert = new TokenAssert(file);
+                var onError = sinon.spy();
```
mikesherov: Is the spy needed in each of these tests? If not, please remove. Thanks!

lib/string-checker.js
```diff
@@ -238,7 +238,6 @@ StringChecker.prototype = {
 
                 source = file.render();
                 sourceTree = JsFile.parse(source, this._esprima, this._configuration.getEsprimaOptions());
-
```
mikesherov: Remove this unnecessary whitespace change. Thanks!

mikesherov:
Thanks again @gero3!

test/token-assert.js
gero3: fixed

test/token-assert.js
```diff
@@ -564,6 +564,38 @@ describe('modules/token-assert', function() {
 
                 assert(!onError.called);
             });
+
+            it('should edit the whitespaceBefore', function() {
+                var file = createJsFile('x\n=y;');
+
+                var tokenAssert = new TokenAssert(file);
+                tokenAssert.on('error', function() {});
```
mikesherov: I should have been more clear. Are the event listeners needed either? Remove this line in each test if not.

gero3: It throws an exception if I don't listen to the event. That is why I'm simply adding the function.

> TypeError: Uncaught, unspecified "error" event.

mikesherov: Ok. Thanks!

mikesherov:
I ended up rewriting the tests because there were a few errors in them. Thanks again @gero3

PR #1098, opened 2015-02-21
requireLineFeedAtFileEnd: make use of assert

gero3:


mikesherov:
Looks goods, just needs more tests: 2 tests in JsFile, 1 for each function you added. 

Thanks again @gero3!

lib/rules/require-line-feed-at-file-end.js
```diff
@@ -34,10 +34,14 @@ module.exports.prototype = {
...
 
     check: function(file, errors) {
-        var lines = file.getLines();
-        if (lines[lines.length - 1] !== '') {
-            errors.add('Missing line feed at file end', lines.length, 0);
-        }
+        var lastToken = file.getLastFileToken();
+        var prevToken = file.getPrevToken(lastToken, {includeComments: true});
+        errors.assert.linesBetween({
+                token: prevToken,
```
mikesherov: This and the next 4 lines are over indented.

lib/rules/require-line-feed-at-file-end.js
```diff
@@ -34,10 +34,14 @@ module.exports.prototype = {
...
-        if (lines[lines.length - 1] !== '') {
-            errors.add('Missing line feed at file end', lines.length, 0);
-        }
+        var lastToken = file.getLastFileToken();
+        var prevToken = file.getPrevToken(lastToken, {includeComments: true});
+        errors.assert.linesBetween({
+            token: prevToken,
+            nextToken: lastToken,
+            atLeast: 1,
+            message: 'Expected a padding newline before closing curly brace'
```
hzoo: Error message should be "Missing line feed at file end" still haha?

gero3: fixed

lib/js-file.js
```diff
@@ -191,6 +191,24 @@ JsFile.prototype = {
     },
 
     /**
+     * Returns the first token for the file from the AST.
+     *
+     * @returns {Object}
+     */
+    getFirstFileToken: function() {
```
mikesherov: One last nit. This should be called `getFirstToken` instead, because the `File` is redundant. Same goes for `getLastToken`.

gero3: fixed

gero3:
Is there something wrong with this pull because appveyor failed??

PR #1101, opened 2015-02-22
(disallow/require)PaddingNewLinesAfterBlocks: Ignoring the end of files

TheSavior:
Fixing an error since we are now adding the EOF token into the mix.

### DIFF
--- a/lib/rules/disallow-padding-newlines-after-blocks.js
+++ b/lib/rules/disallow-padding-newlines-after-blocks.js
@@ -83,7 +83,7 @@ module.exports.prototype = {
             var closingBracket = file.getLastNodeToken(node);
             var nextToken = file.getNextToken(closingBracket);
 
-            while (nextToken !== undefined) {
+            while (nextToken.type !== 'EOF') {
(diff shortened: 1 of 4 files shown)


mikesherov:
:+1: good catch! Have you checked other rules to see if they're affecyed?

TheSavior:
I did a search through for == undefined (to also catch === and !==) and didn't find any other comparisons like this.

mikesherov:
How about `!.*token` regex search?

TheSavior:
Still nope.

There are some `if (!token || !nextToken) {` checks, but those are in the assertions and belong.

There are a couple of rules that use `if(!prevToken)`, but since they are only iterating backwards, we wouldn't have to add a check that stops it when it is EOF instead of on undefined.

It looks like we are all good on adding the EOF token, once we roll out we will definitely want to watch for error reports just in case though.

mikesherov:
:+1:

TheSavior:
Bump?

mikesherov:
Thanks again!

Issue #1064, opened 2015-02-14
Use yeoman/insight to collect usage stats of rules and features


mrjoelkemp:
We currently have some insight into preset popularity based on the github traffic, but no insight into other valuable data-points: rule/preset usage; runtime errors; esnext adoption; and the use of various, big-effort, cli options like autoconfigure and autofixing. 

Should we use a 3rd party library (ex: Mixpanel, https://www.npmjs.com/package/universal-analytics for Google analytics, or another) to determine rules/presets/features that should be moved out of core and give us insight as to where to best devote our dev time?

The stats that we currently have:
![screen shot 2015-02-14 at 1 12 42 pm](https://cloud.githubusercontent.com/assets/865203/6200549/3bb9d20c-b44b-11e4-9ecc-a31b88ddd845.png)

Google Analytics is probably the best bet to keep things free at JSCS' scale.

Is there any hesitation from the community on having usage tracked?

hzoo:
I think would be also cool to somehow notify people of new rules/options they can add to their `.jscsrc` after updating jscs - like a cmd command prompt for them to fix errors/add the new rule or null the rule. Of course they can check the patch notes but this would be easier. 

So it would be similar to the auto-generation of a config file but this would be if they already have a config and want to easily see what new things they would add to it.

zxqfox:
Afaik it's pretty easy to do and we can just ask about it on first use. I'd like to have it in JSCS.

e.g. https://github.com/bower/bower/blob/0ce7053598bf8b79f059c61bdd7afab08497868b/lib/util/analytics.js

mikesherov:
Would be cool when someone uses autoconfigure to ask them to send the final config to us.

mdevils:
Nice idea @mrjoelkemp!
+1 to @mikesherov

mrjoelkemp:
As stated by @markelog outside of this thread, we should use https://www.npmjs.com/package/insight to collect usage data.

Issue #136, opened 2014-01-03
Aggregate and simplify rules


markelog:
Currently, jscs has 49 rules (not counting non-esprima options) and will have more i suspect, much more, most of these rules are pretty simple and can take only one or two possible values like `true/false`, some rules are complete opposite of each other, for example `requireSpaceAfterKeywords`, `disallowSpaceAfterKeywords`, some does not have their significant other, like `disallowSpacesInsideParentheses` doesn't have `requireSpacesInsideParentheses` some complement each other, eg `disallowSpacesInsideArrayBrackets`, `disallowSpacesInsideParentheses`, and so forth...

This amount of rules, their incompleteness could be overwhelming to set up, change and support.

At the beginning, when amount of rules was low, it was simpler and faster to add more features then try to unify them, i think now or in the <a href="https://github.com/mdevils/node-jscs/issues/102">near future</a> it would makes sense to give them more structure.

This –

```js
"requireSpaceAfterKeywords": ["if"],
"disallowSpaceAfterKeywords": ["else"],
"requireSpacesInFunctionExpression": { "beforeOpeningRoundBrace": true, "beforeOpeningCurlyBrace": false },
"disallowMultipleVarDecl": true,
```

could become 

```js
"space after keywords": {
	"require": ["if"],
	"disallow": ["else"]
},
"spaces in function expression": {
	"require": "before opening brace",
	"disallow": "before opening curly brace"
},
"multiple var decl": false

```
or 

```js
"require": {
	"space after keywords": ["if"],
	"spaces in function expression": "before opening brace"
}
"disallow": {
	"space after keywords": ["else"],
	"spaces in function expression": "before opening curly brace",
	"multiple var decl": true,
}
```
or
```js
"require": [
	{ "space after keywords": ["if"] },
	{ "spaces in function expression": "before opening brace" }
],
"disallow": [
	{ "space after keywords": ["else"] }, 
	{ "spaces in function expression": "before opening curly brace" },
	"multiple var decl"
]
```

Or other formats could be adopted, goal is to make rules consistent with each other, if something is allowing something, something could declare the opposite.
If this idea is acceptable, i propose to figure out the new format and what rules are missing.

Although this is pretty big change, but of course, current format should be supported too.

hzoo:
[...]

PR #1102, opened 2015-02-23
Fixes for requirePaddingNewLinesBeforeLineComments (consecutive and in blocks)

TheSavior:
requirePaddingNewLinesBeforeLineComments: Allowing multiple line comments together. Allowing no padding if first line in block.

### DIFF
--- a/lib/rules/require-padding-newlines-before-line-comments.js
+++ b/lib/rules/require-padding-newlines-before-line-comments.js
@@ -3,21 +3,42 @@
  *
  * Type: `Boolean`
  *
- * Values: `true`
+ * Values:
+ * - `true`: always require a newline before line comments
+ * - `Object`:
+ *      - `"allExcept"`: `"firstInBlock"` Comments may be first line of block without extra padding
  *
- * #### Example
- *
+ * #### Examples
  * ```js
  * "requirePaddingNewLinesBeforeLineComments": true
+ * "requirePaddingNewLinesBeforeLineComments": { "allExcept": "firstInBlock" }
  * ```
  *
- * ##### Valid
+ * ##### Valid for `true`
  *
  * ```js
  * var a = 2;
  *
  * // comment
  * return a;
+ *
+ * function() {
+ *
+ *   // comment
+ * }
+ * ```
+ *
+ * ##### Valid for `{ "allExcept": "firstInBlock" }`
+ *
+ * ```js
+ * var a = 2;
+ *
+ * // comment
+ * return a;
+ *
+ * function() {
+ *   // comment
+ * }
  * ```
  *
  * ##### Invalid
@@ -26,6 +47,10 @@
  * var a = 2;
  * //comment
  * return a;
+ *
+ * function() {
+ *   // comment
+ * }
  * ```
  */
 
@@ -36,10 +61,18 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(value) {
-        assert(
-            value === true,
-            'requirePaddingNewLinesBeforeLineComments option requires true value or should be removed'
-        );
+        this._allowFirstInBlock = false;
+
+        if (typeof value === 'object') {
+            assert(typeof value.allExcept === 'string' && value.allExcept === 'firstInBlock',
+                'requirePaddingNewLinesBeforeLineComments option requires the "allExcept" ' +
+                 'property to equal "firstInBlock"');
+            this._allowFirstInBlock = true;
+        } else {
+            assert(value === true,
+                'requirePaddingNewLinesBeforeLineComments option requires true value or object'
+            );
+        }
     },
 
     getOptionName: function() {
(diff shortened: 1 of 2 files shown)


lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -52,8 +52,20 @@ module.exports.prototype = {
...
 
+            var prevToken = file.getPrevToken(comment, {
+                includeComments: true
+            });
+
+            if (prevToken.type === 'Line') {
+                return;
+            }
+
+            if (prevToken.type === 'Punctuator' && prevToken.value === '{') {
```
mikesherov: This should be optional. `allExcept: 'firstInBlock`.

mikesherov: Also, this isn't *really* correct as it allows inside Objects, which aren't blocks, but I suppose it's fine for now.

mikesherov:
@TheSavior, thanks again for contributing!

lib/rules/require-padding-newlines-before-line-comments.js
TheSavior: I'm thinking that since we haven't rolled this rule out yet, we aren't breaking backwards compatibility. I don't think I've met anyone who would expect this rule to force:

```javascript
if (true) {

  // comment
}
```

While I could totally see it being optional in the future and adding more options, I think it is fine to leave it like this for the initial release.

TheSavior: Hmm, your comment on the other PR implies that jQuery would expect the form I put on my last comment. Interesting...

mikesherov: Yeah, that's why I mentioned it :)

TheSavior: How do we decide on the default behavior here? 

It seems like if this rule is used 99% of the time with `allExcept: 'firstInBlock'`then that should be the default. But it would be speculation on both of our sides to try to guess which would be the 'more common' case.

zxqfox: There are #698 and #136 (probably something else) where we can change the default value. And also #1064 could help us to know what default value should be for each rule. ;-)

mikesherov: Default should be the version without exceptions. Principle of least surprise.

TheSavior: @hshoff: Can you weigh in here? Would Airbnb want an override that wouldn't require a blank line before a comment if it is the first in the block? 

Aka, should this be an option at all or just always require a blank line?

hshoff: @TheSavior this would be bad in Airbnb land

```js
// bad
if (true) {

  // comment
}

// good
if (true) {
  // comment
}
```

TheSavior:
Alrighty. An additional configuration option has been added. Can you guys review my doc block? I copied it from another rule but I want to make sure that it will be parsed correctly and is in the correct style.

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -47,13 +83,28 @@ module.exports.prototype = {
...
+            var prevToken = file.getPrevToken(comment, {
+                includeComments: true
+            });
+
+            if (prevToken.type === 'Line') {
+                return;
+            }
+
+            if (allowFirstInBlock && prevToken.type === 'Punctuator' && prevToken.value === '{') {
+                console.log('skip');
```
hzoo: don't need this now

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -26,6 +37,23 @@
  * var a = 2;
  * //comment
  * return a;
+ *
+ * function() {
+ *   // comment
+ * }
+ * ```
+ *
+ * ##### Valid for `{ "allExcept": "firstInBlock" }`
```
hzoo: maybe move this section above `Invalid`? I think that's how we do it

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -3,21 +3,32 @@
...
+ * - `"true"`: always require a newline before line comments
+ * - `Object`:
+ *      - `allExcept`: `firstInBlock` Comments may be first line of block without extra padding
  *
  * #### Example
  *
  * ```js
  * "requirePaddingNewLinesBeforeLineComments": true
  * ```
+ * ```js
```
hzoo: You can just use one ````js`

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -3,21 +3,32 @@
  *
  * Type: `Boolean`
  *
- * Values: `true`
+ * Values:
+ * - `"true"`: always require a newline before line comments
```
hzoo: I don't think you need the double quotes here

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -3,21 +3,32 @@
...
+ * - `Object`:
+ *      - `allExcept`: `firstInBlock` Comments may be first line of block without extra padding
```
zxqfox: But here you need double quotes ;-)

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -47,13 +83,28 @@ module.exports.prototype = {
...
     check: function(file, errors) {
+        var allowFirstInBlock = this._allowFirstInBlock;
+
         file.iterateTokensByType('Line', function(comment) {
             if (comment.loc.start.line === 1) {
                 return;
             }
 
+            var prevToken = file.getPrevToken(comment, {
+                includeComments: true
```
zxqfox: You are against single line object declaration? ;-)

TheSavior: Yup. :)

TheSavior:
Bump?

mikesherov:
@TheSavior no reason to bump. This will land before we release 1.12 regardless of when it actually makes it in

TheSavior:
Ah, gotcha. The flow of when things get merged in on JSCS is just a bit different than what I would expect. Still getting used to it. :)

I was figuring that I would update #1103 after this and #1101 were merged, perhaps I should update it now so it will be ready for you to merge them all at the same time.

mikesherov:
Thanks again!

Issue #1099, opened 2015-02-21
WordPress VIP preset


paulschreiber:
It would be helpful to have a WordPress VIP preset. Their style rules — based on jQuery — are here:
https://make.wordpress.org/core/handbook/coding-standards/javascript/

hzoo:
Right now the Idiomatic (referenced in the wordpress style guide) preset i'm working on is also based on jquery. Shouldn't be too hard to add if we want to do this.

hzoo:
WIP

Wordpress | JSCS
---------------|---------
base styles | jquery preset
indentation with tabs | `"validateIndentation": "\t"`
single quotes | `"validateQuoteMarks": "'"`
100 char limit is not strictly enforced | `"maximumLineLength": null`?
Any ! negation operator should have a following space | `requireSpaceAfterPrefixUnaryOperators: ['!']`

paulschreiber:
@hzoo Thanks! I'm happy to help with this — let me know if I should start a new branch or if you have one already.

hzoo:
Yeah I'l begin a PR soon so we can start adding rules

markelog:
VIP?

paulschreiber:
@markelog WordPress VIP is the paid-hosting version of WordPress. For PHP, the WordPress VIP coding standards add some additional rules to the WordPress coding standards. 
https://vip.wordpress.com/documentation/code-review-what-we-look-for/
https://vip.wordpress.com/documentation/code-and-theme-review-process/
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards

For JavaScript, I don't think there are additional rules; we can refer to this as the "WordPress" preset.

@westonruter and/or @joshbetz can confirm if this is the case.

westonruter:
> It would be helpful to have a WordPress VIP preset. Their style rules — based on jQuery — are here:
https://make.wordpress.org/core/handbook/coding-standards/javascript/

Note that these aren't WordPress.com VIP standards. They are *WordPress Core* standards. So the JSCS naming standard should reflect that. For PHP_CodeSniffer we have a “WordPress” superset standard library with all of the sniffs, and then we have subset standards for “WordPress-Core”, “WordPress-VIP”, and “WordPress-Extra”.

[...]

PR #1105, opened 2015-02-23
Preset: wordpress

hzoo:
For #1099.

Most of the code examples are the same as the jQuery style guide. Can also use this to see if we need to update anymore jQuery preset rules?

I had some issues with `requireYodaConditions`.
In their style guide they use
```
if ( undefOrNull == null ) {
    // Expressions
}
// and
for ( i = 0; i < 100; i++ ) {
```
which I had to switch to make the file pass. So either their style guide needs a change or we need an exception for the rule (most likely) for `for loops` and `checking against null`?

Oh.
`For consistency with the PHP code standards, whenever you are comparing an object to a string, boolean, integer, or other constant or literal, the variable should always be put on the right hand side, and the constant or literal put on the left.`

also I just removed `var` or used `function a() { }` to get around the `requireMultipleVarDecl` errors.

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -10,6 +10,7 @@
  * [MDCS](https://github.com/jscs-dev/node-jscs/blob/master/presets/mdcs.json) — [https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style™](https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style%E2%84%A2)
  * [node-style-guide](https://github.com/jscs-dev/node-jscs/blob/master/presets/node-style-guide.json) - https://github.com/felixge/node-style-guide
  * [Wikimedia](https://github.com/jscs-dev/node-jscs/blob/master/presets/wikimedia.json) — https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript
+  * [Wordpress](https://github.com/jscs-dev/node-jscs/blob/master/presets/wordpress.json) — https://make.wordpress.org/core/handbook/coding-standards/javascript/
  * [Yandex](https://github.com/jscs-dev/node-jscs/blob/master/presets/yandex.json) — https://github.com/yandex/codestyle/blob/master/javascript.md
 
 ## Friendly packages
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -712,6 +712,9 @@ Configuration.prototype.registerDefaultPresets = function() {
     // https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript
     this.registerPreset('wikimedia', require('../../presets/wikimedia.json'));
 
+    // https://make.wordpress.org/core/handbook/coding-standards/javascript/
+    this.registerPreset('wordpress', require('../../presets/wordpress.json'));
+
     // https://github.com/yandex/codestyle/blob/master/javascript.md
     this.registerPreset('yandex', require('../../presets/yandex.json'));
 };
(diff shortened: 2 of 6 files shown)


hzoo:
@paulschreiber thoughts?

paulschreiber:
Thanks! Just tried this out. I've found one problem — it's not allowing me to indent with tabs. When it sees one tab, it gives a `Expected indentation of 4 characters` error.

hzoo:
Hmm that shouldn't be the case though because there aren't an rules for that (no `validateIndentation` rule in this or in the jquery preset). I tried with tabs and have no issues?

paulschreiber:
It should require tabs for indentation (`"validateIndentation": "\t"`?).

Here's what I see:
```
$ ./bin/jscs -p wordpress  /Users/paul/video.js 
Expected indentation of 4 characters at /Users/paul/video.js :
     1 |(function( window, $ ) {
     2 | function setVideoHeight() {
------------^
     3 |  $( this ).height( $( this ).width() / ( 16 / 9 ) );
     4 | }

Expected indentation of 8 characters at /Users/paul/video.js :
     1 |(function( window, $ ) {
     2 | function setVideoHeight() {
     3 |  $( this ).height( $( this ).width() / ( 16 / 9 ) );
----------------^
     4 | }
     5 |
...snip...
```

And the file:
```
(function( window, $ ) {
	function setVideoHeight() {
		$( this ).height( $( this ).width() / ( 16 / 9 ) );
	}

	$( '.video-placeholder' ).each( setVideoHeight );
	$( window ).on( 'resize', function() {
		$( '.video-placeholder' ).each( setVideoHeight );
	});
})( window, jQuery );
```

hzoo:
Oh ok I didn't add that rule (forgot).

Although even if I didn't add the rule it shouldn't expect anything with indentation at all.
(I ran the file with the same jscsrc and also used `-p wordpress` in cli
image

After adding the rule, it will enforce tabs but otherwise it shouldn't have complained about it.

mrjoelkemp:
@hzoo mind rebasing? Would like to land this.

hzoo:
I was wondering if we wanted `"validateIndentation": "\t"` in jquery or this? @mikesherov?

mikesherov:
It belongs in jquery, yes.

hzoo:
Ok rebased + converted to tabs so it won't fail when jquery requires tabs.

mrjoelkemp:
Awesome stuff @hzoo.

paulschreiber:
@mrjoelkemp When will the next version of jscs be released?

hzoo:
@paulschreiber. Should be soon , but if you need it now you can pin the package.json to the commit that has the change.

```js
{
 "jscs": "jscs-dev/node-jscs#4f6115185485706022ba5c71efbca31ce7b21801"
}
```

paulschreiber:
I'm able to install jscs trunk with `npm install` — but the `grunt-jscs` is tied to 1.5.0, which is what I'd like to update. "Soon" is fine.

hzoo:
Oh ok. Although (I'm not completely sure, maybe it depends on `npm -v`) I think that you can install `jscs` at the root project (for this specific version) and any other pacakges (like `grunt-jscs`) wouldn't download `jscs` as a dependency in `node_modules` and actually use the one you specified yourself.

paulschreiber:
I noticed that JSCS is flagging "Yoda conditions" for non-equality operators. #1026 fixed this.

Can you set the WP preset to have `"DisallowYodaConditions": ["==", "===", "!=", !=="]`?

PR #1114, opened 2015-02-24
TokenAssert: simplify and strengthen linesBetween rules

mikesherov:


mikesherov:
@zxqfox @markelog @mdevils can I get 1 review so I can land this?

zxqfox:
Am I right that we revert some assertions because they are invalid for autofixing? Probably we can fix autofix.

:+1: if we can't.

mikesherov:
@zxqfox one of these rules was using assertions incorrectly by passing same token as token and nextToken, which I've now made throw. The other was assertions incorrectly because it's really just a hacky version of disallow semicolons.

PR #1120, opened 2015-02-26
requireSpaceBeforeBlockStatements: Add number of spaces argument

### LINKED ISSUE #936: Allow requireSpaceBeforeBlockStatements to say "at least 1 space"
v0.11.0 changed `requireSpaceBeforeBlockStatements` to throw an error if you have more than one space before the opening brace. Before 0.11, it would only throw if you had zero spaces.

Could you add a rule for requiring *at least* one space, instead of enforcing *exactly* one?

(Essentially, I'd like some way of opting in to what the rule used to do.)

My use-case: I'm using multiple spaces to line up function bodies.

```js
changeName(studentId, ...args)          { this._change(studentId, 'changeName',          ...args) },
changeCreditsNeeded(studentId, ...args) { this._change(studentId, 'changeCreditsNeeded', ...args) }
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7885177-allow-requirespacebeforeblockstatements-to-say-at-least-1-space?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

olegpesok:
Functionality does not change with a boolean flag, but if an integer is
specified then the integer is the MINIMUM number of spaces required before
a block statement.

Fixes #936

### DIFF
--- a/lib/rules/require-space-before-block-statements.js
+++ b/lib/rules/require-space-before-block-statements.js
@@ -1,14 +1,17 @@
 /**
- * Requires space before block statements (for loops, control structures).
+ * Requires space(s) before block statements (for loops, control structures).
  *
- * Type: `Boolean`
+ * Type: `Boolean` or `Integer`
  *
- * Values: `true`
+ * Values:
+ *
+ * - `true` require a single space
+ * - `Integer` require AT LEAST specified number of spaces
  *
  * #### Example
  *
  * ```js
- * "requireSpaceBeforeBlockStatements": true
+ * "requireSpaceBeforeBlockStatements": 1
  * ```
  *
  * ##### Valid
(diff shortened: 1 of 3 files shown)


mikesherov:
@olegpesok thanks for contributing! Rather than adding a `relaxed` parameter, we should get rid of the `spaces` parameter, and instead add an `exactly`, `atLeast`, and `atMost` field like `linesBetween` has to achieve parity. Once we have that, I can review this PR further.

/cc @markelog or @mdevils  or @zxqfox for opinions

mdevils:
@mikesherov  +1

mikesherov:
Can you please rebase this PR and make use of the new options for spacesBetween rather than including a `relaxed` option?

markelog:
Rebased and merged, thank you

PR #1126, opened 2015-02-28
Configuration: show number of violated rules when using auto-configuration

### LINKED ISSUE #1125: Request: show number of violated rules when using auto-configuration
I'm a big fan of the new auto-configuration feature, but it would be even more helpful to me if its output would additionally show me the number of violated rules per preset as well. That'd give me an even clearer picture of which preset matches the tested codebase best. From my understanding of JSCS this information should be available when the auto-configuration finished testing against all presets?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9057481-request-show-number-of-violated-rules-when-using-auto-configuration?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

fubu:
Fixes #1125

### DIFF
--- a/lib/config/generator.js
+++ b/lib/config/generator.js
@@ -129,7 +129,7 @@ Generator.prototype._showErrorCounts = function(statsForPresets) {
     var table = getTable();
 
     statsForPresets.forEach(function(presetStats, idx) {
-        table.push([idx + 1, presetStats.name, presetStats.sum]);
+        table.push([idx + 1, presetStats.name, presetStats.sum, getUniqueErrorNames(presetStats.errors).length]);
     });
 
     console.log(table.toString());
@@ -247,7 +247,7 @@ function getTable() {
             'padding-left': 0,
             'padding-right': 0
         },
-        head: ['', 'Preset', '#Errors']
+        head: ['', 'Preset', '#Errors', '#Rules']
     });
 }
 
--- a/lib/config/generator.js
+++ b/lib/config/generator.js
@@ -73,19 +73,18 @@ Generator.prototype.generate = function(path) {
 
         this._config.preset = presetName;
 
-        var errorList = statsForPresets[presetIndex].errors;
-        errorList = getUniqueErrorNames(errorList);
+        var errorStats = getErrorsByRuleName(statsForPresets[presetIndex].errors);
 
-        var violatedRuleCount = errorList.length;
+        var violatedRuleCount = Object.keys(errorStats).length;
 
         if (!violatedRuleCount) { return this._config; }
 
         console.log(_path + ' violates ' + violatedRuleCount + ' rule' + (violatedRuleCount > 1 ? 's' : ''));
 
-        var errorPrompts = generateRuleHandlingPrompts(errorList);
+        var errorPrompts = generateRuleHandlingPrompts(errorStats);
 
         return this._getUserViolationChoices(errorPrompts)
-        .then(this._handleViolatedRules.bind(this, errorPrompts, errorList))
+        .then(this._handleViolatedRules.bind(this, errorPrompts))
         .then(function() {
             return this._config;
         }.bind(this));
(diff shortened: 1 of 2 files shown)


mrjoelkemp:
Great work. I need to pull it down and verify soon. A test would be great. Any chance you could whip one up?

fubu:
Not sure what exactly should be tested -- probably not that "#Rules" shows up in the output, but maybe that the reported number of violations is correct? (which boils down to a verification of `getUniqueErrorNames(errors)`)

mrjoelkemp:
Agreed about the testing being not too obvious. The best way to test this is to fix the fixture file; currently it's the source code for the config generator, which I thought was cute, but isn't. 

We need a control file that has a guaranteed results set of errors and violated rules. Then we know for sure what the console output should look like and can split lines to assert the number of violated rules.

For now, I'd say a manual verification is okay since coverage remained the same.

mikesherov:
LGTM

lib/config/generator.js
```diff
@@ -203,19 +201,56 @@ function getChecker() {
...
+        return errors[b].violations - errors[a].violations;
+    });
+
     return violatedRuleNames.map(function(ruleName) {
+        var violationCount = errors[ruleName].violations;
+        var fileCount = Object.keys(errors[ruleName].files).length;
         var prompt = assign({}, prompts[1]);
-        prompt.name = colors.green(ruleName) + ': ' + prompt.name;
+
+        prompt.associatedRuleName = ruleName;
```
mrjoelkemp: This is okay, but only because the prompt module doesn't have a meaning associated with the `associatedRuleName` key. The `prompt` object was intended to be an object of settings specifically for the `prompt` module.

lib/config/generator.js
```diff
@@ -203,19 +201,56 @@ function getChecker() {
...
  * @private
+ * @param  {Object[]} errorList
+ * @return {Object}
+ */
+function getErrorsByRuleName(errorsList) {
+    var errors = {};
+
+    errorsList.forEach(function(error) {
+        var rulename = error.rule;
+        if (!errors.hasOwnProperty(rulename)) {
```
mrjoelkemp: Non-blocker: This works too. An alternative is:

```js
errors[rulename] = errors[rulename] || {
  files: {},
  violations: 0
};
```

mrjoelkemp:
@fubu Great work here. Mind rebasing off of master and then we'll land this.

lib/config/generator.js
fubu: Oops, I didn't realize there was another module involved... :) Unfortunately there seems no way to hand `prompt` some metadata (because in my opinion the "associated rule" should be part of the prompt's state rather than being loosely bound via a random list index, as it was before).

lib/config/generator.js
fubu: :+1: didn't think about this

fubu:
I don't, but my git-skills are a bit rusty. What exactely should I do? :)

mrjoelkemp:
git pull - -  rebase upstream master. Assuming you did: git remote add upstream the_jscs_repo_ssh_url. 

Let me know if that doesn't work.

fubu:
Yep, managed to do it in the meantime, thanks! Don't know about the failed build though, the AppVeyor page takes forever to load...

mrjoelkemp:
@fubu I can't get that build to load on appveyor. Any chance you could force push up an insignificant change to trigger a rebuild?

zxqfox:
You can see a part of errors here: https://ci.appveyor.com/project/mdevils/node-jscs/build/154
https://ci.appveyor.com/api/buildjobs/1v3k79cy024mr1ym/log — and full log here (3.4Mb)

mrjoelkemp:
K. Will give a final look and merge. Thanks @zxqfox

mrjoelkemp:
Great work @fubu. The extra info that you added is really helpful. Thanks for contributing.

Issue #1127, opened 2015-03-01
Add tests to verify whitespace fixes for rules


TheSavior:
I have identified the following rules to need tests that verify whitespace fixes. 

 - [x] DisallowCommaBeforeLineBreak
 - [x] DisallowKeywordsOnNewLine
 - [ ] DisallowMixedSpacesAndTabs
 - [ ] DisallowMultipleLineBreaks
 - [x] DisallowNewLineBeforeBlockStatement
 - [x] DisallowObjectKeysOnNewLine
 - [ ] DisallowOperatorBeforeLineBreak
 - [x] DisallowPaddingNewLinesAfterBlocks
 - [x] DisallowPaddingNewLinesAfterUseStrict
 - [ ] DisallowPaddingNewLinesBeforeKeywords
 - [ ] DisallowPaddingNewLinesBeforeLineComments
 - [ ] DisallowPaddingNewLinesInBlocks
 - [ ] DisallowPaddingNewLinesInObjects
 - [ ] DisallowRightStickiedOperators
 - [ ] DisallowSpaceAfterBinaryOperators
 - [ ] DisallowSpaceAfterKeywords
 - [ ] DisallowSpaceAfterLineComments
 - [ ] DisallowSpaceAfterObjectKeysColons
 - [ ] DisallowSpaceAfterObjectKeys
 - [ ] DisallowSpaceAfterPrefixUnaryOperators
 - [ ] DisallowSpaceBeforeBinaryOperators
 - [ ] DisallowSpaceBeforeBlockStatements
 - [ ] DisallowSpaceBeforeKeywords
 - [ ] DisallowSpaceBeforeObjectValues
 - [ ] DisallowSpaceBeforePostfixUnaryOperators
 - [ ] DisallowSpaceBeforeArguments
 - [ ] DisallowSpacesInAnonymousFunctions
 - [x] DisallowSpacesInAnonymousFunctionExpression
 - [ ] DisallowSpacesInCallExpression
 - [ ] DisallowSpacesInConditionalExpression
 - [ ] DisallowSpacesInForStatement
 - [ ] DisallowSpacesInFunctionDeclaration
 - [ ] DisallowSpacesInFunctionExpression
 - [ ] DisallowSpacesInFunction
 - [ ] DisallowSpacesInNamedFunctionExpression
 - [ ] DisallowSpacesInsideArrayBrackets
 - [ ] DisallowSpacesInsideBrackets
[...]

PR #1128, opened 2015-03-01
Adding some more fix tests

TheSavior:
Adding whitespace tests to a few rules in #1127

### DIFF
--- a/test/specs/rules/disallow-comma-before-line-break.js
+++ b/test/specs/rules/disallow-comma-before-line-break.js
@@ -2,24 +2,65 @@ var Checker = require('../../../lib/checker');
 var assert = require('assert');
 
 describe('rules/disallow-comma-before-line-break', function() {
-
     var checker;
+    var input;
+    var output;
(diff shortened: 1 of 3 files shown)


mikesherov:
@TheSavior great work! Thanks again, as always, for contributing! Perhaps we should make a set of test utility functions, so tests can just say something like:

```
assert.fixesString(input, output);
```
?

mdevils:
Great job!

TheSavior:
@mikesherov I totally agree. Once I got these tests written there is a definite pattern. 

Some tests have an assert that the number of errors is 3 while most verify only 1. It seems like other than an integration test, tests should only ever be asserting one error. Does that seem like a reasonable general rule? Are there obvious cases I'm not thinking of?

mikesherov:
We can solve that with a 3rd param that is the expected number of errors, defaulted to one.

mrjoelkemp:
Good stuff @TheSavior

PR #1132, opened 2015-03-02
Extend estraverse rules to support XJS and JSX

### LINKED ISSUE #1131: Update tree-iterator to support JSX (in newer esprima-fb)
Looks like `esprima-fb@13001.1.0-dev-harmony-fb` updated with a `XJS -> JSX` change. https://github.com/facebook/esprima/commit/2d47a78dc50a8cce10bcef7730a0e81f7782f2d2

I guess we should update the files for this similar to #830? I don't know if we also just want to find-replace or just append?

So `XJSIdentifier` would be `JSXIdentifier`.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9099402-update-tree-iterator-to-support-jsx-in-newer-esprima-fb?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1131
Closes gh-1132

Appends the new keys in `lib/tree-iterator.js`.
Not sure if we want to do it this way - I just added another data file and am testing both files - XJS and JSX.

### DIFF
--- a/lib/tree-iterator.js
+++ b/lib/tree-iterator.js
@@ -32,6 +32,17 @@ function iterate(node, cb) {
                 }
             },
             keys: {
+                JSXIdentifier: [],
+                JSXNamespacedName: ['namespace', 'name'],
+                JSXMemberExpression: ['object', 'property'],
+                JSXEmptyExpression: [],
+                JSXExpressionContainer: ['expression'],
+                JSXElement: ['openingElement', 'closingElement', 'children'],
+                JSXClosingElement: ['name'],
+                JSXOpeningElement: ['name', 'attributes'],
+                JSXAttribute: ['name', 'value'],
+                JSXSpreadAttribute: ['argument'],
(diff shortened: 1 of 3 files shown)


mikesherov:
Thanks!

feross:
Could you do an npm release so we can start using the new JSX support?

hzoo:
Right now I'm just using the commit. I think we're planning on releasing tomorrow though?

```js
"jscs": "jscs-dev/node-jscs#d6fcbc48ab4b09bd1f966a4b9495cd6bf2f9631e",
```

feross:
Thanks.

PR #1142, opened 2015-03-05
New rule: disallowSwitchCaseIndentation

### LINKED ISSUE #1111: Request: disallowSwitchCaseIndentation
If this flag is true, then `case` statements should live at the same indentation level as their parent `switch`, a'la Crockford style.

Valid:
````javascript
switch (x) {
case 'b':
    break;
}
````

Invalid:
````javascript
switch (x) {
    case 'b':
        break;
}
````

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8852174-request-disallowswitchcaseindentation?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

yaycmyk:
As is the requirement in Crockford style, this rule enforces that direct-child
case statements within a switch block must not be indented or have a different
starting column than their parent switch. This is enforced on a per-switch basis,
so a switch nested inside another could itself be indented with the former's cases
having to be at the same indentation level.

Closes gh-1111
Fixes #1111

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -529,7 +529,7 @@ Configuration.prototype.hasPreset = function(presetName) {
 };
 
 /**
- * Registers built-in Code Style cheking rules.
+ * Registers built-in Code Style checking rules.
  */
 Configuration.prototype.registerDefaultRules = function() {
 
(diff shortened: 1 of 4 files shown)


mikesherov:
@yaymck, thanks for contributing! This is better handled as an option to the `validateIndentation` rule. Care to give it a shot there instead?

That rule already allows both styles... should be easy enough to enforce based on a Config value. Otherwise, I can handle it myself. Let me know... Either way I'm going to close this for now.

mikesherov:
@yaycmyk ^

yaycmyk:
@mikesherov I tried playing around in that ruleset and found it pretty hard to follow - will let you grab this one if it absolutely can't live as its own rule.

mikesherov:
Yes, it should not live as its own rule.

PR #1143, opened 2015-03-05
Tests: move specs into a subdir so that tests, fixtures, and utilities aren't intermingled

mikesherov:
This is groundwork to introduce a testing utility infra. First thing to do is split actual tests out from fixtures, test options, and utilities.


mdevils:
:+1:

zxqfox:
:+1:

PR #1144, opened 2015-03-06
ValidateIndentation: Adding a broken test for functions as an argument

TheSavior:
@mikesherov While running my autoformatting plugin, this is the main issue I ran into.

This passes:
```javascript
    fn("foo", function() {
    a();
});
```

But should fail and require:

```javascript
    fn("foo", function() {
        a();
    });
```

This is also an issue with arrays. This shouldn't be valid:
```javascript
        var files = [
    'foo',
    'bar',
    'baz'
];
```

I guessed as to where to put the test, let me know if it should be somewhere else. It's also possible that this is supposed to be covered by some rule other than validateIndentation.

### DIFF
--- a/test/specs/rules/validate-indentation.js
+++ b/test/specs/rules/validate-indentation.js
@@ -233,6 +233,13 @@ describe('rules/validate-indentation', function() {
             '});';
             assert(!checker.checkString(source).isEmpty());
         });
+
+        it('should report function in argument with invalid indentation', function() {
+            var source = '    fn("foo", function() {\n' +
+                        '    a();\n' +
+                        '});';
+            assert(!checker.checkString(source).isEmpty());
+        });
     });
 
     describe('fixing', function() {


mikesherov:
Arrays are separate because people are all over the map with how they get indented. Even the JSCS codebase fails if we enforced array indentation.

In regards to the first failure, it's a tough call because this should pass:

```
var longName =
                       fn('foo', function() {
  a();
});
```

I ere on the side of caution while trying to not introduce too much complexity (it's already complex enough!).

Anyway, I'll see if I can fix this soon, but it likely won't make it into 1.12.

TheSavior:
Can you elaborate a bit more on why that example should pass?

Also, perhaps it's time to have indenting rules beyond just the simple all or nothing validateIndentation. I'm thinking about arrays specifically here since as you say, everyone is doing them differently.

mikesherov:
@TheSavior actually, you're correct about first example.

TheSavior:
Good to hear. FWIW, this has almost been a big enough thorn in my side to go back to another sublime formatter plugin. I hope we can figure out the right way to fix it. It seems to be an issue for all blocks split on multiple lines.

TheSavior:
Is there a potential stop gap fix that just causes the autoformatting to format these blocks properly, but doesn't necessarily validate them properly?

For example this code

```javascript
var x = {
  a: 'foo',
b: {
  c: 'bar'
}
}
```

when autoformatted, is turned into this:
```javascript
var x = {
  a: 'foo',
  b: {
  c: 'bar'
}
}
```

It seems like it could theoretically detect the block and indent the entire thing and then revalidate the indentation. Perhaps I'm missing how this is implemented though.

zxqfox:
@mikesherov Do you thought about making several options-flags to explicitly show what is available?
E.g. `allowArraysWithoutIndent`, `allowArraysWithSingleIndent`, etc.

mikesherov:
@TheSavior sorry that this is causing you so many problems. I'll be picking back up on this soon. I've got important work deadlines that are eating up all my FOSS time at the moment.

markelog:
Just got another one - https://github.com/jscs-dev/node-jscs/issues/1287

markelog:
@mikesherov @TheSavior hey there, do any of you have time to take tackle this one?

mikesherov:
Looked into it. A bit more difficult than I thought, due to lots of edge cases, but I'm actively investigating.

PaquitoSoft:
I guess I have the same issue.

With a code like this:
```
(function() {
	'use strict';

	define(function(require) {
		var a = require('a'),
			b = require('b'),
			c = require('c');
```

When running JSCS, it gets converted to this one:
```
(function() {
	'use strict';

	define(function(require) {
		var a = require('a'),
		b = require('b'),
		c = require('c');
```

But I want to keep the indentation.

This is my config file (I used the Crockford preset):
```
{
    "preset": "crockford",
    "disallowDanglingUnderscores": null,
    "disallowSpaceBeforeBinaryOperators": null,
    "disallowSpacesInFunctionDeclaration": null,
    "disallowSpacesInsideArrayBrackets": null
}
```

Am I missing some configuration option?

Thanks and amazing job!

Issue #895, opened 2015-01-09
Add ability to declare acceptable values for the rules


markelog:
Now this responsibility falls on the `configure` method, but it would much easier if this method would be replaced by some declaration like:

```js
values: [{
    type: "object",
    value: { 
        type: "string",
        value: "allExcept"
    }
}, {
    type: "boolean",
    value: true
}];
```

we couldn't fully dispose from the `configure` method, but it will simplify a lot of things

/cc @mdevils

zxqfox:
It can be replaced with a `function({Conf} conf)` with some `conf.registerRuleValue` inside.

Also one rule could handle several rules and values. And also one rule could have aliases (for modern configuration).

But fully declarative version (yours) will also be good.

mdevils:
As I can remember, @mikesherov wanted to do something about it.

markelog:
> As I can remember, @mikesherov wanted to do something about it.

Yeah, i remember something like that, although no one responded when i repeatedly ask about this, hence the ticket

mrjoelkemp:
Looks good and manageable. Nice work, Oleg.

Who will implement the validation using this? I can help port some rules
once the validator is in master.

We just have to make sure no rule currently overrides this.values in its
configure. If none do, the validator can just check for the existence of
values on a rule and run the validator then configure, or just run
configure. Hence, the validator can be shipped quickly without being used.

markelog:
> Who will implement the validation using this?

Whoever is capable and have time to do this :-), after 1.10 my next item is jquery tasks, there is a lot them accumulated out there.

> We just have to make sure no rule currently overrides this.values

`configure` method usually set only private properties (style which @mdevils set long time ago, nice going btw :-) i.e. with "_" prefix, so we should be good there.

> Hence, the validator can be shipped quickly without being used.

That is my thought exactly.

mikesherov:
I think I still have an open issue about this somewhere...

My thoughts were to make rule values declarative rather than having functions. This would also make documentation auto generate able.

zxqfox:
We can try to use here https://github.com/molnarg/js-schema/ or analogue. But it's 24Kb. Probably we can find more lighten solution.

hzoo:
Came across
- https://github.com/arasatasaygin/is.js
- https://github.com/philbooth/check-types.js
- https://facebook.github.io/react/docs/reusable-components.html#prop-validation - maybe even React.propTypes?

hzoo:
[...]

PR #1148, opened 2015-03-07
use this.getOptionName() for consistency, options variable name

hzoo:


markelog:
That is a lot of thoughtful work, wow. Although, i fear we would have to do it all over again, after we address https://github.com/jscs-dev/node-jscs/issues/895.

Will merge this in couple hours.

mrjoelkemp:
LGTM nice work @hzoo!

hzoo:
I saw we had a few errors in docs since we were typing the rule names over and over and sometimes spelling it wrong. Sometimes its because we just copy/paste from another rule and end up with the wrong name. 

Would be interesting if we could get the option name from the file name but maybe that would be weird?

Yeah I think a declarative config would be great.

mrjoelkemp:
@hzoo The file name idea is a good one, but doesn't work for the browser bundle.

markelog:
Merged, as always, thank you :-); btw, don't bother put "Closes" in the commit body, seems tedious to force-push into PR to put correct pull number. We can always do it for ya.

hzoo:
Yeah it is kind of tedious - sometimes I actually do it before I push since I know the next issue number (by assuming no one else is doing a PR/issue at the time I push)

Issue #1080, opened 2015-02-18
Addtional JSCS Users


hzoo:
We can also add 
- JSHint,
- ~~Babel~~ (has custom syntax so it's using custom babel-eslint now) - would actually be a good project to test new es6 features.
- h5bp (html5-boilerplate)

to `Who uses JSCS?`

PR #1149, opened 2015-03-07
Docs: add JSHint and h5bp to list of users

hzoo:
Fixes: #1080
Closes gh-1149

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,5 @@ including presets from popular style guides like jQuery, Airbnb, Google, and mor
  * [Famous](http://famo.us/)
  * [less.js](http://lesscss.org/)
  * [Goodvidio](http://goodvid.io/)
+ * [JSHint](http://jshint.com/)
+ * [html5-boilerplate](https://html5boilerplate.com/)


markelog:
Seems this list becomes pretty big, deserves its own page, or we can just "everybody uses it" :-)

hzoo:
Haha.. maybe we have a few on the front page and another page with everyone? Or we could do logos/pictures instead to fill up the width. Also do we want to give a link to their actual .jscsrc file if it's on github?

markelog:
> Or we could do logos/pictures instead to fill up the width

I remember that is what @mdevils wanted, preferred way for me too

> Also do we want to give a link to their actual .jscsrc file if it's on github?

It would be hard to support those links

mdevils:
I think the best way is to have icons/logos of the most popular users on the main page and have the full list of users on a separate page. How do you think?

markelog:
I think on the main jscs.info page it's okay, but i would try to remove it from `README.md`.

mrjoelkemp:
Closing in favor of the suggested move of this information.

Issue #1030, opened 2015-02-09
`.getReporter()` should require node_modules packages


UltCombo:
Reference: https://github.com/jscs-dev/gulp-jscs/pull/66#discussion_r24322672

Also, perhaps `.getReporter()`'s `reporter` argument could accept a function (which simply returns the function)? Then we wouldn't need to wrap `.getReporter` in the plugins.

UltCombo:
Though accepting the function as argument is not really necessary, a simple `typeof` check in the plugins would be enough.

zxqfox:
Oops. Totally agreed here.

UltCombo:
:+1:

markelog:
Oh, well can do this only in 2.0, since https://github.com/jscs-dev/node-jscs/issues/1150 is a breaking change ((

PR #1150, opened 2015-03-08
Unify and introduce polymorphic require of external plugins

### LINKED ISSUE #1083: Unify and introduce polymorphic require of external plugins
Talking about these tickets - https://github.com/jscs-dev/node-jscs/issues/109, https://github.com/jscs-dev/node-jscs/issues/1030. I think, they all (and `additionalRules`), should follow same logic as `plugins` options does - i.e. take path or name of the potential package dependency.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8705341-unify-and-introduce-polymorphic-require-of-external-plugins?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Fixes #1083, #1030, #109.

Also add more tests and simplification of configuration modules. Some of these changes might be considered as breaking.

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -60,23 +60,8 @@ Configuration.prototype.load = function(config) {
     copyConfiguration(overrides, currentConfig);
 
     var ruleSettings = this._processConfig(currentConfig);
-    var processedSettings = {};
 
-    Object.keys(ruleSettings).forEach(function(optionName) {
-        var rule = this._rules[optionName];
-        if (rule) {
-            var optionValue = ruleSettings[optionName];
-            if (optionValue !== null) {
-                rule.configure(ruleSettings[optionName]);
-                this._configuredRules.push(rule);
-                processedSettings[optionName] = ruleSettings[optionName];
-            }
-        } else {
-            this._unsupportedRuleNames.push(optionName);
-        }
-    }, this);
-
-    this._ruleSettings = processedSettings;
+    this._loadRules(ruleSettings);
 };
 
 /**
@@ -285,19 +270,6 @@ Configuration.prototype._processConfig = function(config) {
         options.plugins.forEach(this._loadPlugin, this);
     }
 
-    // Apply presets
-    var presetName = options.preset;
-    if (presetName) {
-        this._presetName = presetName;
-        assert(typeof presetName === 'string', '`preset` option requires string value');
-        var presetData = this._presets[presetName];
-        assert(Boolean(presetData), 'Preset "' + presetName + '" does not exist');
-        var presetResult = this._processConfig(presetData);
-        Object.keys(presetResult).forEach(function(key) {
-            ruleSettings[key] = presetResult[key];
-        });
-    }
-
     // File extensions
     if (options.fileExtensions) {
         assert(
@@ -367,6 +339,11 @@ Configuration.prototype._processConfig = function(config) {
         this._loadVerbose(options.verbose);
     }
 
+    // Apply presets
+    if (options.hasOwnProperty('preset')) {
+        this._loadPreset(options.preset);
+    }
+
     // NOTE: rule setting must come last in order to
     // override any rules that are loaded from a preset
     Object.keys(config).forEach(function(key) {
@@ -388,6 +365,30 @@ Configuration.prototype._loadPlugin = function(plugin) {
     assert(typeof plugin === 'function', '`plugin` should be a function');
     plugin(this);
 };
+
+/**
+ * Load rules.
+ * @param {Object} ruleSettings
+ * @protected
+ */
+Configuration.prototype._loadRules = function(ruleSettings) {
+    Object.keys(ruleSettings).forEach(function(optionName) {
+        var rule = this._rules[optionName];
+        if (rule) {
+            var optionValue = ruleSettings[optionName];
+            if (optionValue !== null) {
+                rule.configure(ruleSettings[optionName]);
+                this._configuredRules.push(rule);
+
+                this._ruleSettings[optionName] = ruleSettings[optionName]
+            }
+        } else {
+            if (!~this._unsupportedRuleNames.indexOf(optionName)) {
+                this._unsupportedRuleNames.push(optionName);
+            }
+        }
+    }, this);
+}
 
 /**
  * Loads an error filter.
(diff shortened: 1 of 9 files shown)


mikesherov:
what do you consider breaking here? Looks good to me otherwise!

markelog:
For example, you could define `additionalRules` like that:

```
{
"additionalRules":  "foo/bar"
}
```

now, you can do it only like that:
```
{
"additionalRules":  "./foo/bar"
}
```
otherwise jscs would consider it a module name

mikesherov:
Is there any way to preserve the old behavior without introducing too much complexity? Would be a shame to have to hold off till 2.0 for this.

zxqfox:
@markelog Why we can't check module `foo` first, and then local `./foo` directory if module doesn't exist?

markelog:
@mikesherov we could consider releasing 2.0 instead of 1.12, there is some tickets/pulls that could be done/landed, plus yeah, auto-fixing is not a breaking change, but its feels like a lot, it could justify the major in pre-semver times. Even today, in many minds version change perceived not by breakage but by the functionality enhancement. 

/cc @mdevils, @mrjoelkemp 

@zxqfox it still be breaking, basically, we can't distinguish between the path `foo` and module `foo`.

mikesherov:
We've already said that autoformatting needs to be in the 1.x line.

markelog:
You have a firm stance on it, did you make any promises?

mikesherov:
No, no promises made, but we already discussed this. No reason to release 2.0 over such a minor bug. If you'd like to revisit that decision, we should discuss that on the ML, no?

markelog:
You right, will send a letter

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
...
+        return null;
+    }
+};
+
+/**
  * Loads plugin data.
  *
- * @param {String|function(Configuration)} plugin
- * @protected
+ @param {String|null|function(Configuration)} plugin
```
zxqfox: null?

mdevils:
@markelog, why are you doing refactoring, adding documentation and so on here? I can't get the idea of this Pull-request. It's lost in tons of code changes.

mdevils:
@markelog, `"additionalRules":  "./foo/bar"` is a breaking change. Even in Yandex we have lots of projects which will be broken after this PR. And at the same time I don't see the purpose of this change.

markelog:
Sorry Marat, just pulled the thread, i tried to make commits as atomic as possible though, you might be better of reviewing them individually.

> "additionalRules": "./foo/bar" is a breaking change.

That is what i have repeatedly said. 

>  And at the same time I don't see the purpose of this change.

This PR mentions tickets that it would solve.

zxqfox:
Would be great if we split this to several PRs. I need a part of it to finish `.jscsignore` related issue. On the other hand it would be safer to merge them.

@markelog I know it's hard to split it. But still can we make a try?

mdevils:
@zxqfox +1

markelog:
> On the other hand it would be safer to merge them.

How so? I don't follow.

Is there something preventing you to review this by the individual commits?

zxqfox:
> Is there something preventing you to review this by the individual commits?

Only fear of incomplete code.

markelog:
Every commit should atomic, if there is any issues please let me know

lib/config/configuration.js
```diff
@@ -24,22 +24,130 @@ var BUILTIN_OPTIONS = {
  * @name Configuration
  */
 function Configuration() {
+    /**
+     * List of the registered (not used) presets.
+     * @protected
```
mrjoelkemp: Newlines after descriptions?

mrjoelkemp: @zxqfox rule for that? :)

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
 };
 
 /**
+ * Load external module.
+ * @param {String|null} external - path (relative or absolute) or name to the external module
```
mrjoelkemp: Newline

mrjoelkemp:
Nice work here @markelog! :+1:

lib/config/configuration.js
zxqfox: Sorry, there is an issue for that :( https://github.com/jscs-dev/jscs-jsdoc/issues/85

mdevils: +1

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
...
+    }
+};
+
+/**
  * Loads plugin data.
  *
- * @param {String|function(Configuration)} plugin
- * @protected
+ @param {String|null|function(Configuration)} plugin
+ * @private
```
mdevils: Why `private` instead of `protected`? You're overriding, so it is `protected`.

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
...
 };
 
 /**
- * Loads an error filter module
+ * Loads preset.
  *
- * @param {String|null} errorFilter
- * @protected
+ * @param {String|null} preset
+ * @private
```
mdevils: Same question about `private`.

lib/config/node-configuration.js
mdevils: +1

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
...
+        Configuration.prototype._loadPreset.call(this, name);
     }
+};
 
-    Configuration.prototype._loadErrorFilter.call(this, errorFilter);
+/**
+ * Loads an error filter module.
+ *
+ * @param {String|null} filter
+ * @private
```
mdevils: Same question about `private`

lib/config/node-configuration.js
```diff
@@ -45,71 +45,108 @@ NodeConfiguration.prototype.overrideFromCLI = function(program) {
 };
 
 /**
+ * Load external module.
+ * @param {String|null} external - path (relative or absolute) or name to the external module
+ * @param {String} type - type of the module
+ * @return {Module}
```
mdevils: This should be a `@protected` function `_loadExternal`.

PR #1152, opened 2015-03-08
CLI: increase coverage of the "cli" module

markelog:


test/specs/cli.js
```diff
@@ -85,6 +87,52 @@ describe('modules/cli', function() {
...
+            args: ['test/data/cli/error.js'],
+            preset: 'google'
+        });
+
+        return result.promise.always(function() {
+            var i;
+            var config = Checker.prototype.configure.getCall(0).args[0];
+
+            for (i in config) {
+                assert(false);
```
mrjoelkemp: assert(!Object.keys(config)) instead of the for and assert(true)?

markelog: Cool, updated

mrjoelkemp:
Lgtm

zxqfox:
:+1: Same

Can we land this?

PR #1153, opened 2015-03-08
Misc: update dependencies

markelog:
Also, change update policy of the touched packages


mrjoelkemp:
Carets? Why?

markelog:
To get more rapid updates, is there any concerns?

mrjoelkemp:
Yes, package authors shouldn't be trusted :) 

Our users shouldn't have their builds break under their feet because of one of our dependencies pushing a broken minor (even patch) update. So ideally, we'd have fixed versions. Carets create a house of cards in dependency management.

markelog:
If we not gonna use shrinkwrap, dependency tree wouldn't be stable anyway, since fixed deps would work only if all our deps would use fixed deps.

mrjoelkemp:
Good point.

mikesherov:
:+1: land at will.

zxqfox:
I've restarted appveyor, ETIMEOUT happend.

zxqfox:
Hm:
```
.....................................................[93m.[93m.[93m. 
  [93m.[93m.......................................................
```

What's that in `appveyor` logs?

:+1: If it's fine.

zxqfox:
> Carets? Why?

BTW. As far as I remember the last time we face carets we decided to keep tildas until 0.12 release because of support of 2 latest stable versions of node. ;-)

mrjoelkemp:
Nice job @markelog. Thanks for contributing ;)

PR #1154, opened 2015-03-09
Presets: Add validate indentation rule for Yandex

gunnarlium:
The Yandex guide specifies 4 spaces for indentation, but the
`validateIndentation` rule isn't listed in the preset.

I had to add a function call to the default switch path in order for
the tests to pass. Not sure if this is caused by a bug in node-jscs
or not.

### DIFF
--- a/presets/yandex.json
+++ b/presets/yandex.json
@@ -95,5 +95,6 @@
     "disallowMultipleLineBreaks": true,
     "disallowMixedSpacesAndTabs": true,
     "disallowTrailingWhitespace": true,
+    "validateIndentation": 4,
     "validateLineBreaks": "LF"
 }
--- a/test/data/options/preset/yandex.js
+++ b/test/data/options/preset/yandex.js
@@ -95,6 +95,7 @@ switch (value) {
     default:
         // ...
         // no break keyword on the last case
+        doSomething();
 }
 
 // Ternary operators.


zxqfox:
Can you point me to Yandex Guide? Link to document would be great.

/cc @markelog

gunnarlium:
https://github.com/yandex/codestyle/blob/master/javascript.md#general

> One level of indentation is achieved with 4 space characters.

markelog:
@tarmolov @ikokostya could any of you confirm this please?

tarmolov:
It's a good intention to add this rule to the preset but It's a little bit dangerous and can break some projects.

Let me explain why.

It's pretty common in Yandex to write a module declaration as below:
```javascript
modules.define(
    'module1',
    [   
        'module2'
    ],  
    function (
        provide,
        module2
    ) { 
    // ...
    provide(module1);
});
```
As you can see we pruned the extra indent on purpose because it increased the line total indent.

If we apply the new rule, we should add an extra indent in the code.
```javascript
modules.define(
    'module1',
    [
        'module2'
    ],
    function (
        provide,
        module2
    ) {
        var module1 = module2;                                                                                                                                
        provide(module1);
    });
```

For example, `bem-core` library is [in danger](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.vanilla.js#L26) which is used [preset yandex](https://github.com/bem/jscs-bem/blob/master/presets/bem.json#L2) with a couple of overrides.

@ikokostya what do you think?

ikokostya:
I think it's :ok: . Actually we don't follow our codestyle in this place. We can override this rule or rethink style for modules.

ikokostya:
As i can see in http://jscs.info/rules.html we can't disable this rule. @markelog is it right?

zxqfox:
@ikokostya It's a bug if you can't.

All rules can be disabled by nullifying:
```json
{
  "preset": "yandex",
  "validateIdentation": null
}
```

mikesherov:
@tarmolov not true. validateIndentation has special logic to check for the presence of modules, and doesn't require an indent inside the module. ;-)

If that's not working, LMK.

mikesherov:
Ah, I see you're using `modules.define`. It only looks for `define` or `require`. I suppose I can update the detection.

zxqfox:
@mikesherov :+1: would be great if you also add `modules.define` and `modules.require` or option to define these in preset. ;-)

mikesherov:
@zxqfox file a new issue please.

markelog:
@ikokostya, @zxqfox is right, you can disable any rule.

Okay, it seems we can land this.

mrjoelkemp:
Thanks for contributing @gunnarlium.

PR #1162, opened 2015-03-11
requireLineBreakAfterVariableAssignment: allow exception for init part o...

### LINKED ISSUE #1157: requireLineBreakAfterVariableAssignment should have exception for declarations in for loops
Currently the following code causes requireLineBreakAfterVariableAssignment to complain:
```
for (var i = 0, length = myArray.length; i < length; i++) {
...
}
```
I think it would make more sense if this style rule ignored variable declarations in the "init" portion of the for loop.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9419489-requirelinebreakaftervariableassignment-should-have-exception-for-declarations-in-for-loops?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
...f for loop

Fixes #1157.

Returns if `node.parentNode.type === 'ForStatement'`

### DIFF
--- a/lib/rules/require-line-break-after-variable-assignment.js
+++ b/lib/rules/require-line-break-after-variable-assignment.js
@@ -61,6 +61,12 @@ module.exports.prototype = {
     check: function(file, errors) {
         var lastDeclaration;
         file.iterateNodesByType('VariableDeclaration', function(node) {
+            if (node.parentNode.type === 'ForStatement' ||
+                node.parentNode.type === 'ForInStatement' ||
+                node.parentNode.type === 'ForOfStatement') {
+                return;
+            }
+
             for (var i = 0; i < node.declarations.length; i++) {
                 var thisDeclaration = node.declarations[i];
                 if (thisDeclaration.parentNode.kind === 'var') {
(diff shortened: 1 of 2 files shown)


hzoo:
Oh looks like I need `ForStatement, ForInStatement, and ForOfStatement`?

lib/rules/require-line-break-after-variable-assignment.js
```diff
@@ -61,6 +61,12 @@ module.exports.prototype = {
     check: function(file, errors) {
         var lastDeclaration;
         file.iterateNodesByType('VariableDeclaration', function(node) {
+            if (node.parentNode.type === 'ForStatement' ||
```
zxqfox: I hope each `VariableDeclaration` node has `parentNode` property :smiley_cat:

mikesherov: Unless it's the `Program` node, every single one does.

hzoo: Should I change to ` if (node.parentNode.type &&  node.parentNode.type === ...)`?

zxqfox: No, no reason. If `node.parentNode.type` is empty — block will be skipped, if it contains wrong value — it will be also skipped, all is fine as is.

mrjoelkemp:
Nice work @hzoo.

PR #1172, opened 2015-03-16
requireCapitalizedComments: automatically except `jscs` comments

### LINKED ISSUE #1168: requireCapitalizedComments: add jscs as an exclusion
Using inline comments to disable rules will cause this rule to complain. So we either add `jscs` as a built-in exception or change inline comments to support `Jscs:`. The former seems most sensible.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9574546-requirecapitalizedcomments-add-jscs-as-an-exclusion?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jamesreggio:
Comments beginning with `jscs` are used to selectively disable and enable rules within a file. This directive is case-sensitive and lowercase, so we except it from this rule.

Fixes #1168

### DIFF
--- a/lib/rules/require-capitalized-comments.js
+++ b/lib/rules/require-capitalized-comments.js
@@ -94,7 +94,11 @@ module.exports = function() {};
 
 module.exports.prototype = {
     configure: function(options) {
-        this._exceptions = {};
+        // except comments that begin with `jscs`, since these are used to
+        // selectively enable/disable rules within a file
+        this._exceptions = {
+            'jscs': true
+        };
 
         var optionName = this.getOptionName();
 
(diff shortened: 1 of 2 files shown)


jamesreggio:
This may not be the 'right fix', but the issue filed and +1'ed by other folks makes me think that it's reasonable.

mrjoelkemp:
Any reason why you think this isn't the right fix? Looks good to me. Landed. 

Thanks for contributing @jamesreggio!

jamesreggio:
@mrjoelkemp, the rule requires explicit exceptions other styles of style-checker directives (e.g., `/* jshint: ...`), so I wasn't 100% sure that a built-in exception for JSCS would be appropriate. That said, this is the JSCS project (not JSHint), so it works for me.

Thanks for the quick review + merge.

PR #1173, opened 2015-03-16
Fix types and descriptions in documentation

MunGell:
Add missing types, small changes in descriptions


mrjoelkemp:
Thanks for contributing @MunGell!

Issue #480, opened 2014-06-29
Declaratively specify rule option values


mikesherov:
Instead of manually asserting in the `configure` function of a rule, we should be able to specify what values are accepted declaratively. This will also allow us to automate the creation of sample .jscsrc files. In order to do this, we need:

1. A way to set option values.
1. A way to verify provided values against use supplies values.
1. A way to automate error messages when a supplied value doesn't match.
1. A way to retrieve the possible values for a given rule.

markelog:
First and second is https://github.com/jscs-dev/node-jscs/issues/895, consider it as sub-task to this issue

TheSavior:
If we are considering auto generating doc blocks as well to be consistent with the specified options, how do we plan on handling good/bad examples? Seems unlikely that we'd be successful at choosing good options for those in code.

PR #1174, opened 2015-03-16
Change 'Value' to 'Values' to be consistent

MunGell:
Not as important, so moved these changes into separate PR. Just to be consistent – seems like most of the rules have "Values" instead of "Value".

### DIFF
--- a/lib/rules/disallow-anonymous-functions.js
+++ b/lib/rules/disallow-anonymous-functions.js
@@ -3,7 +3,7 @@
  *
  * Type: `Boolean`
  *
- * Values: `true`
+ * Value: `true`
  *
  * #### Example
  *
--- a/lib/rules/disallow-comma-before-line-break.js
+++ b/lib/rules/disallow-comma-before-line-break.js
@@ -3,7 +3,7 @@
  *
  * Type: `Boolean`
  *
- * Values: `true`
+ * Value: `true`
  *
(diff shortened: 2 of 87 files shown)


TheSavior:
It seems like It should be `value` if there is only one choice, and `values` if there is more.

mrjoelkemp:
Agreed with @TheSavior here. Would you mind making that change where appropriate @MunGell? That would be a great win for the docs.

TheSavior:
And as a side note, is this something that we could automate ourselves for the future? Anything that can be enforced by linters or tests is better than manual checking and enforcing down the line.

mrjoelkemp:
Seems like it would depend on #480.

TheSavior:
Not necessarily. I could see us trying to generate those doc blocks automatically, but I could also imagine us trying to figure out what a custom jscs rule (or another system if there was a better way) to validate the contents of docblocks.

MunGell:
I was thinking about changing it to singular form where necessary, but it is looking like we are using just one form in all the rules. I can change it, should I do the same for "Type"/"Types"?

mrjoelkemp:
Yes please. The docs should be grammatically correct.

MunGell:
Hey guys,

This should be corrected now, please take a look at the diff.

mrjoelkemp:
@MunGell Great work here! Thanks for contributing. Landed!

MunGell:
You are welcome!

PR #1184, opened 2015-03-23
Token Assert: normalize whiteSpace assertions to match line assertions

mikesherov:


indexzero:
:+1: +1 lgtm

lib/token-assert.js
```diff
@@ -77,18 +37,79 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
...
+        var msgPostfix = token.value + ' and ' + nextToken.value;
+
+        if (!options.message) {
+            if (exactly === 0) {
+                // support noWhitespaceBetween
+                options.message = 'Unexpected whitespace between ' + msgPostfix;
+            } else if (exactly !== undefined) {
+                // support whitespaceBetween (spaces option)
+                options.message = spaceCount + ' spaces required between ' + msgPostfix;
+            } else if (atLeast === 1 && atMost === undefined) {
```
zxqfox: Maybe switch-case here?
Is it just a BC or what about `atLeast === undefined && atMost === 1` ?
Probably better to use `typeof variable === 'undefined'`.

lib/token-assert.js
```diff
@@ -315,6 +319,43 @@ TokenAssert.prototype.linesBetween = function(options) {
...
+ */
+TokenAssert.prototype._validateOptions = function(options) {
+    var token = options.token;
+    var nextToken = options.nextToken;
+    var atLeast = options.atLeast;
+    var atMost = options.atMost;
+    var exactly = options.exactly;
+
+    if (token === nextToken) {
+        throw new Error('You cannot specify the same token as both token and nextToken');
```
zxqfox: space between `can not`?

lib/token-assert.js
```diff
@@ -77,18 +37,79 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
...
-            message: options.message || 'Unexpected whitespace between ' + token.value + ' and ' + nextToken.value,
+            message: options.message,
             line: token.loc.end.line,
             column: token.loc.end.column,
-            fixed: true
+            fixed: fixed
         });
+    }.bind(this);
+
+    var spacesBetween = Math.abs(nextToken.range[0] - token.range[1]);
```
zxqfox: math.abs? not sure that it's right.

zxqfox:
:+1: Yeah, I like it too ;-)

lib/token-assert.js
mikesherov: Just about BC, this code will go away in 2.0

lib/token-assert.js
mikesherov: This is the same as how `linesBetween` is implemented. It allows you to reverse the tokens if needed.

lib/token-assert.js
zxqfox: Nvm then. I'd also add here a comment about it. ;-)

lib/token-assert.js
mikesherov: existing language. Don't want to change it just yet.

lib/token-assert.js
zxqfox: Ah, nice! :+1:

lib/token-assert.js
```diff
@@ -77,18 +37,79 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
...
     var nextToken = options.nextToken;
-    if (nextToken.range[0] !== token.range[1] &&
-        (options.disallowNewLine || token.loc.end.line === nextToken.loc.start.line)
-    ) {
-        nextToken.whitespaceBefore = '';
+    var atLeast = options.atLeast;
+    var atMost = options.atMost;
+    var exactly = options.exactly;
+
+    if (!token || !nextToken) {
```
TheSavior: I know this is existing logic, but I've wondered before if these should be returning out like they are, or throwing an exception.

lib/token-assert.js
```diff
@@ -77,18 +37,79 @@ TokenAssert.prototype.whitespaceBetween = function(options) {
...
+
+    if (!options.disallowNewLine && token.loc.end.line !== nextToken.loc.start.line) {
+        return;
+    }
+
+    // Only attempt to remove or add lines if there are no comments between the two nodes
+    // as this prevents accidentally moving a valid token onto a line comment ed line
+    var fixed = this._file.getNextToken(options.token, {includeComments: true}) === nextToken;
+
+    var emitError = function(countPrefix, spaceCount) {
```
TheSavior: `var emiterror =` or `function emitError(...`?

lib/token-assert.js
```diff
@@ -239,42 +260,25 @@ TokenAssert.prototype.differentLine = function(options) {
...
 
     if (!token || !nextToken) {
         return;
     }
 
-    if (token === nextToken) {
-        throw new Error('You cannot specify the same token as both token and nextToken');
-    }
-
-    var atLeast = options.atLeast;
```
TheSavior: It looks like these are still used down on 312, did some other logic change that I'm not noticing that makes this work?

test/specs/token-assert.js
```diff
@@ -66,10 +66,12 @@ describe('modules/token-assert', function() {
                 nextToken: tokens[1]
             });
 
-            assert(!onError.called);
+            assert(onError.notCalled);
         });
+    });
 
-        it('should trigger error on invalid space count between tokens', function() {
+    describe('spacesBetween', function() {
```
TheSavior: It seems like spacesBetween and linesBetween should have extremely similar tests. Perhaps not in logic, but definitely in structure. For example. Verify error messages, verify exactly logic, verify atLeast logic, etc. I think the structure of the tests should be made to match. It looks like this is missing tests for missing other options that linesBetween has.

test/specs/token-assert.js
```diff
@@ -77,70 +79,174 @@ describe('modules/token-assert', function() {
...
+                    nextToken: tokens[1],
+                    exactly: 2,
+                    message: 'Custom message'
+                });
+
+                assert.equal(onError.getCall(0).args[0].message, 'Custom message');
             });
 
-            assert(!onError.called);
+            it('should error, but not fix, when a comment exists between the two tokens', function() {
```
TheSavior: Should there be a test for `should error and fix whitespace between tokens without a comment`

TheSavior: `linesBetween` has checks that the right whitespaceBefore was added for the exactly, atLeast, and atMost cases. We probably want those here too.

mikesherov: I suppose I could beef these up.

lib/token-assert.js
mikesherov: line 268-270.

lib/token-assert.js
mikesherov: Matter of personal preference.

lib/token-assert.js
mikesherov: This is a hold over from the fact that you can pass in nulls, and it shouldn't cause an error. This is copied from `linesBetween` and I'll consolidate them further in a future followup but won't change behavior for BC.

lib/token-assert.js
TheSavior: :+1:

markelog:
Since there is only concerns with the tests we can improve that later.

markelog:
I only fear that `spacesBetween ` might become a some sort of <a href="http://en.wikipedia.org/wiki/God_object">god function</a>, since now it's used by a lot of methods and its cyclomatic complexity is pretty high.

TheSavior:
Do you feel differently about linesBetween?

zxqfox:
As I see it's just because of BC, I think we shouldn't worry about it because public API is pretty stable and shouldn't drastically change in future.
And because of that we can decrease this complexity by several waves of refactoring.

We just need it to finish giant leap for mankind ;-)

mikesherov:
It's complexity is only high because of BC and having tons of varying error messages.

It will get better in 2.0 and I wouldn't worry about it.

TheSavior:
Just out of curiosity, I've been seeing it a bunch. What is BC in this context?

zxqfox:
Back compatibility ;-)

PR #1189, opened 2015-03-25
Website: Added link to the preset option from the list of presets

### LINKED ISSUE #1185: disallowSpaceBeforeBinaryOperators doesn't alow , but it is used in Crockford preset
The Crockford preset contains the lines

```json
"disallowSpaceBeforeBinaryOperators": [
        ","
    ],
```

(see https://github.com/jscs-dev/node-jscs/blob/master/presets/crockford.json#L44-L46)
which causes this error

> The value of property "disallowSpaceBeforeBinaryOperators" was "," but must be one of the following values: "?", "+", "-", "/", "*", "=", "==", "===", "!=", "!==", ">", ">=", "<", "<=", null

This is slightly confusing as the listed operators are not "binary" as defined in the ECMAScript spec. (eg &&). Even more confusing the documentation includes the comma operator "'," in the example (see http://jscs.info/rules.html#disallowspaceafterbinaryoperators).

I tested by copying the raw preset contents to a .jscsrc file used by WebEssentials in VisualStudio

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9921368-disallowspacebeforebinaryoperators-doesn-t-alow-but-it-is-used-in-crockford-preset?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SteveALee:
I could not test the internal link as the heading id seems to be mangled for being in the preview and also I do not know if the HTML converter will process correctly.

Closes gh-issues/1185
Fixes #1185

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -2,6 +2,8 @@
 
 ## Presets
 
+Note: the easiest way to use a preset is with the [preset](#preset) option described below. This will use the correct version of the preset for the JSCS version being used. Not all presets may be available.
+
  * [Airbnb](https://github.com/jscs-dev/node-jscs/blob/master/presets/airbnb.json) — https://github.com/airbnb/javascript
  * [Crockford](https://github.com/jscs-dev/node-jscs/blob/master/presets/crockford.json) — http://javascript.crockford.com/code.html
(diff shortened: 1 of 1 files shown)


zxqfox:
:+1: As a hot fix.

But I believe we should rework this document ;-)

PR #1191, opened 2015-03-25
requireLineBreakAfterVariableAssignment: add check for let and const

### LINKED ISSUE #1190: requireLineBreakAfterVariableAssignment doesn't handle 'let' or 'const'
When running with `esnext` set to true, `let a = 'b', c = 'c';` and `const a = 'b', c = 'c';` are not caught, while `var a = 'b', c = 'c';` is correctly caught.

Looking at the code for the rule, it is explicitly checking for `var`:
```
if (thisDeclaration.parentNode.kind === 'var') { ... }
```
Seems like `let` and `const` need to be added to this if statement.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10005589-requirelinebreakaftervariableassignment-doesn-t-handle-let-or-const?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1190. @blazarus

### DIFF
--- a/lib/rules/require-line-break-after-variable-assignment.js
+++ b/lib/rules/require-line-break-after-variable-assignment.js
@@ -65,7 +65,9 @@ module.exports.prototype = {
 
             for (var i = 0; i < node.declarations.length; i++) {
                 var thisDeclaration = node.declarations[i];
-                if (thisDeclaration.parentNode.kind === 'var') {
+                if (thisDeclaration.parentNode.kind === 'var' ||
+                    thisDeclaration.parentNode.kind === 'let' ||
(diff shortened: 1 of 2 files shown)


zxqfox:
Lgtm

mrjoelkemp:
@markelog This is technically a bug fix (though unrelated to the new stuff in 1.12). Are we cool with merging it?

PR #1202, opened 2015-03-27
[WIP] autoformatting for `validateQuoteMark`

indexzero:
:warning: this is my first try learning the new autoformatting APIs. Not ready for merge, looking for guidance. 

@mikesherov the main change I wanted to run by you was the passing of a new `fixed` parameter to `errors.add`. Seems like `errors.add` would be better suited taking just an `errorInfo` object like `_addError`, but I didn't want to be too presumptuous in my refactor. 

Let me know if I'm headed in the right direction and I'll finish off tests, etc. Autoformatting is so cool!

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -31,7 +31,7 @@ Errors.prototype = {
      * @param {Number|Object} line
      * @param {Number} [column] optional if line is an object
      */
-    add: function(message, line, column) {
+    add: function(message, line, column, fixed) {
         if (typeof line === 'object') {
             column = line.column;
             line = line.line;
@@ -46,7 +46,8 @@ Errors.prototype = {
         this._addError({
             message: message,
             line: line,
-            column: column
+            column: column,
+            fixed: fixed
         });
     },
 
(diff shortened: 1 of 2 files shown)


zxqfox:
Looks good ;-)

mikesherov:
@indexzero, its in the right direction. The addition to `errors.add` is fine. 

The patch itself is good but not without bugs. This would break on `var a = "an \"escaped\" string";` no? Please add some tests for single, double, single with escape string, double with escape string, and a test for errors.add.

Thanks again @indexzero !

millermedeiros:
The test cases and implementation might be useful: https://github.com/millermedeiros/esformatter-quotes

I'm almost sure there was a npm package to change string quotes but couldn't find it the day I started esformatter-quotes... Would be good if we could all reuse some external/reliable lib.

mikesherov:
@millermedeiros sure. That makes sense. Thanks for pointing out the package.

TheSavior:
I really want this, so I'm going to try to jump in and help move this forward.

indexzero:
@TheSavior no complaints from me. I am deep in the last thralls of organizing [EmpireJS](http://2015.empirejs.org) so my OSS schedule is swamped.

TheSavior:
@markelog @mikesherov: Are we guaranteed that if there are no errors reported by `checkString` then `fixString` will never change anything?

Issue #1205, opened 2015-03-27
New Rule: disallowNotOperatorsInConditionals


SimenB:
Sorta confusing title. Also, sorry if this is already a rule.

I'd like the following code to to throw
```javascript
if (!clause) {
    // Do something crazy
} else {
    // Do something really crazy
}
```
The correct way would be
```javascript
if (clause) {
    // Do something really crazy
} else {
    // Do something crazy
}
```

If no else is present, no error.

Does this make sense?

TheSavior:
When I was at Google, this was ingrained into my coding habits. I would love a rule for this.

hzoo:
So this would just check for the presence of the not operator in an if with an else?

SimenB:
Yeah, either just negating, or ``==!``.

hzoo:
And only with a single boolean? What if it's (a &&!b)?

SimenB:
In my mind, if you have a single wrapped expression, that you then negate, it should throw.

So from your last 2 examples, the first is fine, the second should throw.

If you think of it as if you removed the negation, and could then just switch the bodies of the ``if`` and the ``else`` and have the same result, then it should throw.

hzoo:
Ok cool. I'm not sure about a good rule name... `disallowNegation...` or `disallowFalsyCheckInIfStatement`..

Unrelated: It might be interesting to check if a expression can be simplified `!(!a || !b)` -> `a && b`

TheSavior:
@mkanat: IIRC, you had a name for this.

SimenB:
(related to boolean-expressions)

What about a rule about maximum number of expressions? Like if you have maximum 2, then ``(a || b)`` is fine, but ``(a || b || c)`` would throw, and you should split it and extract it to a variable or something.
Or what about missing clarifying parentheses? e.g. ``(a && b || c && b)`` can be confusing. It's much easier to read and reason about if it's written as ``((a && b) || (c && d))``

TheSavior:
I could imagine a rule that requires parenthesis for any expression with more than one operator of different importances. That would encompass `var a = 2 + 4 / 8 > 6 && 2 - 4 * 2 < 3` as well as requiring parens for boolean conditionals. Can you make a new issue describing that?

hzoo:
Can work on a PR for this.

TheSavior:
`disallowInvertedConditionalsWithElse`?

TheSavior:
I would imagine this should be okay:
[...]

PR #1210, opened 2015-03-27
New Rule: disallowNotOperatorsInConditionals

hzoo:
WIP. for #1205 

~~Need to figure out a better name~~, also other cases?

~~Looks like I would need to add a check for `!=` (inequality) and change the naming for `!==` to be `strict inequality operator`~~

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -691,6 +691,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-identifier-names'));
 
     this.registerRule(require('../rules/maximum-number-of-lines'));
+
+    this.registerRule(require('../rules/disallow-not-operators-in-conditionals'));
 };
 
 /**
--- a/lib/rules/disallow-not-operators-in-conditionals.js
+++ b/lib/rules/disallow-not-operators-in-conditionals.js
@@ -1,0 +1,106 @@
+/**
+ * Disallows the not, not equals, and strict not equals operators in conditionals.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowNotOperatorsInConditionals": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * if (clause) {
+ *     // Do something really crazy
+ * } else {
+ *     // Do something crazy
+ * }
+ *
+ * if (a == 1) {
+ *     // Do something really crazy
+ * } else {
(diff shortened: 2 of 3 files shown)


SimenB:
@hzoo Could you cover ternaries as well?

Error
``date = !_.isEmpty(date) ? moment(date, 'DD-MM-YYYY') : null;``

Correct
``date = _.isEmpty(date) ? null : moment(date, 'DD-MM-YYYY');``

hzoo:
Ok with that the rule name/docs should change as well. `disallowNotOperatorsInConditionals` to cover both?

SimenB:
Yeah, sounds good :+1:

zxqfox:
Totally agreed about conditionals.

But are you sure about plural in naming?
`disallowNotOperatorsInConditionals` vs `disallowNotOperatorInConditional`

hzoo:
As of now, we will be checking for `!==, !=, !a`
in `if()` and ternary `(a ? b : c)`

markelog:
Still a WIP?

hzoo:
Yeah I just have to change the rule name to `disallowNotOperatorsInConditionals` and add a check for ternary statement and `!=` which I can do soon.

hzoo:
Updated the rule name to `disallowNotOperatorsInConditionals`, fixed examples/tests - added check for `!=` and ternaries.

markelog:
I would ask for opposite rule, but really, who would need that?

zxqfox:
`validate*`?

markelog:
Yeah, was thinking the same thing, but `validateNotOperatorsInConditionals` sounds peculiar to me.

Ideas?

zxqfox:
`validateTestInConditionals`?

markelog:
@hzoo ?

SimenB:
validateNegationInConditionals? validateNoNegationInConditionals?

mikesherov:
We have disallow/require rules without opposites. No need to force it to be named "validate".

markelog:
I think it would be a good naming convention, we can start enforcing it now and continue to use it in new major version

hzoo:
I still like `disallowNotOperatorsInConditionals`

markelog:
Me too, but see above, i'd like establish convention for these kind of rules.

zxqfox:
@markelog What you think about leaving it as is for now and try to rethink it in the way to 2.0? I still want to revive your issue with new shiny configurations.

markelog:
Works for me, will land this soon

hzoo:
Oh I didn't add it to the commit - #1205.

PR #1222, opened 2015-03-31
New Rule: requireAlignedFunctionParameters

### LINKED ISSUE #1164: Multi-line function declaration parameters
Jetbrians' IDE have the following options for function declaration parameters:
- align when multi-line
- new line after __(__
- place __)__  on new line

When all set to true it  gives me the formatting I'm looking for:
```javascript
function someFunctionDeclaration(
  aBunch,
  ofArguments,
  thatDontGo,
  onOneLine
) {
}
```
Is there a similar rule for jscs to format my code this way?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9476885-multi-line-function-declaration-parameters?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bramcordie:
fixed #1164

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -593,6 +593,7 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-quoted-keys-in-objects'));
     this.registerRule(require('../rules/disallow-dangling-underscores'));
     this.registerRule(require('../rules/require-aligned-object-values'));
+    this.registerRule(require('../rules/require-aligned-function-parameters'));
 
     this.registerRule(require('../rules/disallow-padding-newlines-after-blocks'));
     this.registerRule(require('../rules/require-padding-newlines-after-blocks'));
--- a/lib/rules/require-aligned-function-parameters.js
+++ b/lib/rules/require-aligned-function-parameters.js
@@ -94,7 +94,6 @@ module.exports.prototype = {
                 maxParamStartPos = Math.max(maxParamStartPos, parameter.loc.start.column);
             });
 
-
             // make sure all parameters are lined up
             node.params.forEach(function(parameter) {
                 if (parameter.loc.start.column !== maxParamStartPos) {
--- a/lib/rules/require-aligned-function-parameters.js
+++ b/lib/rules/require-aligned-function-parameters.js
@@ -1,0 +1,121 @@
+/**
+ * Requires proper alignment of function parameters.
+ *
+ * Type: `Object` or `Boolean`
+ *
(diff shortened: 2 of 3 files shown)


mikesherov:
Thanks for contributing! Please use token-assert to match the code that is available in `requireAlignedObjectValues`.

bramcordie:
I didn't find a token-assert to check if all the parameters are aligned. Did I miss it? Should I add my own?

hzoo:
@bramcordie you can refer to what https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/require-aligned-object-values.js has. It is using `errors.assert.spacesBetween` now.

bramcordie:
I found a suitable token-assert (differentLine) to check line breaks as you can see in my last commit.

Again, correct me if I'm wrong but I don't think that the __spaceBetween__ assert is suitable here.
_requireAlignedObjectValues_ only checks the space between the tokens of the object name and the colon on the same line. It doesn't work for parameters because the tokens are not on the same line and the first one isn't even a parameter token.
I hope this clears things up and helps figure out if-a/which token assert is appropriate to use here.

hzoo:
I see.. are you talking about the indentation of all the parameters?

It doesn't look like `requireAlignedObjectValues` is checking for that. So maybe this rule doesn't need to check it since it should be covered by `validateIndentation`. If so, `differentLine` is good enough.

If not, you could add that in by using `errors.assert.indentation` like in that rule https://github.com/jscs-dev/node-jscs/blob/f55bbae8bfbd8d57a4f756193d824e6f398ff7a7/lib/rules/validate-indentation.js#L271?

bramcordie:
I think this rule should just check alignment. When the `lineBreakAfterOpeningBraces` option is not turned on you can leave the first parameter on the same line as the start of the function. This way it's not indented.
```javascript
function (thisIs,
          theLongestList,
          ofParametersEverWritten) {}
```

`validateIndentation` does not check the indentation of function parameters. Which is a good because else the use-case above would not be possible.

If I can access the `validateIndentation` value I could enforce it when __all__ parameters are indented on new lines?

markelog:
So everybody is fine with merging this?

zxqfox:
I like this at all. There is something I'd fix but I can't get what is that actually.

So LGTM.

markelog:
Since this rule can't have the opposite one, it's renamed to `validateAlignedFunctionParameters`.
Thank you.

zxqfox:
:+1:

PR #1233, opened 2015-04-02
true/false missing in ES3_KEYWORDS

1999:
http://www.w3schools.com/js/js_reserved.asp
true/false are also ES3 reserved words.

```javascript
DOM.decl('overlay', {
    onSetMod: {
        visible: {
            'true': function () {
                    // ...
                }
            }
        }
}
```

`disallowQuotedKeysInObjects": "allButReserved"` currently doesn't respect true/false object keys in quotes.


markelog:
Thank you, although,  technically speaking , `true/false/null` are reserved words but not keywords, whereas this list should contain only keywords.

btw, it's generally considered a bad idea to give proof links on http://www.w3schools.com

1999:
Okay, so what should we use then to prevent jscs from throwing errors with this code? When we use true without quotes, IE8 doesn't work at all: https://jing.yandex-team.ru/files/dsorin/2015-04-03_1141.png

Btw what's so bad about using w3schools? I could've put a link to MDN but why is it a bad idea?

markelog:
Situation around the http://www.w3schools.com has improved, but it still has bad reputation, for example this how it was couple years ago -https://web.archive.org/web/20130115033517/http://w3fools.com/

It's usually better to give link on the spec itself.

PR #1236, opened 2015-04-05
requirePaddingNewLineAfterVariableDeclaration: fix ES6 false positives

MartinKolarik:


zxqfox:
Looks :+1:

hzoo:
Looks good - similar to #1191. Thought I didn't see any others but maybe we should look for `=== 'var'`.

mikesherov:
:+1:

markelog:
Closing it in favour of  #1191, nevertheless, thank you @MartinKolarik.

hzoo:
Oh @markelog I think we need to add this one since it's the same fix applied to another rule. #1191 was for `requireLineBreakAfterVariableAssignment` and this seems to be the same issue for `requirePaddingNewLineAfterVariableDeclaration`?

Looks like the only other rule it would apply to is `requireMultipleVarDecl`.

markelog:
Oh, shoot #1191 and this one are two different fixes :-), reopening

yaycmyk:
You should also add a test for mixed types.

const, then let for example

markelog:
@yannickcr it's done in https://github.com/jscs-dev/node-jscs/pull/1245/files, thanks for keeping an eye on it.

markelog:
Thank you

Issue #1237, opened 2015-04-05
JSCS with fat arrow functions and airbnb style and JSX, ES6?


justin808:
Has anybody gotten jscs to work with with fat arrow functions and airbnb? I get this message:
```
Missing newline after closing curly brace at ./assets/javascripts/CommentBox.jsx :
    44 |      var newComments = React.addons.update(oldComments, {$push: [newComment]});
    45 |      this.setState({ajaxSending: false, data: newComments, formData: this.emptyFormData});
    46 |    }, (xhr, status, err) => {
-------------^
    47 |      this.logError(xhr, status, err);
    48 |      this.setState({ajaxSending: false});
```
The problem is that jscs does not understand that the braces are within a function call.

Here's the full code for that method:
```
  handleCommentSubmit: function() {
    // `setState` accepts a callback. To avoid (improbable) race condition,
    // `we'll send the ajax request right after we optimistically set the new
    // `state.
    var comment = this.state.formData;
    this.setState({ajaxSending: true});
    $.ajax({
      url: this.props.url,
      dataType: 'json',
      type: 'POST',
      data: {comment: comment}
    }).then(data => { // Server returns all data
      const newComment = data;
      const oldComments = this.state.data;
      var newComments = React.addons.update(oldComments, {$push: [newComment]});
      this.setState({ajaxSending: false, data: newComments, formData: this.emptyFormData});
    }, (xhr, status, err) => {
      this.logError(xhr, status, err);
      this.setState({ajaxSending: false});
    });
  },
```

My jscs file is:
```
{
  "preset": "airbnb",
  "fileExtensions": [".js", ".jsx"],
  "excludeFiles": ["build/**", "node_modules/**"],
  "esprima": "esprima-fb"
}
```

I checked in the run here:
https://github.com/justin808/react-webpack-rails-tutorial/pull/26

justin808:
To run the example, sync up the branch `integrate-jscs` and run:

    (cd client && npm run jscs .)

I get 2 failures:

```
Missing newline after closing curly brace at ./assets/javascripts/CommentBox.jsx :
    44 |      var newComments = React.addons.update(oldComments, {$push: [newComment]});
    45 |      this.setState({ajaxSending: false, data: newComments, formData: this.emptyFormData});
    46 |    }, (xhr, status, err) => {
-------------^
    47 |      this.logError(xhr, status, err);
    48 |      this.setState({ajaxSending: false});

[...]

PR #1239, opened 2015-04-06
Fixing requirePaddingNewLinesAfterBlocks to no longer fail on argument lists

### LINKED ISSUE #1232: requirePaddingNewLinesAfterBlocks seems overly strict
```js
callingAFunction(
  with,
  another,
  function() {
    // ...
  }
);

// more code
```

This results in `Missing newline after closing curly brace`. IMO the function call is a block in itself, so the newline should be required after it. Not after the nested anonymous function.

But JSCS prefers this – which should IMHO be wrong:

```js
callingAFunction(
  with,
  another,
  function() {
    // ...
  }

);
// more code
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10348543-requirepaddingnewlinesafterblocks-seems-overly-strict?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
Fixes #1232

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -72,10 +72,30 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(value) {
-        assert(
-            value === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        this._allowInCallExpressions = false;
+        this._allowInArrayExpressions = false;
+
+        if (typeof value === 'object') {
+            assert(Array.isArray(value.allExcept), this.getOptionName() + ' option requires "allExcept" ' +
+                'to be an array');
+            assert(value.allExcept.length > 0, this.getOptionName() + ' option requires "allExcept" ' +
+                'to have at least one item or be set to `true`');
+
+            value.allExcept.forEach((function(except) {
+                if (except === 'inCallExpressions') {
+                    this._allowInCallExpressions = true;
+                } else if (except === 'inArrayExpressions') {
+                    this._allowInArrayExpressions = true;
+                } else {
+                    assert(false, this.getOptionName() + ' option requires "allExcept" to only have ' +
+                        '`inCallExpressions` or `inArrayExpressions`');
+                }
+            }).bind(this));
+        } else {
+            assert(value === true,
+                this.getOptionName() + ' option requires true value or object'
+            );
+        }
     },
 
     getOptionName: function() {
@@ -83,12 +103,29 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-        file.iterateNodesByType('BlockStatement', function(node) {
+        file.iterateNodesByType('BlockStatement', (function(node) {
 
             var closingBracket = file.getLastNodeToken(node);
             var parentNode = node.parentNode;
 
             var nextToken = file.getNextToken(closingBracket);
+
+            var block = node.parentNode;
+            if (block && block.parentNode) {
+                var parent = block.parentNode;
+
+                if (this._allowInCallExpressions &&
+                    parent.type === 'CallExpression' &&
+                    parent.arguments.indexOf(block) !== -1) {
+                    return;
+                }
+
+                if (this._allowInArrayExpressions &&
+                    parent.type === 'ArrayExpression' &&
+                    parent.elements.indexOf(block) !== -1) {
+                    return;
+                }
+            }
 
             while (nextToken.type !== 'EOF') {
                 var excludeValues = excludes[parentNode.type];
@@ -109,11 +146,11 @@ module.exports.prototype = {
                     token: closingBracket,
                     nextToken: nextToken,
                     atLeast: 2,
-                    message: 'Missing newline after closing curly brace'
+                    message: 'Missing newline after block'
                 });
 
                 return;
             }
-        });
+        }).bind(this));
     }
 };
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -9,12 +9,13 @@
  *      - `"allExcept"`: `Array`
  *          - `"inCallExpressions"` Blocks don't need a line of padding in argument lists
  *          - `"inArrayExpressions"` Blocks don't need a line of padding in arrays
+ *          - `"inProperties"` Blocks don't need a line of padding as object properties
  *
  * #### Example
  *
  * ```js
  * "requirePaddingNewLinesAfterBlocks": true
- * "requirePaddingNewLinesAfterBlocks": { "allExcept": ["inCallExpressions", "inArrayExpressions"] }
+ * "requirePaddingNewLinesAfterBlocks": { "allExcept": ["inCallExpressions", "inArrayExpressions", "inProperties"] }
  * ```
  *
  * ##### Valid
@@ -75,6 +76,16 @@
  *     }
  * ];
  * ```
+* ##### Valid for `{ "allExcept": "inProperties" }`
+ *
+ * ```js
+ * var foo = {
+ *     a: 2,
+ *     b: function() {
+ *     }
+ *     c: 3
+ * ];
+ * ```
  *
  * ##### Invalid
  *
@@ -86,14 +97,6 @@
  *         }
  *         continue;
  *     }
- *     var obj = {
- *         foo: function() {
- *             return 1;
- *         },
- *         bar: function() {
- *             return 2;
- *         }
- *     };
  * }
  * ```
  */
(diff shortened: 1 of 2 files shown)


lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -1,14 +1,20 @@
...
+ * - `Object`:
+ *      - `"allExcept"`: `Array`
+ *          - `"inCallExpressions"` Blocks don't need a line of padding in argument lists
+ *          - `"inArrayExpressions"` Blocks don't need a line of padding in arrays
  *
  * #### Example
  *
  * ```js
  * "requirePaddingNewLinesAfterBlocks": true
+ * "requirePaddingNewLinesAfterBlocks": { "allExcept": ["inCallExpressions", "inArrayExpressions"]}
```
zxqfox: Space before closing brace?

TheSavior: Hah. Good catch. Fixed.

markelog:
`inCallExpressions` and `inArrayExpressions` are specific options for this rule only, i wouldn't want to invent any new names if there wouldn't be any other way to do this.

#1232 could be solved much more easily

TheSavior:
@markelog, if your PR passes the added tests in this PR I'm okay with your solution. From an initial glance, I am guessing that it won't.

TheSavior:
Also referencing #1237 here.

markelog:
See the latest changes/comments in https://github.com/jscs-dev/node-jscs/pull/1246

TheSavior:
It is the same issue with arrays though

```javascript
var a = [
2,
3,
function() {
} // this will expect a blank line after it.
]
```

TheSavior:
I guess I'm not really sure what the issue with the fix in this PR is. Is it that you don't want any exclude options at all, only `true` or `false`? Or is it that you think we should change the defaults?

markelog:
@TheSavior i'm talking about consistency - `allExcept` values, in other rules, either a token list or list of exceptions, here we trying to provide rule options through it. Inventing our own node types or token are not ideal.

Plus why would we need to complicate the rule, if we could do it much more simpler?

markelog:
Sorry @TheSavior, it sometimes awful not to land a nice piece of work :-(. I'm closing this in favour of #1246

That being said, i think we can add support for arrays though.

TheSavior:
I'm not so much concerned about having to throw code away, but mostly the support of arrays. I was the original author of the rule, and discovered when running this that it was breaking in a bunch of places in our codebase that it shouldn't have been. I repeatedly stated on both of our PRs, that this rule needed to pass the tests I added and for some reason arrays were punted as not important enough for the hotfix. 

Frankly, I still can't use this rule until the array fix is in there, and that is what makes me annoyed. Especially as the original author of the rule.

(Obviously I can use a custom rule in our project for the time being, but that is awfully disappointing).

mikesherov:
@markelog we should land support for arrays too.

markelog:
I think we should go further to be consistent there and ignore blocks in objects too, if there is need for that functionality we would need a different rule altogether.

mikesherov:
Agreed.

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -72,10 +113,30 @@ module.exports = function() {};
...
 
     configure: function(value) {
-        assert(
-            value === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        this._allowInCallExpressions = false;
+        this._allowInArrayExpressions = false;
+
+        if (typeof value === 'object') {
```
markelog: This whole block is very complicated, i guess it would be better to work on https://github.com/jscs-dev/node-jscs/issues/895 instead?

TheSavior:
Alright, I added the `isException` helper into this branch and added an option for object properties as mentioned in #1260. 

@markelog would you like to reopen this PR so it grabs the new commit, or should I create a new PR?

markelog:
Did you push these changes?

markelog:
I see falling tests in CI, could you check it out?

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -109,11 +197,11 @@ module.exports.prototype = {
...
                     nextToken: nextToken,
                     atLeast: 2,
-                    message: 'Missing newline after closing curly brace'
+                    message: 'Missing newline after block'
                 });
 
                 return;
             }
-        });
+        }).bind(this));
```
markelog: We usually create a variable for this

markelog:
How about opposite rule?

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -32,9 +39,54 @@
...
+ *     }
+ * ];
+ * ```
+* ##### Valid for `{ "allExcept": "inProperties" }`
+ *
+ * ```js
+ * var foo = {
+ *     a: 2,
+ *     b: function() {
+ *     }
```
markelog: comma

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -1,14 +1,21 @@
...
+ *      - `"allExcept"`: `Array`
+ *          - `"inCallExpressions"` Blocks don't need a line of padding in argument lists
+ *          - `"inArrayExpressions"` Blocks don't need a line of padding in arrays
+ *          - `"inProperties"` Blocks don't need a line of padding as object properties
  *
  * #### Example
  *
  * ```js
  * "requirePaddingNewLinesAfterBlocks": true
+ * "requirePaddingNewLinesAfterBlocks": { "allExcept": ["inCallExpressions", "inArrayExpressions", "inProperties"] }
```
markelog: Couple new lines would help here

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -32,9 +39,54 @@
...
+ *
+ * ```js
+ * var foo = [
+ *     2,
+ *     3,
+ *     function() {
+ *     }
+ * ];
+ * ```
+* ##### Valid for `{ "allExcept": "inProperties" }`
```
markelog: shouldn't this be an array?

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -32,9 +39,54 @@
...
  *         }
  *     };
+ *
+ *     func(
+ *          function() {
+ *          }
+ *
+ *     );
+ *
+ *     var a = [
```
markelog: Maybe more meaningful examples would fit better

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,15 +111,55 @@ var excludes = {
...
-        );
+        if (typeof value === 'object') {
+            assert(Array.isArray(value.allExcept), this.getOptionName() + ' option requires "allExcept" ' +
+                'to be an array');
+            assert(value.allExcept.length > 0, this.getOptionName() + ' option requires "allExcept" ' +
+                'to have at least one item or be set to `true`');
+
+            value.allExcept.forEach((function(except) {
+                if (except === 'inCallExpressions') {
+                    exceptions.CallExpression = true;
```
markelog: Nice solution

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,15 +111,55 @@ var excludes = {
...
 
 module.exports.prototype = {
 
     configure: function(value) {
-        assert(
-            value === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        if (typeof value === 'object') {
+            assert(Array.isArray(value.allExcept), this.getOptionName() + ' option requires "allExcept" ' +
```
markelog: You can call `getOptionName` once assigning it to variable

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,15 +111,55 @@ var excludes = {
...
+            value.allExcept.forEach((function(except) {
+                if (except === 'inCallExpressions') {
+                    exceptions.CallExpression = true;
+                } else if (except === 'inArrayExpressions') {
+                    exceptions.ArrayExpression = true;
+                } else if (except === 'inProperties') {
+                    exceptions.Property = true;
+                } else {
+                    assert(false, this.getOptionName() + ' option requires "allExcept" to only have ' +
+                        '`inCallExpressions` or `inArrayExpressions`');
```
markelog: Seriously, why does everybody use " ` " sign?

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,15 +111,55 @@ var excludes = {
...
+                }
+            }).bind(this));
```
markelog: you can pass context as second argument to the `forEach`

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,15 +111,55 @@ var excludes = {
...
-            value === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        if (typeof value === 'object') {
+            assert(Array.isArray(value.allExcept), this.getOptionName() + ' option requires "allExcept" ' +
+                'to be an array');
+            assert(value.allExcept.length > 0, this.getOptionName() + ' option requires "allExcept" ' +
+                'to have at least one item or be set to `true`');
+
+            value.allExcept.forEach((function(except) {
```
markelog: `((` really subjective

TheSavior:
Sorry for taking so long to get to updating this PR. @markelog I fixed the things you mentioned.

I need to look into why this build is failing more. When I run `npm test` on my computer I don't get any errors.

markelog:
Code is simplified, failures are fixed
Thank you.

PR #1243, opened 2015-04-07
Repaired `node.parentCollection` for switch statements

### LINKED ISSUE #916: requireMultipleVarDecl `true` (or consecutive throws an error) inside switch cases
```
scanEscapeSequence: function (checks) {
    var allowNewLine = false;
    var jump = 1;
    this.skip();
    var char = this.peek();

    switch (char) {
    case "'":
      this.triggerAsync("warning", {
        code: "W114",
        line: this.line,
        character: this.char,
        data: [ "\\'" ]
      }, checks, function () {return state.jsonMode; });
      break;
    case "b":
      char = "\\b";
      break;
    case "f":
      char = "\\f";
      break;
    case "n":
      char = "\\n";
      break;
    case "r":
      char = "\\r";
      break;
    case "t":
      char = "\\t";
      break;
    case "0":
      char = "\\0";

      // Octal literals fail in strict mode.
      // Check if the number is between 00 and 07.
      var n = parseInt(this.peek(1), 10);
      this.triggerAsync("warning", {
        code: "W115",
        line: this.line,
        character: this.char
      }, checks,
      function () { return n >= 0 && n <= 7 && state.directive["use strict"]; });
      break;
    case "u":
      char = String.fromCharCode(parseInt(this.input.substr(1, 4), 16));
      jump = 5;
      break;
    case "v":
      this.triggerAsync("warning", {
        code: "W114",
        line: this.line,
        character: this.char,
        data: [ "\\v" ]
      }, checks, function () { return state.jsonMode; });

      char = "\v";
      break;
    case "x":
      var  x = parseInt(this.input.substr(1, 2), 16);

      this.triggerAsync("warning", {
        code: "W114",
        line: this.line,
        character: this.char,
        data: [ "\\x-" ]
      }, checks, function () { return state.jsonMode; });

      char = String.fromCharCode(x);
      jump = 3;
      break;
    case "\\":
      char = "\\\\";
      break;
    case "\"":
      char = "\\\"";
      break;
    case "/":
      break;
    case "":
      allowNewLine = true;
      char = "";
      break;
    }

    return { char: char, jump: jump, allowNewLine: allowNewLine };
  }
```

reports:

```
jscs src/lex.js 
TypeError: Cannot read property 'type' of undefined
    at /usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:8:24
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/usr/local/lib/node_modules/jscs/lib/js-file.js:322:42)
    at Object.consecutive [as _check] (/usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:4:10)
    at Object.module.exports.check (/usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:86:28)
    at null.<anonymous> (/usr/local/lib/node_modules/jscs/lib/string-checker.js:110:22)
    at Array.forEach (native)
    at StringChecker.checkString (/usr/local/lib/node_modules/jscs/lib/string-checker.js:108:35)
    at null.<anonymous> (/usr/local/lib/node_modules/jscs/lib/checker.js:40:25)
    at Array.0 (/usr/local/lib/node_modules/jscs/node_modules/vow/lib/vow.js:699:39)
```

It looks like `parentCollection`, which is only used by this one rule may be being populated incorrectly...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7702448-requiremultiplevardecl-true-or-consecutive-throws-an-error-inside-switch-cases?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1163: rule requireMultipleVarDecl: true crashes jscs cli
When I have the following .jscs.json file:
```json
{
    "requireMultipleVarDecl": true
}
```

And run the commandline tool, I get an TypeError: 

```
$ jscs --version
1.11.3
$ jscs .
TypeError: Cannot read property 'type' of undefined
    at /usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:41:24
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/usr/local/lib/node_modules/jscs/lib/js-file.js:378:42)
    at Object.consecutive [as _check] (/usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:37:10)
    at Object.module.exports.check (/usr/local/lib/node_modules/jscs/lib/rules/require-multiple-var-decl.js:119:28)
    at null.<anonymous> (/usr/local/lib/node_modules/jscs/lib/string-checker.js:157:22)
    at Array.forEach (native)
    at StringChecker.checkString (/usr/local/lib/node_modules/jscs/lib/string-checker.js:155:35)
    at null.<anonymous> (/usr/local/lib/node_modules/jscs/lib/checker.js:40:25)
    at Array.<anonymous> (/usr/local/lib/node_modules/jscs/node_modules/vow/lib/vow.js:699:39)
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9439615-rule-requiremultiplevardecl-true-crashes-jscs-cli?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

twolfson:
# Explanation
After moving to `estraverse` for `tree-iterator` in #817, we introduced a regression when resolving `node.parentCollection`. We want to resolve the sibling nodes for a node by looking at the property its parent is referring to.

For example, in the script `function a() {} function b() {}`, the `Program` node would have a `body` of an array of 2 `FunctionDeclaration` nodes. If we wanted to find this array from the perspective of `a`, then we would look at its parent node and the key it used to access it (e.g. `body`).

In most cases for `VariableDeclaration's`, the containing node has this declaration inside of a `body` key (e.g. `Program`, `FunctionDeclaration`, `BlockStatement`).

https://github.com/estree/estree/blob/9a35a3d091af6ff9cf7fadfe27c49e22533b2bce/spec.md#programs

https://github.com/estree/estree/blob/9a35a3d091af6ff9cf7fadfe27c49e22533b2bce/spec.md#functions

https://github.com/estree/estree/blob/9a35a3d091af6ff9cf7fadfe27c49e22533b2bce/spec.md#blockstatement

> Fwiw `IfStatement`/`CatchClause`/etc all have a `BlockStatement` as the children which is the proper container for these `VariableDeclaration's`.

However in the case of a `SwitchCase`, we use `consquent` as the key:

https://github.com/estree/estree/blob/9a35a3d091af6ff9cf7fadfe27c49e22533b2bce/spec.md#switchcase

# Problem/solution
The problem we had was that `tree-iterator` wasn't stopping once it found the key. It continued traversing up to the root node (`Program`). As a result, the key was always `body` and this worked great for most cases (e.g. `Program`, `FunctionDeclaration`, `BlockStatement`). However, in the case of `SwitchStatement`, it would yield `undefined`.

In the case of our test failures, the `parentCollection` would be `[undefined]` which means a bad `indexOf` lookup, leading to an error thrown when accessing a property of `undefined`.

To remedy this, we stop looping at the first parent's key via a `break`.

In this PR:

- Added regression tests
- Added patch to `tree-iterator`

Fixes #916. Fixes #1163.

**Notes:**

It should be noted that `parentCollection` is only used by `require-multiple-var-decl`. Additionally, this touches `_buildNodeIndex` which may lead to performance speed up (yey).

https://github.com/jscs-dev/node-jscs/blob/v1.12.0/lib/js-file.js#L728-L738

### DIFF
--- a/lib/tree-iterator.js
+++ b/lib/tree-iterator.js
@@ -18,6 +18,7 @@ function iterate(node, cb) {
                         var pathElement = path.pop();
                         if (typeof pathElement === 'string') {
                             collectionKey = pathElement;
+                            break;
                         }
                     }
 
--- a/test/specs/rules/require-multiple-var-decl.js
+++ b/test/specs/rules/require-multiple-var-decl.js
@@ -26,6 +26,12 @@ describe('rules/require-multiple-var-decl', function() {
         it('should not report separated var decl', function() {
             assert(checker.checkString('var x; x++; var y;').isEmpty());
         });
+        it('supports var decl not contained by a parent with a `body` property (#916, #1163)', function() {
+            assert(checker.checkString('switch (1) { case 1: var x; }').isEmpty());
+        });
+        it('should report consecutive var decl not contained by a parent with a `body` property', function() {
+            assert(checker.checkString('switch (1) { case 1: var x; var y; }').getErrorCount() === 1);
+        });
     });
 
     describe('onevar', function() {


twolfson:
Ah, I almost forgot. I have skipped adding regression tests to `tree-iterator` as it seems `tree-iterator` lacks any in-depth tests at all =/

https://github.com/jscs-dev/node-jscs/blob/v1.12.0/test/specs/tree-iterator.js

Let me know if you want me to add these/what they should contain besides the regression tests.

markelog:
Nicely done! Thank you

PR #1246, opened 2015-04-07
paddingNewLinesAfterBlocks: ease up on function calls

### LINKED ISSUE #1232: requirePaddingNewLinesAfterBlocks seems overly strict

markelog:
Fixes #1232

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -105,6 +105,15 @@ module.exports.prototype = {
                     return;
                 }
 
+                // Check if this block is used in call expression
+                if (parentNode.parentNode.type === 'CallExpression') {
+                    for (var i in parentNode.parentNode.arguments) {
+                        if (parentNode.range === parentNode.parentNode.arguments[i].range) {
+                            return;
+                        }
+                    }
+                }
+
                 errors.assert.linesBetween({
                     token: closingBracket,
                     nextToken: nextToken,
--- a/test/specs/rules/require-padding-newlines-after-blocks.js
+++ b/test/specs/rules/require-padding-newlines-after-blocks.js
@@ -85,4 +85,13 @@ describe('rules/require-padding-newlines-after-blocks', function() {
     it('should not report missing padding in function chain', function() {
         assert(checker.checkString('[].map(function() {})\n.filter(function(){})').isEmpty());
     });
+
+    it('should not be so strict in function calls (#1232)', function() {
+        assert(checker.checkString('callingAFunction("with", "another", function() {}\n)').isEmpty());
+        assert(checker.checkString('foo(\narr.map(function() {\n})\n);').isEmpty());
+    });
+
+    it('should not be so strict in function calls if it is not the last argument (#1232)', function() {
+        assert(checker.checkString('callingAFunction("with",function() {},\n"another")').isEmpty());
+    });
 });


mikesherov:
What about function calls that aren't the last parameter?

hzoo:
@mikesherov I was suggesting check `nextToken.value === ','` as well if that works? I think that is the issue for #1237.

markelog:
@hzoo nope, have an idea, checking it out

markelog:
Okay, how about now?

hzoo:
Yeah #1237 passes with that.

markelog:
Oh, cool :-)

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -105,6 +105,15 @@ module.exports.prototype = {
                     return;
                 }
 
+                // Check if this block is used in call expression
+                if (parentNode.parentNode.type === 'CallExpression') {
```
mikesherov: much better. Finally, using the ast as expected. ;-)

mikesherov:
:+1:

mikesherov:
@TheSavior 's comment still applies.

TheSavior:
I still don't think this is the right solution as I don't believe this passes the tests that I added in my solution. @mikesherov please look at @1239 for comparison.

lib/rules/require-padding-newlines-after-blocks.js
zxqfox: According to @TheSavior comment: feels like we need to add an exception for `Program` node. No?

TheSavior: See https://github.com/jscs-dev/node-jscs/pull/1239/files#diff-3c958cdc2aa18d8507669afbc4372ee5R154

mikesherov: @TheSavior 's solution is correct in that it won't barf on top level blocks. However, I don't think we need to add any options.

TheSavior: My solution also handles arrays, I'm happy to remove the configuration options.

markelog: So test-case, code-example?

> My solution also handles arrays

I'm not sure if we need that at the moment

TheSavior: @markelog Can you make a case why we wouldn't want this to be handled in arrays?

markelog: We require newline char in objects but not in arrays? What is the diff?

markelog: Alrighty, i don't see an example when we would need exception for `parentNode.parentNode` top-level element. 

If find one, let me know

Issue #812, opened 2014-12-05
Allow differentiation of call expressions in the requireSpacesInsideParentheses rule


jamesallardice:
The style guide that I adhere to contains the following conventions:

```js
var x = getX(y, z); // No spaces inside call expression parens

if ( x ) {
  // Spaces inside the parens of control flow structures and function definitions
}
```

The `requireSpacesInsideParentheses` rule does not provide any mechanism for differentiating between these two cases. I think there's a few possible solutions:

 - A custom rule if this isn't something that would be widely useful
 - A modification to `requireSpacesInsideParentheses`. The rule already accepts an object so a property could be added to this. A set of values could also be added to the `except` array:

```js
"requireSpacesInsideParentheses": {
  "all": true,
  "except": [ "if", "while" /*...*/ ]
}
```

 - A new rule could be added

hzoo:
Is this closed now?

markelog:
No, that pr only for grouping parentheses

hzoo:
Going to be closing older/inactive issues and labeling with `orphaned` to track anything we might want to reopen again later (if we want to add it again or something has a PR). Thanks :smiley: !

Issue #431, opened 2014-06-10
"nested" option for disallowSpacesInsideParentheses rule


markelog:

markelog:
I guess this is no need for that anymore

PR #1254, opened 2015-04-10
New Rule: requireSpacesInsideParenthesizedExpression

gibson042:
A proof of concept for distinguishing the several parenthetical productions (statement affix, function definition parameters, function call arguments, and arbitrary expressions), inspired by https://github.com/jquery/sizzle/pull/330/files#r27985694 . This PR implements only **requireSpacesInsideParenthesizedExpression**, but if you like the concepts herein I will add **disallowSpacesInsideParenthesizedExpression** so it can merge.

The obvious next step is to introduce similar pairs for **SpacesInsideParenthesizedArguments**, **SpacesInsideParenthesizedParameters**, and **SpacesInsideParenthesizedStatement** (names subject to change, but chosen for convenient alphabetic grouping) in a pattern analogous to that embodied by the "**SpacesInFunction**" rules but rooted at "**SpacesInsideParentheses**".

I believe this also relates to existing work:
Ref https://github.com/jscs-dev/node-jscs/pull/1069#discussion_r24805456
Ref gh-812
Ref gh-431


gibson042:
@jscs-dev is there something more to do for feedback on this?

mrjoelkemp:
@gibson042 Sorry this slipped through the cracks. This needs to be rebased and the coverage needs to be increased.

@hzoo Does this look good aside from the above issues? I see it's holding up the Idiomatic Preset PR: https://github.com/jscs-dev/node-jscs/pull/1069. Would like to land both (this and Idiomatic) prs.

gibson042:
Rebase complete. Why did the coverage report not pick up use of the new util functions by rules requiring them, though? But at any rate, they are now specifically tested.

mikesherov:
> Why did the coverage report not pick up use of the new util functions by rules requiring them, though?

That's a feature, not a bug. The unit coverage runner, written by @mdevils, explains the philosophy here:  https://github.com/unit-coverage/unit-coverage#the-idea

gibson042:
Where does this PR stand? The missing coverage seemed reasonable to me (and consistent with preexisting files), but I can extend it if there's a non-decrease requirement. More than that, though, I want feedback on the approach and the ideas... as noted, this is a POC.

lib/rules/require-spaces-inside-parenthesized-expression.js
```diff
@@ -0,0 +1,120 @@
...
+
+    check: function(file, errors) {
+        var exceptions = this._exceptions;
+
+        file.iterateTokenByValue('(', function(token) {
+            var nextToken = file.getNextToken(token, {includeComments: true});
+            var value = nextToken.value;
+
+            // Skip empty parentheses and explicit exceptions
+            if (value === ')' || value in exceptions) {
```
mrjoelkemp: Why not set ')' as a default part of the exceptions object? You can then simplify this `if` statement.

lib/utils.js
```diff
@@ -181,6 +187,117 @@ exports.getFunctionNodeFromIIFE = function(node) {
...
 
 /**
+ * Returns the type of AST node or ECMAScript production in which the provided
+ * open parenthesis token is included.
+ *
+ * @param {Object} token
+ * @param {JsFile} file
+ * @returns {String}
+ */
+exports.categorizeOpenParen = function(token, file) {
```
mrjoelkemp: I'm not particularly fond of adding more function to utils. At the very least, these should be functions dangled on the rule export unless other rules will use them. If they're going to be reused, then these functions could be properties of a `TokenCategorizer` class that's within `lib`.

mrjoelkemp:
This is really great work, @gibson042. Bravo.

lib/rules/require-spaces-inside-parenthesized-expression.js
gibson042: Because it's only an exception in one direction—if I universally except ")", then the immediately following close-parenthesis logic would erroneously allow "((…))".

lib/utils.js
gibson042: Good idea. So it would be analogous to string-checker.js or tree-iterator.js?

mrjoelkemp: Yes

gibson042:
Thanks for the review. Am I correct in assuming that this has legs and that I should round it out with the above `TokenCategorizer` suggestion and **disallowSpacesInsideParenthesizedExpression** inverse rule?

gibson042:
(and then the **SpacesInsideParenthesizedArguments** [function call], /**SpacesInsideParenthesizedParameters** [function definition], and **SpacesInsideParenthesizedStatement** [`if`/`for`/`while`/etc.] pairs, of course). Any thoughts on the names?

mrjoelkemp:
Sounds like a plan to me. Maybe wait for the second go ahead from
@markelog, @hzoo, or @zxqfox.

test/specs/utils.js
```diff
@@ -113,6 +113,215 @@ describe('modules/utils', function() {
...
+            });
+        });
+    });
+
+    describe('categorizeCloseParen', function() {
+        var sharedFile = createJsFile(
+            '(((function(){ function f(){} if(0) return((f)(0, (1), ((2)))+0); throw(f()+(0)); })))'
+        );
+        var closeParens = [];
+        sharedFile.iterateTokenByValue(')', function(token) {
```
zxqfox: I believe it should be wrapped to `before` statement inside this `describe`.

gibson042:
Ok, I believe this is ready. Points for special consideration:
* comments handling: is configuration like `allExcept: ['//', '/*', '*/']` reasonable?
* naming: **ParenthesizedExpression**, **ParenthesizedArguments**, **ParenthesizedParameters**, and **ParenthesizedStatement** alphabetize conveniently, but may not be as intuitive as e.g. **ParenthesizedExpression**, **Call**(**Expression**)?**Arguments**, (**Function**)?**Parameters**, and **StatementParentheses**

markelog:
How would we enforce 
```js
(options || { x: true } ).x
```
but disallow
```js
(options || { x: true }).x
```
?

test/specs/rules/require-spaces-inside-parenthesized-expression.js
```diff
@@ -0,0 +1,169 @@
...
+            });
+
+            assert(checker.checkString('(foo)').getErrorCount() === 2);
+            assert(checker.checkString('(/**/ foo /**/)').isEmpty());
+            assert(checker.checkString('(//\n foo //\n)').getErrorCount() === 1);
+            assert(checker.checkString('(/*comment*/ foo /*comment*/)').isEmpty());
+            assert(checker.checkString('(//comment\n foo //comment\n)').getErrorCount() === 1);
+        });
+
+        it('should not look inside comments', function() {
```
markelog: is this description correct?

markelog:
Did you run it on core source?

> comments handling: is configuration like allExcept: ['//', '/*', '*/'] reasonable?

It is for me, don't see any better way

> Naming: ParenthesizedExpression, ParenthesizedArguments, ParenthesizedParameters, and ParenthesizedStatement alphabetize conveniently, but may not be as intuitive as e.g. ParenthesizedExpression, Call(Expression)?Arguments, (Function)?Parameters, and StatementParentheses

What would those rules do? Could you give me couple examples?

test/specs/rules/require-spaces-inside-parenthesized-expression.js
gibson042: Yes. These rules should look only at the characters beginning or ending comments, not at their contents.

gibson042:
> How would we enforce
```js
(options || { x: true } ).x
```
but disallow
```js
(options || { x: true }).x
```
?

I'm not sure I understand... can you provide an English explanation of desired behavior? I've followed the other JSCS patterns, so it's possible to require spaces with a finite set of exceptions and also possible to disallow spaces with a finite set of exceptions, but AFAIK _not_ possible to require spaces **only** for a finite set of "inside" characters.

> Did you run it on core source?

For jQuery? Not yet.

> > Naming: ParenthesizedExpression, ParenthesizedArguments, ParenthesizedParameters, and ParenthesizedStatement alphabetize conveniently, but may not be as intuitive as e.g. ParenthesizedExpression, Call(Expression)?Arguments, (Function)?Parameters, and StatementParentheses

> What would those rules do? Could you give me couple examples?

**ParenthesizedArguments** or **CallArguments** or **CallExpressionArguments**: `foo(‸a, b, …‸)`, `foo.bar(‸a, b, …‸)`, `(foo)(‸a, b, …‸)`, etc.
**ParenthesizedParameters** or **FunctionParameters** or **Parameters**: `function(‸a, b, …‸) { … }`, `function foo(‸a, b, …‸) { … }`, etc.
**ParenthesizedStatement** or **StatementParentheses**: `if (‸…‸)`, `for(‸…; …; …‸)`, `catch (‸x‸)`, etc.
**ParenthesizedExpression**: everything else

gibson042:
> I'm not sure I understand... can you provide an English explanation of desired behavior? I've followed the other JSCS patterns, so it's possible to require spaces with a finite set of exceptions and also possible to disallow spaces with a finite set of exceptions, but AFAIK _not_ possible to require spaces **only** for a finite set of "inside" characters.

I stand corrected... [disallowSpacesInsideParentheses](http://jscs.info/rule/disallowSpacesInsideParentheses.html) has an `only` option, although it appears to be unique. Would it be acceptable to add and/or reassess that outside of this PR?

> > Did you run it on core source?

> For jQuery? Not yet.

I checked master-branch jquery.js with `{ disallowSpacesInsideParenthesizedExpression: true }`, and found over 100 exceptions (which is not surprising, because jQuery will likely use the _other_ three rules and leave this one unspecified... cf. https://github.com/jquery/sizzle/pull/330/files#r27985694 ). However, file analysis took _two minutes_. I profiled jscs, and traced it down to inefficiencies in `estraverse.traverse` by way of `JsFile#iterate` by way of `JsFile#getNodeByRange`. Not wanting to dive into the dependency, I addressed the outermost of these in e97c8ebf32a6a14283dad06684386975d03a9e60, which is technically out of scope for this PR but lowered runtime to 1.5 seconds (a reduction of **99%**). If you'd rather see a separate PR for that, I could accommodate.

lib/rules/disallow-spaces-inside-parenthesized-expression.js
```diff
@@ -0,0 +1,124 @@
...
+            // Skip non-expression parentheses
+            var type = TokenCategorizer.categorizeOpenParen(token, file);
+            if (type !== 'ParenthesizedExpression') {
+                return;
+            }
+
+            errors.assert.noWhitespaceBetween({
+                token: token,
+                nextToken: nextToken,
+                message: 'Illegal space after opening grouping round bracket'
```
gibson042: I suppose this is also worth calling out. "Round bracket" reads weird to me, but it seems to be the terminology used here.

markelog: We had a discussion about it somewhere, for standardizing error output and rule names, in the new approach it would be cool to follow common terminology - https://en.wikipedia.org/wiki/Bracket.

In other words, i think it would be okay to call it "parentheses".

markelog:
> Would it be acceptable to add and/or reassess that outside of this PR?

Totally 

> I addressed the outermost of these in e97c8eb, which is technically out of scope for this PR but lowered runtime to 1.5 seconds (a reduction of 99%). 

Wow! That would be great to see it in the new release, i guess another PR would be best.

If no one has any objections, i will land this a bit later today.

Awesomely done Richard!

lib/rules/disallow-spaces-inside-parenthesized-expression.js
hzoo: I think it was https://github.com/jscs-dev/node-jscs/issues/698#issuecomment-61195949

https://github.com/jscs-dev/node-jscs/issues/538

gibson042:
"Round bracket" references replaced and `getNodeByRange` improvement moved to a [new PR](https://github.com/jscs-dev/node-jscs/issues/1713).

markelog:
Hooray!

PR #1258, opened 2015-04-12
paddingNewLinesBeforeLineComments: code and comment on the same line

### LINKED ISSUE #1194: requirePaddingNewLinesBeforeLineComments should not consider code and comment on the same line
There should be an option to allow this to be valid:

```javascript
var x = 'foo'; // jshint ignore:line
```

Right now it will complain that it needs to be preceeded by a new line.

There is already a configuration for an allExcept:

```
"requirePaddingNewLinesBeforeLineComments": {
  "allExcept": "firstAfterCurly"
}
```

allExcept should probably be an array. 

```
"requirePaddingNewLinesBeforeLineComments": {
  "allExcept": ["firstAfterCurly", "inlineComments"]
}
```

Thoughts?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10057043-requirepaddingnewlinesbeforelinecomments-should-not-consider-code-and-comment-on-the-same-line?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Fixes #1194

### DIFF
--- a/lib/rules/require-padding-newlines-before-line-comments.js
+++ b/lib/rules/require-padding-newlines-before-line-comments.js
@@ -18,6 +18,7 @@
  *
  * ```js
  * var a = 2;
+ * var b = 3; // comment
  *
  * // comment
  * return a;
@@ -84,6 +85,13 @@ module.exports.prototype = {
 
         file.iterateTokensByType('Line', function(comment) {
             if (comment.loc.start.line === 1) {
+                return;
+            }
+
+            var firstToken = file.getFirstTokenOnLine(comment.loc.start.line);
+
+            // Should not consider code and comment on the same line (#1194)
+            if (firstToken !== undefined && firstToken.type !== 'EOF') {
                 return;
             }
 
--- a/test/specs/rules/require-padding-newlines-before-line-comments.js
+++ b/test/specs/rules/require-padding-newlines-before-line-comments.js
@@ -85,6 +85,12 @@ describe('rules/require-padding-newlines-before-line-comments', function() {
         it('should report error if first line after a curly', function() {
             assert(checker.checkString('if (true) {\n// comment\n}').getErrorCount() === 1);
         });
+
+        it('should not consider code and comment on the same line (#1194)', function() {
+            assert(checker.checkString('var a; \n var b; //comment\n').isEmpty());
+            assert(checker.checkString('var a; \n var b; //comment\nvar c;').isEmpty());
+            assert(checker.checkString('/**/var a; \n var b// comment\n').isEmpty());
+        });
     });
 
     describe('value allExcept: firstAfterCurly', function() {


test/specs/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -85,6 +81,12 @@ describe('rules/require-padding-newlines-before-line-comments', function() {
         it('should report error if first line after a curly', function() {
             assert(checker.checkString('if (true) {\n// comment\n}').getErrorCount() === 1);
         });
+
+        it('should not consider code and comment on the same line (#1194)', function() {
+            assert(checker.checkString('var a; \n var b; //comment\n').isEmpty());
+            assert(checker.checkString('var a; \n var b; //comment\nvar c;').isEmpty());
+            assert(checker.checkString('/**/var a; \n var b// comment\n').isEmpty());
```
TheSavior: Is this meant to be an unclosed block comment?

test/specs/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -46,10 +46,6 @@ describe('rules/require-padding-newlines-before-line-comments', function() {
             checker.configure({ requirePaddingNewLinesBeforeLineComments: true });
         });
 
-        it('should report missing padding before line comment', function() {
```
TheSavior: I'm not sure why this was removed. This seems like exact test that this rule is supposed to fail.

lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -87,6 +88,13 @@ module.exports.prototype = {
                 return;
             }
 
+            var firstToken = file.getFirstTokenOnLine(comment.loc.start.line);
+
+            // Should not consider code and comment on the same line (#1194)
+            if (firstToken !== undefined && firstToken.type !== 'EOF') {
```
TheSavior: `if (firstToken !== comment)`?

test/specs/rules/require-padding-newlines-before-line-comments.js
markelog: Nope, just in case assertion

test/specs/rules/require-padding-newlines-before-line-comments.js
markelog: Removed for the debug purposes, reverted

lib/rules/require-padding-newlines-before-line-comments.js
markelog: Could you elaborate?

TheSavior: Why do we need to check for undefined and EOF? It seems like we could just check if the first token on the line **IS** the comment. If the comment isn't the first token on the line, return out.

test/specs/rules/require-padding-newlines-before-line-comments.js
TheSavior: ah `/* */` but without a space. Gotcha. For some reason this confused me. All is well, ignore me. :P

zxqfox:
Let's land this?

PR #1260, opened 2015-04-12
requirePaddingNewLinesAfterBlocks: add array and object exceptions

markelog:
Ref gh-1239

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -22,16 +22,6 @@
  *
  *         continue;
  *     }
- *
- *     var obj = {
- *         foo: function() {
- *             return 1;
- *         },
- *
- *         bar: function() {
- *             return 2;
- *         }
- *     };
  * }
  * ```
  *
@@ -45,14 +35,6 @@
  *         }
  *         continue;
  *     }
- *     var obj = {
- *         foo: function() {
- *             return 1;
- *         },
- *         bar: function() {
- *             return 2;
- *         }
- *     };
  * }
  * ```
  */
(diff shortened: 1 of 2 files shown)


mikesherov:
:+1:

TheSavior:
:+1: Thanks

markelog:
Added object exception

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,6 +49,27 @@ var excludes = {
...
 };
 
+var exceptions = {
+    'CallExpression': true,
+    'ArrayExpression': true,
+    'Property': true
+};
+
+function isException(parent) {
+    var parParent = parent.parentNode;
```
zxqfox: grandpa?

zxqfox:
LGTM

lib/rules/require-padding-newlines-after-blocks.js
markelog: okay :-)

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -67,6 +49,27 @@ var excludes = {
...
 };
 
+var exceptions = {
+    'CallExpression': true,
+    'ArrayExpression': true,
+    'Property': true
+};
+
+function isException(parent) {
+    var grandpa = parent.parentNode;
```
zxqfox: :smile:

TheSavior:
I **disagree** with the change to this rule that doesn't enforce the extra line in objects. That was one of the key things I **needed** when I wrote this rule. Airbnb's style guide agreed and approved that. Why was it removed here?

zxqfox:
@TheSavior Thanks for coming. Perhaps we need not to restore functionality you needed here in separate rule to left rules simple and focused.

And also update presets to be consistent with new rules.

Thoughts?

markelog:
Hm, after some thinking, i would prefer if we would merge https://github.com/jscs-dev/node-jscs/pull/1239 and this one together, while adding third option for ignoring object expressions.

How is that sound?

TheSavior:
I'm fine with that. When you say merge the two together, what do you mean? Does this PR provide any other behavior? Or should we just add the third option to #1239 and merge that one?

markelog:
I'm thinking we could move exceptions check to the `isException` helper

PR #1262, opened 2015-04-13
requirePaddingNewLinesBeforeLineComments fix for file starting with line comment

### LINKED ISSUE #1261: requirePaddingNewLinesBeforeLineComments breaks if no previous token
```
if (prevToken.type === 'Line') {
    return;
}

if (allowFirstAfterCurly && prevToken.type === 'Punctuator' && prevToken.value === '{') {
    return;
}
```

If `prevToken` is undefined here, this code will break.  I happen to have some files where, likely, there's a newline to start the file (so `comment.loc.start.line` is not 1).  Chances are, there should be an early return here if `prevToken` is `undefined`.  

In any case, running jscs on a large project only to see:

`Warning: Cannot read property 'type' of undefined Use --force to continue.` isn't ideal.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11061838-requirepaddingnewlinesbeforelinecomments-breaks-if-no-previous-token?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
Fix for file starting with line comment.

Fixes #1261

### DIFF
--- a/lib/rules/require-padding-newlines-before-line-comments.js
+++ b/lib/rules/require-padding-newlines-before-line-comments.js
@@ -83,13 +83,9 @@ module.exports.prototype = {
         var allowFirstAfterCurly = this._allowFirstAfterCurly;
 
         file.iterateTokensByType('Line', function(comment) {
-            if (comment.loc.start.line === 1) {
-                return;
-            }
-
             var prevToken = file.getPrevToken(comment, {includeComments: true});
 
-            if (prevToken.type === 'Line') {
+            if (!prevToken || prevToken.type === 'Line') {
                 return;
             }
 
--- a/test/specs/rules/disallow-padding-newlines-before-line-comments.js
+++ b/test/specs/rules/disallow-padding-newlines-before-line-comments.js
@@ -37,4 +37,8 @@ describe('rules/disallow-padding-newlines-before-line-comments', function() {
     it('should not report missing padding with block comment', function() {
         assert(checker.checkString('var a = 2;\n/* comment */').isEmpty());
     });
+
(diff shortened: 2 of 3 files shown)


lib/rules/require-padding-newlines-before-line-comments.js
```diff
@@ -83,13 +83,9 @@ module.exports.prototype = {
         var allowFirstAfterCurly = this._allowFirstAfterCurly;
 
         file.iterateTokensByType('Line', function(comment) {
-            if (comment.loc.start.line === 1) {
```
zxqfox: Does it affects shebangs?
```
#!/usr/bin/env node
// etc.?
```

TheSavior: What do you feel like the expected behavior here? I would expect a blank line between the shebang and comment.

zxqfox: I'd say
```
#!/usr/bin/env

// line comment
// etc.
```
But
```
// line comment
// etc.
```

Well, it's very rare and hard to implement case (unless we have a isShebanged flag in jsfile). Nvm! ;-)

zxqfox: Am I drunk? Not sure. Looks like it works well already ;-)

PR #1264, opened 2015-04-13
Adding a test helper for asserting errors and fixes

TheSavior:
Enabling tests to be easily written to confirm that we are writing tests that are fixing properly.

Refs @mikesherov's recommendation in #1128
Should be useful for updating the rules in #1127
Needed for #1202

Issue #1201, opened 2015-03-27
Autoformat: validateQuoteMark


indexzero:
Should be a straight-forward enough rule to fix tokens on. PR coming shortly.

TheSavior:
This is something that should definitely have a lot of tests. I could see it running into issues with strings with/without escape chars for the opposite quote type. I could also see it being an issue with multiple concatenated strings with different quote types. Multi line strings would be another good thing to check.

p3k:
:+1:

Issue #1113, opened 2015-02-24
Fix field in rule declaration


markelog:
See discussion in https://github.com/jscs-dev/node-jscs/pull/1087#issuecomment-75753478

markelog:
@mrjoelkemp @zxqfox @mikesherov @mdevils would like some reaction to this, there is some rules that could not be used with basic assertions without bloating amount of functions for this module.

mikesherov:
@markelog want to talk about it after 1.12 is out. Sorry for delaying discussion, just have lots going on.

zxqfox:
Sorry for delay, guys.

> But there doesn't seem to be many other needs for additional assertions, and I think the fear of an overblown is just that, fear.

:+1: 

> even if we add assertions to every rule in our codebase it wouldn't mean any external rule could use it. Which would mean auto-fixing wouldn't possible for the them.

:+1: 

> On the other hand, give users API to change your code is dangerous.

:+1:, but it's a fear ;-) If you using a plugin — ask plugin developer why it fails.

If you using JSCS as a plugin to IDE or Some smart editor — usually you have a choice to not save file and, moreover, cancel changes with Ctrl+Z/Cmd+Z.

What you think about Assertion API. Is it ready to be public?

PR #1265, opened 2015-04-14
Adding tests for autofixing validate quotes

TheSavior:
Continuing off of #1201 and #1202. I don't have push permissions to that PR and I needed to rebase the branch, so sadly I had to create a new PR.

This is rebased off of #1264 to be able to have the new test fix assertions. That PR should be merged first.

The code seemed to pass the tests with no additional changes. The code currently does not fix multi line strings, but I think this is a great first step and is functionally mergeable as is. 

@indexzero Awesome job on this.

### DIFF
--- a/lib/rules/validate-quote-marks.js
+++ b/lib/rules/validate-quote-marks.js
@@ -106,10 +106,16 @@ module.exports.prototype = {
                     return;
                 }
 
+                token.value = token.value.replace(
+                    new RegExp('^' + mark + '(.*)' + mark + '$'),
+                    opposite[mark] + '$1' + opposite[mark]
+                );
+
                 errors.add(
                     'Invalid quote mark found',
                     token.loc.start.line,
-                    token.loc.start.column
+                    token.loc.start.column,
+                    true
                 );
             }
         });
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -31,7 +31,7 @@ Errors.prototype = {
      * @param {Number|Object} line
      * @param {Number} [column] optional if line is an object
      */
-    add: function(message, line, column) {
+    add: function(message, line, column, fixed) {
         if (typeof line === 'object') {
             column = line.column;
             line = line.line;
@@ -46,7 +46,8 @@ Errors.prototype = {
         this._addError({
             message: message,
             line: line,
-            column: column
+            column: column,
+            fixed: fixed
         });
     },
 
--- a/test/specs/rules/validate-quote-marks.js
+++ b/test/specs/rules/validate-quote-marks.js
@@ -1,8 +1,10 @@
 var Checker = require('../../../lib/checker');
 var assert = require('assert');
+var assertHelpers = require('../../lib/assertHelpers');
 
 describe('rules/validate-quote-marks', function() {
     var checker;
+    var rules;
 
     beforeEach(function() {
         checker = new Checker();
@@ -24,12 +26,17 @@ describe('rules/validate-quote-marks', function() {
     });
 
     describe('option value \' ', function() {
+        rules = { validateQuoteMarks: '\'' };
+
         beforeEach(function() {
-            checker.configure({ validateQuoteMarks: '\'' });
+            checker.configure(rules);
         });
 
-        it('should report double quotes in strings', function() {
-            assert(checker.checkString('var x = "x";').getErrorCount() === 1);
+        assertHelpers.reportAndFix({
+            name: 'double quotes in strings',
+            rules: rules,
+            input: 'var x = "x";',
+            output: 'var x = \'x\';'
         });
 
         it('should not report single quotes in strings', function() {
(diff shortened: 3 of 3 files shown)


TheSavior:
Also, I have run this on our large code base and didn't find any unintended consequences.

mikesherov:
Can you move the one file to `test/lib/assertHelpers`?

mikesherov:
This LGTM with that one suggested change. Thanks again as usual @TheSavior !

TheSavior:
@mikesherov, that comment is applicable to #1264. I'll move it there and rebase this. That PR should be merged first.

zxqfox:
@TheSavior #1264 was merged!

TheSavior:
Alright, rebased off master.

zxqfox:
Something goes wrong definitely:
- You adding test/assertHelpers.js in 0940a14
- Then you moving it out in 991a7a0

I'd try to put b3b3ff1 before the rest and squash the rest to one.

Like start `git rebase -i origin/master`, and put instructions like:
```
pick b3b3ff1
pick 0940a14
f 8e93382
f 991a7a0
```

Can you try it?

TheSavior:
There we go. I just cleaned it up and squashed it all down to one.

TheSavior:
*phew*. Builds passed.

zxqfox:
:+1: Fine now ;-) Thanks!

LGTM.

@markelog @mikesherov Final look?

TheSavior:
@zxqfox @mikesherov marked it as reviewed-once before the rebase.

markelog:
Needed @mdevils to sign off, i'd say it should be done either through `token-assert` module or though `fix` property in rule declaration

TheSavior:
For reference, the discussion about how to modify `errors` to enable this behavior started here: https://github.com/jscs-dev/node-jscs/pull/1202

lib/rules/validate-quote-marks.js
```diff
@@ -106,10 +106,16 @@ module.exports.prototype = {
                     return;
                 }
 
+                token.value = token.value.replace(
+                    new RegExp('^' + mark + '(.*)' + mark + '$'),
```
markelog: scarrrry

zxqfox: Can't get how ruin this ;-) You know string without `'` or `"`? ES6 template strings?

TheSavior: Yeah, came from #1202 that I pulled into this one. It seems to work and pass the tests, so I'm not super worried about it. It might need to change if we try to fix multi line strings, but I wouldn't bother changing it until that comes up. We shouldn't be scared of regex just because it is regex. Even though we both seemed to have the same reaction :P

zxqfox:
@TheSavior Yeah, but: "Who knew life, that is no longer hurries" © Omar Khayyam

@markelog Thanks for summoning. ;-)

lib/rules/validate-quote-marks.js
gibson042: That seems overly complicated and slow for what it's doing (which is basically `token.value = opposite[mark] + token.value.slice(1, -1) + opposite[mark]`), and what it's doing seems inaccurate:
* as noted, it will fail on multi-line strings because `.` doesn't match newline (jQuery uses `[\w\W]` for a true "any character" match)
* it will also fail on strings containing unescaped quote characters, e.g. converting raw `"'"` into `'''`

I _**think**_ what you want to do is escape all unescaped quote characters _in_ the string, then change the quote characters _of_ the string... something like:
```js
var unescapedOpposite = {
	'"': /'/g,
	"'": /"/g
};
…
token.value = opposite[mark] +
	str.slice(1, -1).split(/((?:\\[\w\W])+)/).map(function(part, i) {
		return i % 2 === 0 ?
			part.replace(unescapedOpposite[mark], "\\$&") :
			part;
	}).join("") +
	opposite[mark];
```

It's also possible to do with a probably-faster but arguably more complex single-pass `replace`:
```js
var unescapedOpposite = {
	'"': /((?:\\[\w\W]|[^\\])*?)(')/g,
	"'": /((?:\\[\w\W]|[^\\])*?)(")/g
};
…
var start = 0;
token.value = opposite[mark] +
	token.value.slice(1, -1).replace(unescapedOpposite[mark], function(match, prefix, mark, offset) {
		var escaped = offset === start ?
			prefix + "\\" + mark :
			match;
		start = offset + match.length;
		return escaped;
	}) +
	opposite[mark];
```

zxqfox:
```
str = '"asd qwe"';
re = new RegExp('^' + '"' + '(.*)' + '"' + '$');
function _swap(str, q) { return q + str.substr(1, str.length - 2) + q; }

str.replace(re, "'$1'"); // '\'asd qwe\''
_swap(str, '\'') // '\'asd qwe"\''

console.time(); for (var i = 0; i < 1000000; i++) str.replace(re, "'$1'"); console.timeEnd(); // 617ms
console.time(); for (var i = 0; i < 1000000; i++) _swap(str, '\''); console.timeEnd(); // 115ms
```

Safer and faster. Thoughts?

markelog: I'd would use something time-tested, like https://github.com/sindresorhus/to-single-quotes/blob/master/index.js

zxqfox: :+1:

TheSavior: I'm accepting Pull Requests. ;) I didn't write the logic on this one, I pulled it from another PR and updated #1201 and #1202 here.

mdevils: It will probably fail on this:

```
var string = "123'456"; // -> '123'456' — syntax error
```

And also. Let's not change token values inside rules. Let's add another assertion to Assert module.

mikesherov:
@mdevils, there's no reusability for that assertion. I'd vote to keep it out of assertion library until we find reuse for it to keep assertion library from growing out of control.

markelog:
@mdevils +1 to the Mike point

mdevils:
@markelog @mikesherov OK

markelog:
So we decided who not to do, who about we decide how to do it :-).

https://github.com/jscs-dev/node-jscs/issues/1113

?

mdevils:
@markelog my parser failed to process your message)

markelog:
We either fix stuff in `check` method or introduce `fix` property for this "specific" rules that could not be included in the `token-assert` and try to isolate its execution hence link to the relevant issue.

PR #1267, opened 2015-04-14
validateQuoteMarks: support disabling

avdv:
When using a preset which configures the validateQuoteMarks rule, there
was no way to opt-out.

Change this rule to allow `validateQuoteMarks: false` which disables
this rule explicitely.

### DIFF
--- a/lib/rules/validate-quote-marks.js
+++ b/lib/rules/validate-quote-marks.js
@@ -7,6 +7,7 @@
  *  - `"\""`: all strings require double quotes
  *  - `"'"`: all strings require single quotes
  *  - `true`: all strings require the quote mark first encountered in the source code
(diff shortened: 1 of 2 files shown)


hzoo:
You should be able to null any rule. Not sure if its clear enough but in the [docs](http://jscs.info/overview.html#preset) it says you can do so in your `.jscsrc`

```js
{
    "preset": "jquery",
    "requireCurlyBraces": null
}
```

mikesherov:
@avdv thanks for contributing! @hzoo's comment is correct, so closing this as WONTFIX.

avdv:
@hzoo Thanks a lot, I must admit that I overlooked that in the docs... :open_mouth:

Issue #1271, opened 2015-04-15
Can we catch exceptions and dump inner state of checker?


zxqfox:
```
$ node_modules/.bin/jscs .
TypeError: Cannot read property 'type' of undefined
    at /Users/remnev/stat-counters/node_modules/jscs/lib/rules/require-multiple-var-decl.js:41:24
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/Users/remnev/stat-counters/node_modules/jscs/lib/js-file.js:406:42)
...
```

Case for test:
```js
switch (ab.nodeType) {
    case 1:
        var ad = ab.nodeName;
        break;
}
```
To have some additional info about checking file, etc.

TheSavior:
IIRC, there is a fix out for the specific issue mentioned.

However, I totally agree that we should never be throwing errors to the user from inside the Java code base. Hopefully unit tests avoid these cases but this is a good example of where that was not the case.

It's a fine line though. From the users side, we should wrap JSCS completely in a try catch to be able to say "an error occurred when running JSCS". However, we would then lose all of the debugging information. Perhaps the debugging information would be printed as part of a verbose flag that was mentioned in another issue.

zxqfox:
:+1: Agreed. But we shouldn't stop checking file on exceptions imho. One rule's fail shouldn't ruin all the rest.

And yes, `--verbose` should definitely expose this information to user. It helps a lot for debugging and bug reporting.

zxqfox:
Related issues: #916, #1163

TheSavior:
That is a great point about disabling the rule and continuing running with the others.

zxqfox:
I'm gonna land this if nobody against. Take one more look please.

/cc @mikesherov @TheSavior @hzoo

hzoo:
Looks good?

zxqfox:
@hzoo For me? Seems like. ;-) Is it a question?

Issue #1043, opened 2015-02-10
Integrate Performance Monitoring


mikesherov:
- [ ] Find benchmarking framework like bestie/benchmarkjs
- [ ] write bin script that runs benchmarks
  - [ ] should attempt to check a large file like the jquery source code
  - [ ] should compare the benchmark against an upper limit (our budget)
  - [ ] should exit with an error if over the limit
- [ ] integrate benchmark bin script with travis
- [ ] note process by which we update our budget periodically (perhaps as a pre-release task)

zxqfox:
Guys from travis denied continous benchs: https://github.com/travis-ci/travis-ci/issues/352

Anyway, we can make a static script that will calculate a par of server's performance and then just normalize the result timings of bench via these values: timing s1 / performance s1 vs timing s2 / performance s2, etc.

We can run bench for many of our current tests, just we should mark them somehow.

If we go this way we need some script that fills a lot of key performance factors for js that can change with different hardware.

mikesherov:
@zxqfox, perhaps the way forward is to install the previous minor version of jscs and compare the current codebase against the latest release. This is then relative and immune to the "different hardware" problem.

zxqfox:
@mikesherov Some tagged version? Or just previous? I'm afraid of different set of functionality. Can be hard to compare. ;-(

markelog:
It was optional thing to do anyway

Issue #453, opened 2014-06-18
Improve error output


markelog:
Recent release of `1.5.0` <a href="https://github.com/mdevils/node-jscs/pull/366">showed</a>, on my opinion, that we need to improve error output, either provide warning module or adapt current error module API (rename maybe) to show notifications/warnings, without line and column number.

Also this change is required for #238 issue, just like it might beneficial for plugins and rules notification.

hzoo:
Any updates on this?

PR #1273, opened 2015-04-15
Don't BSOD on throws in rules

zxqfox:
See #1271 

I'm not sure about cases, I see now just one: we shouldn't stop checking if 1 rule failed in 1 file.

For that case I think we can notify user about that via common mechanism: errors object.

Your thoughts about message?

### DIFF
--- a/lib/string-checker.js
+++ b/lib/string-checker.js
@@ -130,7 +130,16 @@ StringChecker.prototype = {
 
         this._configuredRules.forEach(function(rule) {
             errors.setCurrentRule(rule.getOptionName());
-            rule.check(file, errors);
+
+            try {
+                rule.check(file, errors);
+            } catch (e) {
+                errors.add('Error running rule ' + rule.getOptionName() + ': ' +
+                    'This is an issue with JSCS and not your codebase.\n' +
+                    'Please file an issue (with the stack trace below) at: ' +
+                    'https://github.com/jscs-dev/node-jscs/issues/new\n' +
+                    e.stack, 1, 0);
+            }
         }, this);
 
         this._configuration.getUnsupportedRuleNames().forEach(function(rulename) {
--- a/test/specs/string-checker.js
+++ b/test/specs/string-checker.js
@@ -357,6 +357,33 @@ describe('modules/string-checker', function() {
         });
     });
 
+    describe('throwing rules', function() {
+        function _beforeEach(_verbose) {
+            checker = new StringChecker({verbose: _verbose});
+            // register rule that throw
+            checker.registerRule({
+                configure : function() {},
+                getOptionName: function() { return 'thrower'; },
+                check : function() {
+                    throw Error('Here we are!');
+                }
+            });
+            checker.configure({thrower: true});
+        }
+
+        it('should be handled internally with verbose', function() {
+            _beforeEach(true);
+
+            var errs = checker.checkString('var a');
+            assert.equal(errs.getErrorCount(), 1);
+
+            var err = errs.getErrorList()[0];
+            assert.equal(err.rule, 'thrower');
+            assert.ok(err.message.indexOf('Error running rule thrower:') !== -1);
+            assert.ok(err.message.indexOf('Error: Here we are!') !== -1);
+        });
+    });
+
     describe('presets', function() {
         testPreset('airbnb');
         testPreset('crockford');


test/specs/string-checker.js
```diff
@@ -357,6 +357,32 @@ describe('modules/string-checker', function() {
...
+            });
+        });
+
+        it('should be handled internally', function() {
+            var errs = checker.checkString('var a');
+            assert.equal(errs.getErrorCount(), 1);
+
+            var err = errs.getErrorList()[0];
+            assert.equal(err.rule, 'thrower');
+            assert.equal(err.message, 'Error: Here we are!');
```
TheSavior: If we are going to alert the user to the error here, I'd think we should get more specific.

> Error running rule 'thrower'. This is an issue with JSCS and not your codebase. 

and then if in verbose, provide the actual error information and stack trace.

zxqfox: Fixing. I'm not sure that we should add it as an usual codestyle error. Moreover, I'm sure that we shouldn't use usual `errors.add` here. Probably only with `warn` level?

zxqfox:
Folks, I need review here.

/cc @TheSavior @mikesherov @markelog @hzoo @mrjoelkemp

lib/errors.js
```diff
@@ -72,6 +78,25 @@ Errors.prototype = {
...
+        if (!this._verbose) {
+            return message;
+        }
+
+        if (typeof message !== 'object') {
+            return this._currentRule + ': ' + message;
+        }
+
+        return 'Error running rule `' + this._currentRule + '`: ' +
+            'This is an issue with JSCS and not your codebase.\n' +
```
markelog: "Please report it in..."?

lib/string-checker.js
```diff
@@ -130,7 +130,12 @@ StringChecker.prototype = {
 
         this._configuredRules.forEach(function(rule) {
             errors.setCurrentRule(rule.getOptionName());
-            rule.check(file, errors);
+
+            try {
```
markelog: I wonder how it would reflect on the performance

lib/string-checker.js
```diff
@@ -130,7 +130,12 @@ StringChecker.prototype = {
...
+                rule.check(file, errors);
+            } catch (e) {
+                errors.add(e, 1, 0);
```
markelog: How about you would form error message here (or in special helper) that way `Error#add` wouldn't have a polymorphic  signature and your intention would be clear and not spread across modules?

lib/string-checker.js
zxqfox: Idk why but `time mocha` execution time like 5% lower in branch than in master. Called several times... results the same.

`issues/1271`:
```
  3497 passing (17s)

real    0m18.497s
user    0m16.493s
sys     0m0.408s
```

`master`:
```
  3490 passing (20s)

real    0m21.552s
user    0m19.225s
sys     0m0.460s
```

lib/string-checker.js
zxqfox: I believe we need a single warn/error messaging mechanism. Typed errors can help us to have it.

I know that's pretty risky way but it's just a proposal and I think it's fine or very close to be.

lib/errors.js
zxqfox: :+1: I'm gonna add it as a separate line:
```
Error running rule `someThrowingRule: This is an issue with JSCS and not your codebase.
Please file an issue about it with stack trace in tracker: https://github.com/jscs-dev/node-jscs/issues/new 
```

Fine? Correct my english ;-)

lib/string-checker.js
markelog: > Typed errors can help us to have it with Errors object.

Not sure if i understand you, why exactly we need to pass it to another function?

lib/string-checker.js
markelog: 5% is a lot, but `try..catch` function, (like from <a href="https://github.com/petkaantonov/bluebird/wiki/Optimization-killers">here</a>)  probably wouldn't get us anywhere, it worth a try though

lib/errors.js
markelog: Fine by me, although i'm not sure why we always use " ` " sign

lib/string-checker.js
zxqfox: 1. Error is an abstract entity about a problem. We have standard notification algorithm (by passing string to errors.add) that notifies user about code style problems. Why we shouldn't use this algorithm to notify user about other troubles? I don't get why `errors.add` is a code style issues collector.

At other hand user shouldn't know by default about problems with code (TypeErrors, etc.). So I prefer to hide it under verbose flag.

I'd like to know what you thinking about current `Errors` object and errors, warnings, etc.
Would be great to know your view at that problem.

lib/string-checker.js
zxqfox: There is a workaround, btw. The main idea here is to move try-catch out of other code. Is it a real problem?

lib/errors.js
zxqfox: :smile: well, politics!

lib/string-checker.js
markelog: If that is only use-case for passing whole error object, i'd say we shouldn't complicate the `Errors#add`, just `<your error> + error.stack` is enough. 

Warnings would be nice to have for this case and for rule deprecation but it would mean changing reporters API.

lib/string-checker.js
markelog: > Is it a real problem?

On the small code-base no, on the large one, considering cumulative impact with other patches... might be, if you look on user perception - http://www.sitepoint.com/comparison-javascript-linting-tools#jscs

> Slowest of the four, but this is not a problem in typical use

Which might be a not fair conclusion, but a thing we should keep an eye on. "Slowest" - don't want to be categorized like that

zxqfox: Comparing typical linters with codestyle checker is like spoons to knifes. Yes, both written on nodejs, both checking your code, but jscs requires valid code and recommends using some linter before launching itself.

Each project that using JSCS using it with (eslint or jslint or jshint). Do you know a project where JSCS superseding an usual linter tool? I don't ;-)

markelog: I can debate it for a long time, which would lead nowhere, so i really don't want to argue about that, if you ask "if that is a real problem" - yes, i think it is, if we can do something about that, i would prefer to do it.

zxqfox: Anyway, code in this branch is faster than in master (about 5%). `try-catch` is almost made like @petkaantonov recommends. Still don't see a problem here ;-(

An article is not a real benchmark. Optimizations (that try-catch prevents) requires a lot of other things to happen. 

Personally I don't like try-catchs but I don't see another way here to catch an exception and to increase stability (and friendliness).

zxqfox: Btw, @markelog, we already have an issue about that: https://github.com/jscs-dev/node-jscs/issues/1043

markelog: > execution time like 5% lower in branch than in master

Hm, i interpret this phrase as this changes would add 5%. So you somehow make it faster? :anguished:

lib/string-checker.js
zxqfox: I don't see how we need to change API with custom Error (stringify-able) objects passed instead of plain string.

Atm we have one big Errors object with a list of messages and their locations (and some other flags).
I think we can drastically simplify that by making several Error types that will contain the same for code style errors (message, location, rule, fixed flag), and something else for other.

I'm not insist but just an idea.

Anyway, I don't now see how to make it work without hardcoding here output.

lib/string-checker.js
zxqfox: Time lower = Speed higher ;-)

V8 magic or something. Not sure how...

I suggest to talk about speed in #1043 since we all know that we need it and there are a lot of things we can speed up. If we've got troubles with perfomance because of this try-catch statement we could move it out into tryCatch helper function (as well as in bluebird) and I don't see here a problem.

lib/errors.js
zxqfox: Added url and dropped ticks

lib/errors.js
```diff
@@ -72,6 +78,27 @@ Errors.prototype = {
...
+            return message;
+        }
+
+        if (typeof message !== 'object') {
+            return this._currentRule + ': ' + message;
+        }
+
+        return 'Error running rule ' + this._currentRule + ': ' +
+            'This is an issue with JSCS and not your codebase.\n' +
+            'Please file an issue about it with stack trace in tracker: ' +
```
hzoo: Maybe it could say: `Please file an issue (with the stack trace below) at\n https://github.com/jscs-dev/node-jscs/issues/new `

Or `about it with the stack trace below in our tracker: `

zxqfox: Fixed, thanks!

lib/string-checker.js
markelog: I was pushing for the opmization because i was under impression we would take hit. If there is nothing hurting the performance then its all good by me.

#1043 is not about perf improvements by about perf monitoring, we would need address every bottleneck on case-by-case basis

zxqfox: Totally agreed. :palm_tree: 

Still perf monitoring will allow us to make step only forward in performance.

lib/string-checker.js
markelog: If you talking about other use-cases, i guess it would better to discuss in separate issue, it probably would require proof of concept.

> Anyway, I don't now see how to make it work without hardcoding here output.

You either "hardcode" it here or in `errors` module, but doing it here we would clearly see the genesis of error and would not complicate `add` method.

zxqfox: Yup, but there we can nicely shut these errors. Vice-versa. ;-(

markelog: What preventing you to check `verbose` flag here?

zxqfox: Well, nothing. Problem is a message will go not to reporter but directly to stdout.

markelog: Why?

zxqfox: Because of nature of `console.log`. You know another way to do it?

markelog:
```js
error.add( "message" + e.stack, 1, 0);
```

zxqfox: Okay. This will mix code style issues with internal throws. And if something goes wrong in rules (we will catch an exception or something) it will report as a code style check not passed. Is it fair for user?

mikesherov: Yes, it's fair to the user. Better they see the error reported rather than it being possibly swallowed by their IDE.

lib/errors.js
```diff
@@ -72,6 +78,27 @@ Errors.prototype = {
     },
 
     /**
+     * Formats message according to message data and _verbose flag
+     *
+     * @param {String|Error} message
+     */
+    _prepareMessage: function(message) {
+        if (!this._verbose) {
```
mikesherov: I think this should happen regardless of verbose flag. Reporting issues is not something that should be dictated by a verbose flag or not.

lib/errors.js
```diff
@@ -61,10 +61,16 @@ Errors.prototype = {
             return;
         }
 
+        var message = errorInfo.message;
+        if (!(this._verbose || typeof message !== 'object')) {
+            // uhm?
```
mikesherov: when is this case hit?

zxqfox: If we pass an Error object here.

lib/errors.js
zxqfox: Internal error in a rule is not an issue for a user's code.

lib/string-checker.js
zxqfox: Better to report only what user wants (and not our bugs).

markelog: Yeah, no better way to do it, i'm not sure why asking though, in this pull you're proposing to do exact same thing, just in another module.

markelog: btw, we're already done it for deprecated rules and for syntax errors provided by the esprima, i believe other linters do the same

zxqfox: Can we try to make "warnings" mechanism to report internal issues without failing code style check?

This PR is not critical anyway. User now gets a huge amount of fun, and to reduce it we need to hardcode somewhere a try-catch.

zxqfox: Oh guys. You are so strong in your vision on it ;-)

Making what you want.

markelog: > Can we try to make "warnings" mechanism to report internal issues without failing code style check?

We can try - https://github.com/jscs-dev/node-jscs/issues/453, i'm not sure it would better though

lib/string-checker.js
```diff
@@ -130,7 +130,16 @@ StringChecker.prototype = {
 
         this._configuredRules.forEach(function(rule) {
             errors.setCurrentRule(rule.getOptionName());
-            rule.check(file, errors);
+
+            try {
+                rule.check(file, errors);
+            } catch (e) {
+                errors.add('Error running rule ' + rule.getOptionName() + ': ' +
```
markelog: How it look in the console, could post a screenshot?

zxqfox: image

markelog:
LGTM

mikesherov:
:+1:

PR #1278, opened 2015-04-17
requireDollarBeforejQueryAssignment: Add ignoreProperties rule value

### LINKED ISSUE #1266: jQuery identifiers must start with a $ - failing for object properties
Using JSCS/gulp on a project which assigns a jquery object to an object property triggers the `jQuery identifiers must start with a $` error. However, the property is part of a configuration object for another javascript in bower, and outside our control.

```
jQuery identifiers must start with a $ at assets/js/controllers/project/fields.js :
    24 |        $scope.initFormBuilder = function() {
    25 |            $scope.formbuilder = new Formbuilder({
    26 |                targets: $('.formbuilder'),
------------------------^
```

Assigning it to a variable first solves the issue, but is unnecessary in this specific case.

Using gulp-jscs 1.4.0 on Mac OS X 10.10.2.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11092674-jquery-identifiers-must-start-with-a-failing-for-object-properties?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zxqfox:
Fixes #1266

### DIFF
--- a/lib/rules/require-dollar-before-jquery-assignment.js
+++ b/lib/rules/require-dollar-before-jquery-assignment.js
@@ -1,10 +1,9 @@
 /**
  * Require a $ before variable names that are jquery assignments.
  *
- * Type: `Boolean`
+ * Types: `Boolean` or `String`
  *
- * Value: `true`
- *
+ * Values: `true` or `"ignoreProperties"`
  *
  * #### Example
  *
@@ -12,13 +11,34 @@
  * "requireDollarBeforejQueryAssignment": true
  * ```
  *
- * ##### Valid
+ * ##### Valid example for mode `true`
  *
  * ```js
  * var $x = $(".foo");
+ * var y = {
+ *   $x: $(".bar")
+ * };
  * ```
  *
- * ##### Invalid
+ * ##### Invalid example for mode `true`
+ *
+ * ```js
+ * var x = $(".foo");
+ * var y = {
+ *   x: $(".bar")
+ * };
+ * ```
+ *
+ * ##### Valid example for mode `ignoreProperties`
+ *
+ * ```js
+ * var $x = $(".foo");
+ * var y = {
+ *   x: $(".bar")
+ * };
+ * ```
+ *
+ * ##### Invalid example for mode `ignoreProperties`
  *
  * ```js
  * var x = $(".foo");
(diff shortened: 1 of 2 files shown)


zxqfox:
Feels like what we need. Code is a little bit complicated but looks like we need to left just variables check with `ignoreProperties`. Am I right?

/cc @viveleroi

TheSavior:
Frankly, I'm surprised this works. Mostly because I would think we shouldn't return if ignoreProperties is true, but only if we are inside of an object *and* ignoreProperties is true. 

However, I couldn't find a test that previously had an error, that should still have an error with this relaxing modification, so maybe it is okay.

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -53,6 +79,10 @@ module.exports.prototype = {
                 left = token.id;
                 varName = left.name;
                 right = token.init;
+
+            } else if (ignoreProperties) {
+                return;
+
```
zxqfox: @markelog @mikesherov What our code style said about padding lines about `else if`s?

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -61,6 +91,7 @@ module.exports.prototype = {
 
                 varName = left.name || left.property.name;
                 right = token.right;
+
```
zxqfox: random whitespace.

markelog:
Inconsistency in rule values is frustrating though

zxqfox:
@markelog Let's fix it! I've used `ignoreProperties` because it was suggested here https://github.com/jscs-dev/node-jscs/issues/1266#issuecomment-92838858

Any ideas?

markelog:
`allExcept` comes to mind, but this would not save us, since we already use `ignoreProperties` in another rule.

2.0 should save us all

zxqfox:
@markelog Made an issue for that: #1303

PR #1281, opened 2015-04-17
Verbose option and dependent parts

### LINKED ISSUE #1193: Add "verbose" option as a config option
Would be easier for some people and would be preferable for the grunt plugin

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10052707-add-verbose-option-as-a-config-option?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
* Add "verbose" option as a config option

* Do not show rule for "Unsupported rule" error 

* Deprecate params for "Checker" constructor

Fixes #1193

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -19,10 +19,7 @@ module.exports = function(program) {
     var checkerPromise;
     var defer = Vow.defer();
     var promise = defer.promise();
-    var checker = new Checker({
-        verbose: program.verbose,
-        esnext: program.esnext
-    });
+    var checker = new Checker();
     var args = program.args;
     var returnArgs = {
         checker: checker,
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -64,11 +64,25 @@ Errors.prototype = {
         this._errorList.push({
             filename: this._file.getFilename(),
             rule: this._currentRule,
-            message: this._verbose ? this._currentRule + ': ' + errorInfo.message : errorInfo.message,
+            message: this._prepareMessage(errorInfo),
             line: errorInfo.line,
             column: errorInfo.column,
             fixed: errorInfo.fixed
         });
+    },
+
+    /**
+     * Prepare error message.
+     *
+     * @param {Object} errorInfo
+     * @private
+     */
+    _prepareMessage: function(errorInfo) {
+        if (this._verbose && this._currentRule) {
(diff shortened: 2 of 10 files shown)


mrjoelkemp:
Update readme to remove it as a cli option? 

This is possibly considered a breaking change if folks rely on the output of the CLI using verbose in flag form.

markelog:
> Update readme to remove it as a cli option?

You can still override it though console, need to update docs though

lib/cli.js
```diff
@@ -19,10 +19,7 @@ module.exports = function(program) {
     var checkerPromise;
     var defer = Vow.defer();
     var promise = defer.promise();
-    var checker = new Checker({
-        verbose: program.verbose,
-        esnext: program.esnext
```
zxqfox: Unused esnext?

lib/errors.js
```diff
@@ -72,6 +72,20 @@ Errors.prototype = {
     },
 
     /**
+     * Prepare error message.
+     *
+     * @param {Object} errorInfo
+     * @private
+     */
+    _prepareMessage: function(errorInfo) {
+        if (this._verbose && this._currentRule) {
```
zxqfox: Not sure it's a good idea ;-(

zxqfox:
:+1: at all.

lib/errors.js
markelog: Meaning?

lib/cli.js
markelog: Could you elaborate?

markelog:
> Is it breaking change?

No

lib/cli.js
zxqfox: Didn't see where you readd it to checker. Suggesting that it wasn't used in code.

lib/errors.js
zxqfox: Just thinking about what Errors object should do. Reporters can format this kind of errors more wisely.
Nvm, it's another story.

mikesherov:
:+1:

PR #1291, opened 2015-04-21
requireOperatorBeforeLineBreak: autofix by moving operator only

### LINKED ISSUE #1288: autofix creates long lines
```js
if (thisIsAReallyLongNameOhYesItIs
    && thisIsAReallyLongNameOhYesItIs2
    && thisIsAReallyLongNameOhYesItIs3
    && thisIsAReallyLongNameOhYesItIs4) {
  alert(1)
}
```

Run:
```jscs --preset=google -x test.js```

Expected:
jscs will put the operator on the previous line

Actual:
jscs merges all the lines together, creating one really long line and creating a new warning

```js
if (thisIsAReallyLongNameOhYesItIs && thisIsAReallyLongNameOhYesItIs2 && thisIsAReallyLongNameOhYesItIs3 && thisIsAReallyLongNameOhYesItIs4) {
  alert(1)
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11643355-autofix-creates-long-lines?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nicks:
Fixes #1288

### DIFF
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -230,6 +230,7 @@ TokenAssert.prototype._setEmptyLineIndentation = function(lineNumber, newWhitesp
  *
  * @param {Object} options.token
  * @param {Object} options.nextToken
+ * @param {Boolean} [options.stickToPreviousToken]
  * @param {String} [options.message]
  */
 TokenAssert.prototype.sameLine = function(options) {
@@ -261,6 +262,8 @@ TokenAssert.prototype.differentLine = function(options) {
  * @param {Object} [options.atLeast] At least how many lines the tokens are apart
  * @param {Object} [options.atMost] At most how many lines the tokens are apart
  * @param {Object} [options.exactly] Exactly how many lines the tokens are apart
+ * @param {Boolean} [options.stickToPreviousToken] When auto-fixing stick the
+ *     nextToken onto the previous token.
  */
 TokenAssert.prototype.linesBetween = function(options) {
     var token = options.token;
@@ -298,7 +301,7 @@ TokenAssert.prototype.linesBetween = function(options) {
         }
 
         if (fixed) {
-            this._augmentLineCount(nextToken, lineCount);
+            this._augmentLineCount(options, lineCount);
         }
 
         this.emit('error', {
@@ -358,11 +361,17 @@ TokenAssert.prototype._validateOptions = function(options) {
 /**
  * Augments token whitespace to contain the correct number of newlines while preserving indentation
  *
- * @param {Object} token
+ * @param {Object} options.nextToken
+ * @param {Boolean} [options.stickToPreviousToken]
  * @param {Number} lineCount
  */
-TokenAssert.prototype._augmentLineCount = function(token, lineCount) {
+TokenAssert.prototype._augmentLineCount = function(options, lineCount) {
+    var token = options.nextToken;
     if (lineCount === 0) {
+        if (options.stickToPreviousToken) {
+            var nextToken = this._file.getNextToken(token, {includeComments: true});
+            nextToken.whitespaceBefore = token.whitespaceBefore;
+        }
         token.whitespaceBefore = ' ';
         return;
     }
--- a/test/specs/rules/require-operator-before-line-break.js
+++ b/test/specs/rules/require-operator-before-line-break.js
@@ -1,6 +1,7 @@
 var Checker = require('../../../lib/checker');
 var assert = require('assert');
 var operators = require('../../../lib/utils').binaryOperators.slice();
+var reportAndFix = require('../../lib/assertHelpers').reportAndFix;
 
 describe('rules/require-operator-before-line-break', function() {
     var checker;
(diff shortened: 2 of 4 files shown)


lib/token-assert.js
```diff
@@ -230,6 +230,7 @@ TokenAssert.prototype._setEmptyLineIndentation = function(lineNumber, newWhitesp
  *
  * @param {Object} options.token
  * @param {Object} options.nextToken
+ * @param {boolean} [options.moveWhitespaceToNextToken]
```
nicks: i'm unhappy with the name of this option and how it fits into the overall api. I'm open to better suggestions.

zxqfox: Did you think about passing fixing callbacks?

zxqfox: I'll just left it here:
```
var a, b, c = 1;
// vs
var a, b,
    c = 2;
// vs
var a, b;
var c = 1;
// vs
var a;
var b;
var c = 1;
```

test/specs/rules/require-operator-before-line-break.js
```diff
@@ -65,4 +66,21 @@ describe('rules/require-operator-before-line-break', function() {
...
+        rules: { requireOperatorBeforeLineBreak: true },
+        input: [
+          'if (x',
+          '    && y) {',
+          '  alert(z);',
+          '}'
+        ].join('\n'),
+        output: [
+          'if (x &&',
+          '    y) {',
```
zxqfox: I'm pretty sure we also need to check cases with single and multiline comments.

lib/token-assert.js
nicks: i like the idea of using a callback. I updated the PR to do that instead!

re: different line-wrapping options: yeah, the test deliberately tickles that question. My general philosophy is that if the user put a line break there, we should assume there's a good reason for there to be a line-break in there somewhere.

You could imagine a future jscs rule like "minimizeLineBreaksInExpressions" that warned you if a line-break in an expression was unnecessary, and autofixed it. This would make more sense to me as an orthogonal rule, rather than having the individual operator rules make this decision.

test/specs/rules/require-operator-before-line-break.js
nicks: not sure what you had in mind, but i added a test case with a comment?

zxqfox:
```
          'if (x // something',
          '    && y) {',
          '  alert(z);',
// and
          'if (x /* something */',
          '    && y) {',
          '  alert(z);',
// and
          'if (x',
          '    /* something */ && y) {',
          '  alert(z);',
```

zxqfox:
Thanks for you work here ;-)

test/specs/rules/require-operator-before-line-break.js
nicks: added!

zxqfox: Please add cases with `/* comment before ampersands */ &&` and `x // line comment between var and ampersand`

nicks: i added them above with `should not autofix line comment on first line`? did you not see them? or are those insufficient for some reason?

zxqfox: Oh, I don't see them ;-)

mikesherov:
Thanks again for contributing! @zxqfox I'm not a huge fan of having to pass a callback. That puts the responsibility on individual rules to have to reimplement the same mechanics which should be abstracted away by the assertion API.

zxqfox:
@mikesherov I don't insist on callbacks (actually, I'm against them for common cases) but we should think about very hard cases.

If you have some good solution for this case — it's great! But I still don't see how to do it without callback ;-(

I think we should harvest common cases (as well as we did with assertion API some time ago) and then reinvent this Fixing API. Right?

But, if you have any ideas — I'd like to discuss, I don't like callbacks too ;-)

zxqfox:
> re: different line-wrapping options: yeah, the test deliberately tickles that question. My general philosophy is that if the user put a line break there, we should assume there's a good reason for there to be a line-break in there somewhere.

Probably we can think here about how to find a good place for these newlines at autofix stage. But there are many edge cases as I see. ;-(
 
> You could imagine a future jscs rule like "minimizeLineBreaksInExpressions" that warned you if a line-break in an expression was unnecessary, and autofixed it. This would make more sense to me as an orthogonal rule, rather than having the individual operator rules make this decision.

:+1: Nice idea ;-)

mdevils:
I believe we should not introduce callback here. It would be better to start adding options about how to fix this error. What do you think?

zxqfox:
@mdevils Waiting for ideas ;-)

zxqfox:
Related #1296

nicks:
To recap: I originally had `{boolean} [options.moveWhitespaceToNextToken]` and we agreed that didn't feel great. I changed it to `options.fixCallback`. I agree the callback is not optimal either, but it seemed like a good way to move forward without committing to an API.

I'm happy to change it to whatever you all think is best. But I'm new to this codebase, and don't have a good sense of what's best.

zxqfox:
That's my fault, I suggest to use callback here. It's because I don't think we can find a good way to handle a couple of cases. But still, we have a hope ;-)

mdevils:
@zxqfox, in #1296 you just need to add a rule which requires single space after `:`. Should be enough?

mdevils:
I think having an option `{boolean} [options.stickToPreviousToken]` is fine. Or isn't it?

zxqfox:
@mdevils As I feel it should be resolved by moving tokens thru whitespaces (without adding extra spaces if possible).

stickToPrevious/stickToNext/keepNewline?

mdevils:
> @mdevils As I feel it should be resolved by moving tokens thru whitespaces (without adding extra spaces if possible).

Right now we don't have a good way to move tokens :neutral_face:

zxqfox:
@mdevils Yup. Should it be like `api.moveTokenThruWhitespaces(target, stickToToken)`?

`x\n    /* foo */ && y\n`
→ `x && \n    /* foo */y\n` or, better:
→ `x &&\n    /* foo */ y\n`

mdevils:
@zxqfox, The problem is not about missing method, but about missing concept. I believe this should be discussed separately.

zxqfox:
@mdevils Yup, I just suggested a way to move them. On the top of it we can try to think about providing several options via assert-api for fixing (like `stickToPreviousToken`, etc.)

nicks:
great! thanks for the feedback. I changed it to stickToPreviousToken.

mdevils:
Looks good to me :+1:

zxqfox:
:+1: 

> Coverage decreased (-0.06%)

But... ;-) Need tests for TokenAssert

zxqfox:
@nicks :+1: Great, thank you, Nick!

PR #1292, opened 2015-04-21
Preset: Mozilla

SimenB:
My first try at making a preset, hopefully not too bad. Is there any rule about the order of the rules in the preset? Alphabetic, some grouping, etc.

Style-guide: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style

The Java part mentions `One declaration per line`. Should `requireMultipleVarDecl` be included?

It's mentioned that `case` that declares variables should have braces, but not others. Any rule fitting for that? In JS not so relevant, maybe (unless using `let`)?

K&R bracing style, any way to differentiate between class/function declarations and control structures? Like options to `requireNewlineBeforeBlockStatements`

> Break long conditions after `&&` and `||`

Any way to assert that?

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -718,6 +718,9 @@ Configuration.prototype.registerDefaultPresets = function() {
     // https://github.com/felixge/node-style-guide#nodejs-style-guide
     this.registerPreset('node-style-guide', require('../../presets/node-style-guide.json'));
 
+    // https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#JavaScript_practices
+    this.registerPreset('mozilla', require('../../presets/mozilla.json'));
+
     // https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript
     this.registerPreset('wikimedia', require('../../presets/wikimedia.json'));
 
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -10,6 +10,7 @@ Note: the easiest way to use a preset is with the [preset](#preset) option descr
  * [Grunt](https://github.com/jscs-dev/node-jscs/blob/master/presets/grunt.json) — http://gruntjs.com/contributing#syntax
  * [jQuery](https://github.com/jscs-dev/node-jscs/blob/master/presets/jquery.json) — https://contribute.jquery.org/style-guide/js/
  * [MDCS](https://github.com/jscs-dev/node-jscs/blob/master/presets/mdcs.json) — [https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style™](https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style%E2%84%A2)
+ * [Mozilla](https://github.com/jscs-dev/node-jscs/blob/master/presets/mozilla.json) — https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style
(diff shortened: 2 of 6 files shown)


lib/config/configuration.js
```diff
@@ -718,6 +718,9 @@ Configuration.prototype.registerDefaultPresets = function() {
     // https://github.com/felixge/node-style-guide#nodejs-style-guide
     this.registerPreset('node-style-guide', require('../../presets/node-style-guide.json'));
 
+    // https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#JavaScript_practices
```
SimenB: Not sure if linking directly to `#JavaScript_practices` is necessary, as most rules comes from the `#Naming_and_Formatting_code`-part

zxqfox:
Thanks! Personally, I like it. But looks like too early to merge, a lot of parts of code are not covered. Is it possible to fix it?

/cc @markelog 

p.s. note that we have a non-verbal rule about presets in core: they should be used by a lot of people. I hope we'll find these guys ;-)

markelog:
I would like somewhere from mozilla verify this preset, providing real code-example would be preferable too.

SimenB:
@markelog I'm not sure where to find good examples of real code. https://github.com/mozilla/metrics-graphics/ seems to be amongst the most popular JS repos Mozilla's got, but they use single quotes, even though the style guide says not too, as a quick example.

@jrburke Could you, or someone else from Mozilla, give the rules a look, and maybe provide a project with some source that can be used to verify the rules against?

hzoo:
I like alphabetizing the rules although I don't think that's consistently done.

I did see an ES6 style guide for [mozilla/addon-sdk](https://github.com/mozilla/addon-sdk/wiki/Coding-style-guide).

SimenB:
@hzoo That style guide also says to use double quotes, while the source files in the same repo use single.

markelog:
@tofumatt care to comment? Is there any projects following mozilla code-style you know?

tofumatt:
I would think these code styles apply to, if anything, JavaScript inside Firefox (devtools). But looking at the history of that page, the JS styles haven't been updated in years.

The bigger point here is there is certainly no *one Mozilla style* regarding JS. There seem to be at least two: that used by "webdev" (tends to be more clientside and node) and that used by "Firefox" (tends to be node-style but mostly used inside Firefox).

For instance, these styles wouldn't fit with our recent [node-firefox](https://github.com/mozilla/node-firefox) project, which has its own [style guide](https://github.com/mozilla/node-firefox-build-tools/blob/master/.jscsrc). Nor would it match [localForage](https://github.com/mozilla/localForage)'s guide.

I don't think it would be accurate to say there's a "Mozilla" style for JS; we generally just define our style guide in our own rules and all of the projects I've worked on in the past few years include linters/style-checkers in the test suite. At any rate, I don't think it's these styles.

Maybe someone from devtools (@ednapiranha?) could chime in, but I don't think I would call this the "Mozilla" style nor do I think one is useful.

SimenB:
This can probably just be closed then :smile:

zxqfox:
@SimenB Or renamed to `firefox`/`something-else`?

SimenB:
Well sure, but if nobody really uses it at Mozilla, nobody else will probably use it either. Just bloats the presets. Right?

markelog:
@SimenB nevertheless, thanks you for the discussion.
@tofumatt thank you for the reply.
@zxqfox let's be conservative here, if community doesn't need it, we shouldn't force it.

zxqfox:
@markelog :+1: okey ;-)

Issue #768, opened 2014-11-13
requireMultipleVarDecl, "onevar", and "require" statements


indexzero:
Here is something that I know a lot of Node programmers do:

``` js
var fs = require('fs'),
    path = require('path');

var rootdir = process.env.HOME || process.env.USERPROFILE,
    configdir = path.join(rootdir, '.configfolder'); 
```

So there are two (and only two) var blocks. One for `require` statements **only** and another for everything else. It is only valid in the top-level scope of a file. My proposal would be 

```
"requireMultipleVarDecl": "onevarExceptRequire"
```

PR #1310, opened 2015-04-28
disallowMultipleVarDecl: add exception for require statements

stefanosala:
This PR adds a `exceptRequire` option for `disallowMultipleVarDecl` to allow multiple var declarations only when requiring external lib, tipically at the beginning of a nodejs module.

### DIFF
--- a/lib/rules/disallow-multiple-var-decl.js
+++ b/lib/rules/disallow-multiple-var-decl.js
@@ -1,13 +1,16 @@
 /**
  * Disallows multiple `var` declaration (except for-loop).
  *
- * Types: `Boolean` or `String`
+ * Types: `Boolean` or `Object`
  *
  * Values:
  *
  * - `true` disallows multiple variable declarations except within a for loop
- * - `'strict'` disallows all multiple variable declarations
- * - `'exceptUndefined'` allows declarations where all variables are not defined
+ * - `Object`:
+ *    - `'strict'` disallows all multiple variable declarations
+ *    - `'allExcept'` array of exceptions:
+ *       - `'undefined'` allows declarations where all variables are not defined
+ *       - `'required'` allows declarations where all variables are importing external modules with require
  *
  * #### Example
  *
@@ -24,17 +27,28 @@
  * for (var i = 0, j = arr.length; i < j; i++) {}
  * ```
  *
- * ##### Valid for `strict`
+ * ##### Valid for `{ strict: true }`
  *
  * ```js
  * var x = 1;
  * var y = 2;
  * ```
  *
- * ##### Valid for `exceptUndefined`
+ * ##### Valid for `{ allExcept: ['undefined'] }`
  *
  * ```js
  * var a, b;
+ * var x = 1;
+ * var y = 2;
+ *
+ * for (var i = 0, j = arr.length; i < j; i++) {}
+ * ```
+ * ##### Valid for `{ allExcept: ['require'] }`
+ *
+ * ```js
+ * var a = require('a'),
+ *     b = require('b');
+ *
  * var x = 1;
  * var y = 2;
  *
@@ -58,15 +72,34 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(options) {
-        assert(
-            options === true ||
-            options === 'strict' ||
-            options === 'exceptUndefined',
-            this.getOptionName() + ' option requires a true value, "strict", or "exceptUndefined"'
-        );
+        // support for legacy options
+        if (typeof options !== 'object') {
+            assert(
+                options === true ||
+                options === 'strict' ||
+                options === 'exceptUndefined',
+                this.getOptionName() +
+                    ' option requires a true value, "strict", "exceptUndefined", or an object'
+            );
 
-        this.strictMode = options === 'strict';
-        this.exceptUndefined = options === 'exceptUndefined';
+            var _options = {
+                strict: options === 'strict',
+                allExcept: []
+            };
+
+            if (options === 'exceptUndefined') {
+                _options.allExcept.push('undefined');
+            }
+
+            return this.configure(_options);
+        }
+
+        if (Array.isArray(options.allExcept)) {
+            this._exceptUndefined = options.allExcept.indexOf('undefined') > -1;
+            this._exceptRequire = options.allExcept.indexOf('require') > -1;
+        }
+
+        this._strictMode = options.strict === true;
     },
 
     getOptionName: function() {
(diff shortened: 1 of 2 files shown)


zxqfox:
Do we need an opposite?

zxqfox:
@stefanosala Thanks for you work here ;-)

stefanosala:
I'm not sure we would really need an opposite, but maybe I'm too focused on our internal workflow. If that's required, I can implement it easily, just let me know.

hzoo:
The issue would be for #768 (for tracking)

zxqfox:
Oh, right. :+1: 

Thanks, @hzoo

stefanosala:
Yes, that was my starting point but it was mentioning `requireMultipleVarDecl`, so it was a bit confusing :) anyway, thanks.

hzoo:
Well in that case I guess we might want the opposite haha.

stefanosala:
I'll add it in a separate PR.

stefanosala:
@zxqfox news on this? :)

mrjoelkemp:
Great work here. 

In terms of the option naming, what about using  allExcept: ['require'] to adhere to our rule configuration style?

stefanosala:
@mrjoelkemp makes sense! Is this what you mean? I didn't refactor too much the code, just added the allExcept option and kept the rest. How would you handle deprecation of the old parameters?

mrjoelkemp:
Crap, didn't catch that `exceptUndefined` was an existing value. We'd have to keep the `exceptUndefined` but squeeze in the `allExcept: ['require']`. Otherwise, it's a breaking change that would only allow this to be shipped in 2.0. Sorry about that.

Does this complicate the implementation or force you to backpedal too much?

stefanosala:
I already added the support for both styles (`exceptUndefined` or `exceptRequired` and `allExcept: ['require', 'undefined']` :) I think this is the way to go, but I'm open to suggestions or requests.

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -76,22 +98,53 @@ module.exports.prototype = {
...
+                return;
+            }
+
+            // allow multiple var declaration with all require
+            // var a = require("a"), b = require("b")
+            if (exceptRequire && allRequireStatements) {
+                return;
+            }
+
+            // allow multiple var declarations only with require && undefined
```
mrjoelkemp: Love these comments! <3

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -58,15 +69,26 @@ module.exports = function() {};
...
+                this.exceptRequire = options.allExcept.indexOf('require') > -1;
+            }
+
+            this.strictMode = options.strict === true;
+        } else {
+            assert(
+                options === true ||
+                options === 'strict' ||
+                options === 'exceptUndefined' ||
+                options === 'exceptRequire',
```
mrjoelkemp: I'd prefer if this new option was not supported. If users want to exclude `require`s, then they should use the `allExcept` form (which would also help them migrate to 2.0).

mrjoelkemp:
LGTM. @hzoo and @zxqfox, can you guys review?

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -58,15 +69,26 @@ module.exports = function() {};
...
+                this.getOptionName() + ' option requires a true value, "strict", "exceptUndefined", or "exceptRequire"'
```
hzoo: We'll need to change the message to include: an object `allExcept` or something.

lib/rules/disallow-multiple-var-decl.js
hzoo: So we would want to keep `exceptUndefined` for backward compatibility but not add `exceptRequire` and use `allExcept` for 2.0 purposes?

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -76,22 +99,53 @@ module.exports.prototype = {
...
+            // var a = require("a"), b = require("b")
+            if (exceptRequire && allRequireStatements) {
+                return;
+            }
+
+            // allow multiple var declarations only with require && undefined
+            // var a = require("a"), b = require("b"), x, y
+            if (exceptUndefined && exceptRequire &&
+                (definedVariables.length - requireStatements.length) === 0 &&
+                (definedVariables.length + requireStatements.length) === node.declarations.length) {
```
hzoo: Not a big deal, but you can just do `definedVariables.length === requireStatements.length` here right?

stefanosala: Good catch, there was a bug in the previous code.

lib/rules/disallow-multiple-var-decl.js
mrjoelkemp: Yes. Exactly.

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -58,15 +69,25 @@ module.exports = function() {};
...
-            options === 'strict' ||
-            options === 'exceptUndefined',
-            this.getOptionName() + ' option requires a true value, "strict", or "exceptUndefined"'
-        );
-
-        this.strictMode = options === 'strict';
-        this.exceptUndefined = options === 'exceptUndefined';
+        if (typeof options === 'object') {
+            if (Array.isArray(options.allExcept)) {
+                this.exceptUndefined = options.allExcept.indexOf('undefined') > -1;
```
zxqfox: Since we making here a lot of rework I'd add `_` to protected properties.

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -58,15 +69,25 @@ module.exports = function() {};
...
+                this.exceptRequire = options.allExcept.indexOf('require') > -1;
+            }
+
+            this.strictMode = options.strict === true;
+        } else {
```
zxqfox: Non-blocker:
Not sure about this `else`-block. We can add a return in `if`. Personally, I'd swap these two blocks because `options` as object is not an exception in common.

zxqfox:
I've left 2 more non-critical notes. LGTM, thanks!

stefanosala:
@zxqfox thanks for the review, I just pushed a couple of commits that should solve all the issues. Please let me know if everything is ok and I'll be happy to squash the commits.

zxqfox:
@stefanosala Thanks! LGTM

stefanosala:
:+1: squashed!

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -40,6 +40,17 @@
  *
  * for (var i = 0, j = arr.length; i < j; i++) {}
```
zxqfox: Ouch, sorry. One more nit: Need to update README.

https://github.com/Receiptful/node-jscs/blob/feature/except-require/lib/rules/disallow-multiple-var-decl.js#L4-L10

stefanosala: Right! I updated it with the new syntax.

lib/rules/disallow-multiple-var-decl.js
```diff
@@ -1,21 +1,23 @@
 /**
  * Disallows multiple `var` declaration (except for-loop).
  *
- * Types: `Boolean` or `String`
+ * Types: `Object`
```
zxqfox: Oh, I think it's too early to deprecate `true` value ;-(

Thoughts?

stefanosala: Well, you can still use it if you want, but the doc now says you can already start to use the new syntax. I'm open to changing it back anyway if you want :)

zxqfox: /cc @hzoo @mrjoelkemp

zxqfox: @stefanosala Yeah, just we should be more implicit. We have a lot of rules with true value with default behaviour (so I'm not sure we should deprecate it), but `string` values is deprecated and we can note it right now.

We have an issue about new format for 2.0 where we going to rework all rules (that's why we trying to prepare them to object form).

stefanosala: :+1: I got your concern and changed the doc accordingly.

test/specs/rules/disallow-multiple-var-decl.js
```diff
@@ -79,4 +79,52 @@ describe('rules/disallow-multiple-var-decl', function() {
             assert(checker.checkString('var x, y = 2, z;').getErrorCount() === 1);
         });
     });
+
+    describe('exceptRequire', function() {
+        beforeEach(function() {
+            checker.configure({ disallowMultipleVarDecl: { allExcept: ['require'] } });
+        });
+        it('should not report mutiple var decls with require', function() {
```
mrjoelkemp: `multiple`

test/specs/rules/disallow-multiple-var-decl.js
```diff
@@ -79,4 +79,52 @@ describe('rules/disallow-multiple-var-decl', function() {
...
+            assert(checker.checkString('var first = require("first"), second = require("second");').isEmpty());
+        });
+        it('should report mutiple var decls with require mixed with normal', function() {
```
mrjoelkemp: `multiple`

test/specs/rules/disallow-multiple-var-decl.js
```diff
@@ -79,4 +79,52 @@ describe('rules/disallow-multiple-var-decl', function() {
...
+            assert.equal(1, checker.checkString('var first = require("first"), second = 1;').getErrorCount());
+        });
+        it('should report mutiple var decls', function() {
```
mrjoelkemp: `multiple`

mrjoelkemp:
Feel free to squash and I'll land this @stefanosala. Really great work here! Thanks for contributing.

stefanosala:
Yay, thanks! I'll update #768 with similar changes.

mrjoelkemp:
Landed! Woohoo!

zxqfox:
:smirk_cat:

PR #1314, opened 2015-04-28
disallowMultipleSpaces: fixed configuration error message

fluidsonic:
Fixed wrong & confusing error message when `disallowMultipleSpaces` option has an invalid value.


markelog:
Thank you!

PR #1316, opened 2015-04-28
Fix missing quotes in a docstring example

astorije:


mrjoelkemp:
Thanks for contributing @astorije

astorije:
Argh, sorry for the wrong commit message!

*@astorije slaps himself for not reading the CONTRIBUTING file...*

mrjoelkemp:
No worries at all. We don't usually get contributions that abide by the format, so maintainers usually change the commit message before merging.

Thanks again for the help.

PR #1317, opened 2015-04-28
Update mixup between rules in docstring example

astorije:
`requirePaddingNewLinesBeforeExport` was mentioned instead of `disallowPaddingNewLinesBeforeExport`.


mrjoelkemp:
Thanks @astorije!

Issue #1198, opened 2015-03-26
Difference between disallowSpacesInsideArrayBrackets and disallowSpacesInsideBrackets?


TheSavior:
I can't tell the difference between these two rules. `arrayBrackets` seems to comment on a rule that is deprecated that `brackets` still uses. Does that imply that the rule was renamed and we somehow ended up with a copy?

hzoo:
I think this was part of this discussion: https://github.com/jscs-dev/node-jscs/issues/923. I think `requireSpacesInsideArrayBrackets` is only on arrays and disallowSpacesInsideBrackets is for all `[]`?

fluidsonic:
These rules seem to be broken at the moment.

We use the configuration below to require this format:
```js
var a = [ 1, 2 ]  // spaces required in array creation
var b = a[0]      // spaces forbidden for array/object access
```
In the first line, jscs both requires and forbids the use of spaces so there's no way to validate the code.

`.jscsrc`:
```json
{
	"disallowSpacesInsideBrackets": true,
	"requireSpacesInsideArrayBrackets": "all"
}
```

`test.js`:
```js
'use strict'

var a = [ 1, 2 ]
var b = [1, 2]
```

Output:
```
Illegal space after opening bracket at test.js :
     3 |var a = [ 1, 2 ]
-----------------^
     4 |var b = [1, 2]

Illegal space before closing bracket at test.js :
     3 |var a = [ 1, 2 ]
----------------------^
     4 |var b = [1, 2]

One space required after opening bracket at test.js :
     3 |var a = [ 1, 2 ]
     4 |var b = [1, 2]
-----------------^

One space required before closing bracket at test.js :
     3 |var a = [ 1, 2 ]
     4 |var b = [1, 2]
---------------------^
```

markelog:
@fluidsonic we will try to release a patch with a fix for this bug soon, let's move this discussion in https://github.com/jscs-dev/node-jscs/issues/1319

PR #1319, opened 2015-04-28
Broken test for requireSpacesInsideArrayBrackets: 'all'

SimenB:
`disallow` passes, `require` fails.

I'm assuming that the require one should cover this case?

### DIFF
--- a/test/specs/rules/disallow-spaces-inside-array-brackets.js
+++ b/test/specs/rules/disallow-spaces-inside-array-brackets.js
@@ -103,6 +103,10 @@ describe('rules/disallow-spaces-inside-array-brackets', function() {
             it('should report illegal space in both cases', function() {
                 assert(checker.checkString('[ 1, 2 ];').getErrorCount() === 2);
             });
+
+            it('should not report with no spaces after function', function() {
+                assert(checker.checkString('var x = "hello-there".split("-")[0];').isEmpty());
+            });
         });
 
         describe('when braces on different lines', function() {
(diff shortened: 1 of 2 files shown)


SimenB:
Maybe related to #1198?

test/specs/rules/disallow-spaces-inside-array-brackets.js
```diff
@@ -103,6 +103,10 @@ describe('rules/disallow-spaces-inside-array-brackets', function() {
             it('should report illegal space in both cases', function() {
                 assert(checker.checkString('[ 1, 2 ];').getErrorCount() === 2);
             });
+
+            it('should not report with no spaces after function', function() {
+                assert(checker.checkString('var x = "hello-there".split("-")[0];').isEmpty());
```
markelog: Same would happen for 

```js
assert(checker.checkString('var x = "hello-there".split("-")[ 0 ];').isEmpty());
```

markelog:
See for more info - https://github.com/jscs-dev/node-jscs/issues/429
Use `(require | disallow)SpacesInsideBrackets` rules instead

SimenB:
@markelog Should `requireSpacesInsideArrayBrackets` be deprecated? `string.split` returns an array, so it's at least deceptively named. `requireSpacesInsideLiteralArrayBrackets`?

markelog:
After #429 was fixed `(require | disallow)SpacesInsideArrayBrackets` started to catch those cases, but we received couple issues about it, some people consider this behaviour as unexpected and want these rules to miss them.

This is why we introduced new more strict rules and removed this logic from `(require | disallow)SpacesInsideArrayBrackets`

SimenB:
Ah, okay. Thanks!

markelog:
@SimenB fyi, when you change your message github doesn't notify the participants, could you open a pull for that?

PR #1322, opened 2015-04-29
Update require-capitalized-comments.js

XhmikosR:


lib/rules/require-capitalized-comments.js
```diff
@@ -52,7 +52,7 @@
  * ```
  *
  * ```js
- * "requireCapitalizedComments": { allExcept: ["jshint"] }
+ * "requireCapitalizedComments": { "allExcept": ["jshint"] }
```
TheSavior: Why is this necessary?

XhmikosR: Because without it an error is thrown. The property needs to be quoted.

TheSavior: Ah yes, json, not js. Ignore me. ;)

XhmikosR: No worries, it took me a while to spot it myself just to realize it was a
bad copy paste from the docs.

markelog: Although it is not fully documented but we can accept config declared through js format

markelog:
Thank you

lib/rules/require-capitalized-comments.js
XhmikosR: Isn't this exact comment used for the rules page? I looked at the website
source and came to the conclusion this file was used.

markelog: You are correct, but in order to update the website we need to pull these changes manually, we will release 1.13.1 today and i will update site with it

PR #1323, opened 2015-04-29
New Rule: (require|disallow)PaddingNewlinesInFiles

astorije:
I am actually more interested in `disallowPaddingNewlinesInFiles` (to make sure there is no blank line at the top and bottom of the file, regardless of the EOF line feed) but I figured it would be nice to mirror and I started with the `require` rule.

I am not finished on the `disallow` rule, but since this is my first rule, any feedback will be helpful :-)

I am having a hard time interacting this with `requireLineFeedAtFileEnd`.
When `requireLineFeedAtFileEnd` and `requirePaddingNewlinesInFiles` are both specified, the file should end with 2 lines. Similarly, when `requireLineFeedAtFileEnd` and `disallowPaddingNewlinesInFiles`, the files must still end with an newline. Any idea how I could do that? Is it possible to know, from a `check()` method, if another rule (in that case, `requireLineFeedAtFileEnd`) is enabled?

### DIFF
--- a/lib/rules/require-padding-newlines-in-files.js
+++ b/lib/rules/require-padding-newlines-in-files.js
@@ -1,0 +1,53 @@
+ /**
+ * Requires the file to start and/or end with a newline.
+ *
+ * Types: `Boolean` or `String`
+ *
+ * Values: `true`, `"beginningOfFileOnly"` or `"endOfFileOnly"`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requirePaddingNewlinesInFiles": true
+ * "requirePaddingNewlinesInFiles": "beginningOfFileOnly"
+ * "requirePaddingNewlinesInFiles": "endOfFileOnly"
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === true || options === 'beginningOfFileOnly' || options === 'endOfFileOnly',
+            this.getOptionName() + ' option requires value to be `true`, `beginningOfFileOnly` or `endOfFileOnly`'
+        );
+
+        if (options === 'beginningOfFileOnly') {
+            this._atBOF = true;
+            this._atEOF = false;
+        } else if (options === 'endOfFileOnly') {
+            this._atBOF = false;
+            this._atEOF = true;
+        } else {
+            this._atBOF = true;
+            this._atEOF = true;
+        }
+    },
+
+    getOptionName: function() {
+        return 'requirePaddingNewlinesInFiles';
+    },
+
+    check: function(file, errors) {
+        if (this._atBOF && file.getLines()[0] !== '') {
+            errors.add('Missing padding new line at beginning of file', 1, 0);
+        }
+
+        if (this._atEOF && file.getLines()[file.getLines().length - 2] !== '') {
+            errors.add('Missing padding new line at end of file', 1, 0);
+        }
+    }
+};
--- a/lib/rules/require-padding-newlines-in-files.js
+++ b/lib/rules/require-padding-newlines-in-files.js
@@ -25,16 +25,8 @@ module.exports.prototype = {
             this.getOptionName() + ' option requires value to be `true`, `beginningOfFileOnly` or `endOfFileOnly`'
         );
 
-        if (options === 'beginningOfFileOnly') {
-            this._atBOF = true;
-            this._atEOF = false;
-        } else if (options === 'endOfFileOnly') {
-            this._atBOF = false;
-            this._atEOF = true;
-        } else {
-            this._atBOF = true;
-            this._atEOF = true;
-        }
+        this._atBOF = options === true || options === 'beginningOfFileOnly';
+        this._atEOF = options === true || options === 'endOfFileOnly';
     },
 
     getOptionName: function() {
@@ -42,12 +34,14 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-        if (this._atBOF && file.getLines()[0] !== '') {
+        var lines = file.getLines();
+
+        if (this._atBOF && lines[0] !== '') {
             errors.add('Missing padding new line at beginning of file', 1, 0);
         }
 
-        if (this._atEOF && file.getLines()[file.getLines().length - 2] !== '') {
-            errors.add('Missing padding new line at end of file', 1, 0);
+        if (lines.length > 1 && this._atEOF && lines[lines.length - 2] !== '') {
+            errors.add('Missing padding new line at end of file', lines.length - 2, 0);
         }
     }
 };
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -640,6 +640,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-padding-newlines-before-line-comments'));
     this.registerRule(require('../rules/require-padding-newlines-before-line-comments'));
 
+    this.registerRule(require('../rules/require-padding-newlines-in-files'));
+
     this.registerRule(require('../rules/disallow-trailing-comma'));
     this.registerRule(require('../rules/require-trailing-comma'));
 
--- a/test/specs/rules/require-padding-newlines-in-files.js
+++ b/test/specs/rules/require-padding-newlines-in-files.js
@@ -1,0 +1,46 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/require-padding-newlines-in-files', function() {
+    var checker;
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    it('should report no padding new lines at file start and end', function() {
+        checker.configure({ requirePaddingNewlinesInFiles: true });
+        assert(checker.checkString('var x;\n').getErrorCount() === 2);
+    });
+
+    it('should not report padding new lines at file start and end', function() {
+        checker.configure({ requirePaddingNewlinesInFiles: true });
+        assert(checker.checkString('\nvar x;\n\n').isEmpty());
+    });
+
+    it('should report no padding new line at file start', function() {
+        checker.configure({ requirePaddingNewlinesInFiles: 'beginningOfFileOnly' });
+        assert(checker.checkString('var x;').getErrorCount() === 1);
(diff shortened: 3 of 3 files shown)


astorije:
AppVeyor failed, but it looks like a false alarm to me... right?

```
 1) modules/checker fixing "before each" hook:
    Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

zxqfox:
@astorije I've restarted appveyor.

2secs is not enough for tests?

lib/rules/require-padding-newlines-in-files.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === true || options === 'beginningOfFileOnly' || options === 'endOfFileOnly',
+            this.getOptionName() + ' option requires value to be `true`, `beginningOfFileOnly` or `endOfFileOnly`'
+        );
+
+        if (options === 'beginningOfFileOnly') {
+            this._atBOF = true;
```
zxqfox: What if:
```
this._atBOF = options === true || options === 'beginningOfFileOnly';
this._atEOF = options === true || options === 'endOfFileOnly';
```
?

lib/rules/require-padding-newlines-in-files.js
```diff
@@ -0,0 +1,53 @@
...
+    getOptionName: function() {
+        return 'requirePaddingNewlinesInFiles';
+    },
+
+    check: function(file, errors) {
+        if (this._atBOF && file.getLines()[0] !== '') {
+            errors.add('Missing padding new line at beginning of file', 1, 0);
+        }
+
+        if (this._atEOF && file.getLines()[file.getLines().length - 2] !== '') {
```
zxqfox: It can fail if file has only 1 line. I think we can check `file.getLines().length` at function start and skip files with less than 3 lines.

Also would be nice to call `file.getLines()` just once and store to a variable.

lib/rules/require-padding-newlines-in-files.js
astorije: Doh, of course! I knew these statements were gross, why did I miss this factorization?!

Thanks, fixing now :-)

zxqfox: :-D :+1:

lib/rules/require-padding-newlines-in-files.js
astorije: Yes, I meant to do that and forgot, sorry. Thanks for pointing it out!

astorije:
Thanks @zxqfox, your comments are addressed in d07d185.

Regarding timeout, I'm afraid I have no idea, I am not modifying this test, so I expected AppVeyor to be slow on his side.

zxqfox:
Yup, unfortunately it happens too often with appveyor. ;-(

lib/rules/require-padding-newlines-in-files.js
astorije: > I think we can check `file.getLines().length` at function start and skip files with less than 3 lines.

Does my fix in d07d185 is fine by you? I am checking that the file is at least 2 lines, but I can do 3 if you think it's more relevant.

zxqfox: Yeap, thanks! :+1: 

Not sure about 3. Just thinking about:
- empty file (one line) — both rules will error
- two lines:
  ```
  // ← this line will be checked by both rules
  // ← this line by nothing, should it be empty?
  ```
- three lines
  ```
  // ← this line by BOF
  // ← this line by EOF
  // should be empty?
  ```

So 3+ lines is working env for this rule at my pov. That's why I said it ;-)

Not sure, btw, what's with the last line? Should it be empty?

astorije: Good question, but this depends on `requireLineFeedAtFileEnd` :-/

If `requireLineFeedAtFileEnd` is set to `true`, this will pass:

```
// ←this line by BOF
var x;
// ←this line by EOF
// ←this is the line feed
```

If not, this will pass:

```
// ←this line by BOF
var x;
// ←this line by EOF, *and* it is also the line feed
```

Forgetting the BOF rule (ie using `endOfFileOnly`), this will fail if `requireLineFeedAtFileEnd` is set:
```
var x;
// ←New line, which is the line feed
```

Think of it in the opposite rule (I should have started with it actually). If you set `disallowPaddingNewlinesInFiles` to `true`, you want your files to be:

```js
var x;
```

But if you **also** set `requireLineFeedAtFileEnd` to `true`, then only files with that format must pass:

```js
var x;
// Line feed, empty line
```

Although the last line is empty, it respects `requireLineFeedAtFileEnd` (the line feed being often used by systems such as git) **and** `disallowPaddingNewlinesInFiles` (as both lines at top and bottom are style choices this time).

Does that make sense?
It seems like I have to interact with another rule (know if it is set), is there a way to do that? If not, do you see an alternative use case we could choose to avoid this issue?

zxqfox: /cc @markelog @mikesherov 

Can we detect that rule is used (line feed at eof is required)?

```
             ← BOF-padding
             ← code
             ← EOF-padding
             ← line feed at EOF
```

zxqfox: @astorije Agree that we should know it to check intelligently. ;-( Otherwise we just going to force requireLineFeedAtEOF with this rule.

astorije: Thanks for your input @zxqfox! I'll wait for @markelog and @mikesherov's feedback to see if we can detect such thing before starting the `disallow` rule.

zxqfox:
@markelog @mikesherov Guys, we need you here

mdevils:
Hello @astorije!

Thank you for your submission!

Why not just create `disallowLineFeedAtFileEnd` and `disallowLineFeedAtFileStart`?

astorije:
Hi @mdevils!
`disallowLineFeedAtFileEnd` would not help unfortunately: the line feed at the end is arguably something one might want (for example, git complains when not present), but not necessarily having one (or several) extra blank lines at the end of the file.

There are explanations and examples in inline conversations between @zxqfox and me, but think of this.

To allow (force) the following:

```
var x;       ← code
             ← line feed at EOF
```

and disallow the following:

```
var x;       ← code
             ← EOF-padding
             ← line feed at EOF
```

You need to set both `requireLineFeedAtFileEnd` and `disallowPaddingNewlinesInFiles` to `true`. Does that make more sense now?
(In a nutshell, this new `disallow` rule is to disallow all padding lines at top and bottom while still requiring the line feed when it is required)

astorije:
Guuuuys, any help on this? :-)

mdevils:
OK, I understand your explanation as a human, but how machine will understand this? I mean, you won't be able to combine `disallowPaddingNewlinesInFiles` and `requireLineFeedAtFileEnd`.

astorije:
Actually @mdevils, these are exactly what I want to combine :-)

Here are the use cases (focusing only on end of files):

- `disallowPaddingNewlinesInFiles` and `requireLineFeedAtFileEnd`: **exactly 1 blank line** at the end of files (for the line feed)
- `requirePaddingNewlinesInFiles` and `requireLineFeedAtFileEnd`: **at least 2 blank lines** at the end of files, 1 of them being for the line feed (and the other one(s) for style preferences, clarity, ... whatever your team has decided)
- `requirePaddingNewlinesInFiles` only **or** `requireLineFeedAtFileEnd` only: **at least 1 blank line** at the end of files
- `disallowPaddingNewlinesInFiles` only: **exactly 0 blank lines** at the end of files

So if by "machine" you mean jscs, then yes, the `*PaddingNewlinesInFiles` rules needs to know if `requireLineFeedAtFileEnd` is specified to complete. Good news is that since there is no `disallowLineFeedAtFileEnd`, this need is not reciprocal!

mdevils:
I believe this needs more discussion.

/cc @markelog @mikesherov @mrjoelkemp

astorije:
Still no luck? :(

mdevils:
> requirePaddingNewlinesInFiles and requireLineFeedAtFileEnd: at least 2 blank lines at the end of files, 1 of them being for the line feed (and the other one(s) for style preferences, clarity, ... whatever your team has decided)
> requirePaddingNewlinesInFiles only or requireLineFeedAtFileEnd only: at least 1 blank line at the end of files

Right now we don't have mechanism which would allow rule to access other rule's settings. And it doesn't to me right to have this mechanism.

I would propose just to go for your requirements and disallow padding newlines in files.

mrjoelkemp:
Can we not just use an EOF token in the rules for padding to know the difference between a padding newline and a newline for the EOF?

This would avoid needing coordination between the sets of rules.

mdevils:
@mrjoelkemp I'm not sure how it may help us. Can you elaborate on this?

mrjoelkemp:
The last newline in a file should be the EOF token, in that it signifies the end of the file.

The padding newlines rules should only act on newlines that are not the EOF token. Within the requirePaddingNewlines check, for all found newlines, if it's not an EOF token, throw an error.

mdevils:
If I understood you correctly, you are proposing to have ambiguous EOF token which includes or does not include last LF?

mdevils:
Because right now EOF token is literally the end of file, outside of all ranges, after the last CF, so currently EOF token never has whitespace data.

markelog:
What is the status of this? @mrjoelkemp?

mrjoelkemp:
Disregard my comments. This is fine to go in if Marat gave the lgtm.

astorije:
@markelog I still have some work to do on this.

Sorry others things in life and projects took priority. I was actually thinking this weekend about excavating this PR and I definitely want this to make it into master so I'll finish these rules for sure. The when is less certain though.

Thanks for your patience.

markelog:
Friendly ping

markelog:
Okay, feel bad closing this, ping us up if you change your mind

astorije:
Hi @markelog,

Apologies that I missed your ping earlier. As I said in my last comment, there were other priorities in my life and in my FOSS projects so I had to put this on the side. But I will definitely finish this sooner or later as I need it as well. I just need to close some higher items on my todolist to make this the top one again.

Sorry about wasting your time though. When I get to work on that, I will continue to use this PR for posting questions and comments (last thing I remember was being stuck as explained in earlier comments), and will open a new PR for the code.

Thanks for being responsive and open to new ideas!

Issue #1081, opened 2015-02-18
New Rule: Disallow Identical Destructuring Names?


iamdanfox:
ES2015's destructuring assignment has a [shorthand form][1] - it would be cool if JSCS could flag up code that could be shortened by using this syntax.

For example, the new rule would flag up this code:

    var {left: left, top: top} = this.refs.container.getDOMNode().getBoundingClientRect();

But allow the following lines:

    var {left, top} = this.refs.container.getDOMNode().getBoundingClientRect();
    var {left: someName, top: otherName} = this.refs.container.getDOMNode().getBoundingClientRect();

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Simple_example_2

mikesherov:
@iamdanfox great idea. Any interest in implementing this feature? The only caveat is that the rule itself needs to check if it's in ES6 mode using `file.getDialect`.

iamdanfox:
I'm afraid I'm completely new to the codebase, but might be able to try implementing it.

From looking through `lib`, would you recommend using `file.iterateNodesByType` with `VariableDeclarator`/`FunctionDeclaration`?

Presumably we'd then need to go through tokens of `node.left`?

mikesherov:
@iamdanfox I'd imagine you'd want this to apply to all `objectExpressions`, no? so `file.iterateNodesByType` would work. Also, you don't **need** to implement this if you'd rather wait for us to get around to it. We just like asking issue contributors if they'd like to become code contributors to. Thanks again for contributing!

iamdanfox:
The [MDN article][1] shows a lot of different variants (e.g. nested destructuring, within `for of` loops, as function parameters and computed object property names).

Going for the simple assignment case sounds look a good place to start.  Would `ObjectExpression` catch things like:
    
    var obj = {x: 1, y: 2};
    var x, y;
    {x: x, y: y} = obj;

I was wondering whether `AssignmentExpression` might be appropriate?

(I think it's probably sensible for me to let a more experienced contributor implement this - I can't really commit the time to produce something good!)

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring

jasonkarns:
I would add that as a user, I would expect a configuration option to this rule that allowed or disallowed mixing. For instance:

`var {left, top: topper} = ...`

[...]

PR #1325, opened 2015-04-30
New Rules: (require|disallow)ParenthesesAroundArrowParam (ES6)

### LINKED ISSUE #1307: New Rules: (require|disallow)ParenthesesAroundArrowParams (ES6)
[airbnb/javascript#8.3](https://github.com/airbnb/javascript#8.3) 

~~would want `requireParenthesesAroundArrowParams`~~
would want `disallowParenthesesAroundArrowParams`. airbnb/javascript#347

Always use parentheses around the arguments. Omitting the parentheses makes the functions less readable and only works for single arguments.

> Why? These declarations read better with parentheses. They are also required when you have multiple parameters so this enforces consistency.

```js
// bad
[1, 2, 3].map(x => x * x);
// good
[1, 2, 3].map((x) => x * x);
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12376741-new-rules-require-disallow-parenthesesaroundarrowparams-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Require/disallow parentheses around arrow function expressions with a single parameter.

```
// disallow
[1, 2, 3].map(x => x * x);
// require
[1, 2, 3].map((x) => x * x);
```

Fixes #1307.

~~Going to add the `disallow` rule after some feedback.~~

I just copied the `isWrapped` function from the `requireParenthesesAroundIIFE` rule.

~~I wasn't sure about the es6 test - so I did `file.getDialect() !== 'es6'` based on what @mikesherov said - I don't think we have any rules like this yet?~~

Also I used `ArrowFunctionExpression` since `ArrowExpression` didn't work - where should I be looking to know what to use (normally I can just find the jscs codebase). In the `validateIndentation` rule I see `ArrowExpression`.

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -567,6 +567,11 @@ Configuration.prototype.registerDefaultRules = function() {
         These rules are linked explicitly to keep browser-version supported.
     */
 
+    /* ES6 only */
+    this.registerRule(require('../rules/require-parentheses-around-arrow-param'));
+    this.registerRule(require('../rules/disallow-parentheses-around-arrow-param'));
+    /* ES6 only (end) */
+
     this.registerRule(require('../rules/require-curly-braces'));
     this.registerRule(require('../rules/disallow-curly-braces'));
     this.registerRule(require('../rules/require-multiple-var-decl'));
--- a/lib/rules/require-parentheses-around-arrow-param.js
+++ b/lib/rules/require-parentheses-around-arrow-param.js
@@ -1,0 +1,70 @@
+/**
+ * Requires parentheses around arrow function expressions with a single parameter.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireParenthesesAroundArrowParam": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * [1, 2, 3].map((x) => x * x);
+ * // params are always required for multiple parameters
+ * [1, 2, 3].map((x, y, z) => x * x);
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * [1, 2, 3].map(x => x * x);
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireParenthesesAroundArrowParam';
+    },
+
+    check: function(file, errors) {
+        function isWrapped(node) {
+            var openParensToken = file.getPrevToken(file.getFirstNodeToken(node));
+            var closingParensToken = file.getNextToken(file.getLastNodeToken(node));
+            var closingTokenValue = closingParensToken ? closingParensToken.value : '';
+
+            return openParensToken.value + closingTokenValue === '()';
+        }
+
+        file.iterateNodesByType('ArrowFunctionExpression', function(node) {
+            var params = node.params;
+            var firstParam = params[0];
+
+            if (params.length === 1 && !isWrapped(firstParam)) {
+                errors.add(
+                    'Wrap arrow function expressions in parentheses',
+                    firstParam.loc.start
+                );
+            }
+        });
+    }
+
+};
--- a/lib/rules/disallow-parentheses-around-arrow-param.js
+++ b/lib/rules/disallow-parentheses-around-arrow-param.js
@@ -1,0 +1,70 @@
+/**
+ * Disallows parentheses around arrow function expressions with a single parameter.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowParenthesesAroundArrowParam": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * [1, 2, 3].map(x => x * x);
+ * // params are always required for multiple parameters
+ * [1, 2, 3].map((x, y, z) => x * x);
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * [1, 2, 3].map((x) => x * x);
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true,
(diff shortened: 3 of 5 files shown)


lib/rules/require-parentheses-around-arrow-param.js
```diff
@@ -0,0 +1,76 @@
...
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireParenthesesAroundArrowParam';
+    },
+
+    check: function(file, errors) {
+        if (file.getDialect() !== 'es6') {
```
markelog: Why?

markelog:
> Also I used ArrowFunctionExpression since ArrowExpression didn't work - where should I be looking to know what to use (normally I can just find the jscs codebase). In the validateIndentation rule I see ArrowExpression.

No place better then https://github.com/jquery/esprima/blob/master/esprima.js#L108-169

lib/rules/require-parentheses-around-arrow-param.js
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        }
+
+        function isWrapped(node) {
```
markelog: Might make sense to expose it in helper module

lib/rules/require-parentheses-around-arrow-param.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Requires parentheses around arrow function expressions with a single parameter.
```
markelog: Might make sense to add "es6 only rule"  or something like that

lib/config/configuration.js
```diff
@@ -727,6 +727,8 @@ Configuration.prototype.registerDefaultRules = function() {
 
     this.registerRule(require('../rules/validate-newline-after-array-elements'));
 
+    this.registerRule(require('../rules/require-parentheses-around-arrow-param'));
```
markelog: Might make sense to move es6 require rules in special helper, or at least in spacial place in this method

lib/rules/require-parentheses-around-arrow-param.js
hzoo: Not sure what @mikesherov meant in https://github.com/jscs-dev/node-jscs/issues/1081#issuecomment-74891074 otherwise.

lib/rules/require-parentheses-around-arrow-param.js
hzoo: Yeah I think it would be ok to put it in `utils` then?

lib/rules/require-parentheses-around-arrow-param.js
hzoo: Maybe just `ES6: Requires parentheses...` or we add another category like `Version: ES6` similar to `Type, Version`

lib/rules/require-parentheses-around-arrow-param.js
markelog: If there is no nodes like that it wouldn't break. It could save time if `esnext` is not enabled, but if user enabled es6-only rule we could say their know what their are doing

hzoo: Yeah I was thinking that too - I forgot to add `esnext` to the test file and the nodes of course didn't return anything. And I guess if we are using esnext by default later..

lib/rules/require-parentheses-around-arrow-param.js
markelog: Second one sounds better, but any of them will do, i think

lib/rules/require-parentheses-around-arrow-param.js
markelog: Yeah

lib/rules/require-parentheses-around-arrow-param.js
markelog: > And I guess if we are using esnext by default later..

Well, i'm starting to doubt it would be a best course of action, since if we would have - 

```js
disallowKeywords: {
  es6: ['var']
}
```
and add it to the airbnb preset while enabling `esnext` by default - :boom:

lib/rules/require-parentheses-around-arrow-param.js
hzoo: isWrapped is using file so I would need to add that to the method signature like
`exports.isWrapped = function(file, node) {`. How would I write the test for this since normally we do `checker.checkString`?

markelog: Generate the file instance, take the node, passed them to `isWrapped`, what else is there?

markelog: You can do that in another pull though

hzoo: Oh I just don't know the method to create a file instance?

lib/rules/require-parentheses-around-arrow-param.js
hzoo: Should I remove this check? Otherwise I think I'm good with this (and then do the `isWrapped` PR)

markelog: Would do you think? Do you think it needed? Or not?

lib/rules/require-parentheses-around-arrow-param.js
markelog: See https://github.com/jscs-dev/node-jscs/blob/fb9ce4fe6f55c83c96fa311a0973dfa77e5740af/test/specs/js-file.js

markelog:
> Coverage decreased (-0.05%) 

?

hzoo:
It's because of the `if (file.getDialect() !== 'es6') {` which might not be needed?

lib/rules/require-parentheses-around-arrow-param.js
hzoo: I guess not because it won't error anyway... I can remove it from the new rules.

You can assume that if someone is using an es6 rule they would have esnext enabled but it's always nice to have things that help the user experience - don't know if we could have an error if someone enables an es6 rule without esnext.

markelog: @mikesherov what do you think?

mikesherov: I don't think we need this check. The parser will throw up before we ever get here, no?

hzoo: It will just ignore it since `file.iterateNodesByType('ArrowFunctionExpression'` is empty. Ok will remove

hzoo:
Ok squashed and removed `getDialect` check.

PR #1335, opened 2015-05-03
New Rule: requireTemplateStrings (ES6)

### LINKED ISSUE #1330: New Rule: requireTemplateStrings or disallowStringConcat (ES6)
[airbnb/javascript#6.4](https://github.com/airbnb/javascript/tree/master#6.4)

When programmatically building up strings, use template strings instead of concatenation.

> Why? Template strings give you a readable, concise syntax with proper newlines and string interpolation features.

```js
  // bad
  function sayHi(name) {
    return 'How are you, ' + name + '?';
  }

  // bad
  function sayHi(name) {
    return ['How are you, ', name, '?'].join();
  }

  // good
  function sayHi(name) {
    return `How are you, ${name}?`;
  }
```

Not sure but I think we could use the logic in `disallowImplicitTypeConversion` unless we have a helper to check for string concatenation?

```js
file.iterateNodesByType('BinaryExpression', function(node) {
  if (node.operator === '+' && ...)
```

maybe if `node.left.type === 'Literal' && node.right.type === 'Identifier'` and vice-versa?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12838873-new-rule-requiretemplatestrings-or-disallowstringconcat-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1330.

Initial try?

Can finish this after #1325 is done to fix location of line for `lib/config/configuration.js`

### DIFF
--- a/lib/rules/require-template-strings.js
+++ b/lib/rules/require-template-strings.js
@@ -19,6 +19,7 @@
  * function sayHi(name) {
  *     return `How are you, ${name}?`;
  * }
+ * `a ${b + c}`
  * ```
  *
  * ##### Invalid
@@ -27,6 +28,7 @@
  * function sayHi(name) {
  *     return 'How are you, ' + name + '?';
  * }
+ * "a" + (b + c)
  * ```
  */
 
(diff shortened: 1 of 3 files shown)


lib/rules/require-template-strings.js
```diff
@@ -0,0 +1,69 @@
...
+ * function sayHi(name) {
+ *     return `How are you, ${name}?`;
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * function sayHi(name) {
+ *     return 'How are you, ' + name + '?';
```
yaycmyk: what about a single-join concatenation?

```js
return 'How are you, ' + name;
```

hzoo:
Ok testing on other code.. 

```js
// Identifer, Literal
val + 1
val + true
val + 'a'
val + "a"
// Identifer, TemplateLiteral
val + `a`
```

Will need to actually check the value of the literal to prevent errors on `val + 1` when it's intended to add numbers, so I guess add a check for `typeof value !== 'number'`?

hzoo:
Yeah another implementation from `ember-suave` here: https://github.com/dockyard/ember-suave/blob/4303e94782cb18f70f072fe4e211619e11a2e42a/lib/rules/require-template-strings-for-concatenation.js.

Should the rule be renamed to: `requireTemplateStringsForConcatenation`?

markelog:
Just send you a pull, both names are fine with me

hzoo:
Ok I just rebased it and moved the rules together

hzoo:
Ok messed up the rebase so your commit only has the changes in the rule if that's ok. Or are we going to rebase into to 1 commit anyway? (I added a test for coverage).

PR #1337, opened 2015-05-04
New Rule: (disallow|require)ObjectKeysOnNewLine

### LINKED ISSUE #1249: Existing rule for only allowing one key definition in object declaration per line?
I just added `  "requirePaddingNewLinesInObjects": true` to our projects configuration and formatted:

```javascript
var a = {b: 1, c: 2}
```

Which turned into

```javascript
var a = {
    b: 1, c: 2
};
```

I was looking for an additional rule to add that would require it to be like this:

```javascript
var a = {
    b: 1, 
    c: 2
};
```

but didn't see anything in the docs that would do that. Is there a rule for this yet? It looks like the closest I can find is `requireLineBreakAfterVariableAssignment` but that seems to only care about `var` statements, and not object declarations.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10739405-existing-rule-for-only-allowing-one-key-definition-in-object-declaration-per-line?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
Fixes #1249

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -599,6 +599,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/validate-indentation'));
     this.registerRule(require('../rules/disallow-trailing-whitespace'));
     this.registerRule(require('../rules/disallow-mixed-spaces-and-tabs'));
+    this.registerRule(require('../rules/require-object-keys-on-new-line'));
+    this.registerRule(require('../rules/disallow-object-keys-on-new-line'));
     this.registerRule(require('../rules/require-keywords-on-new-line'));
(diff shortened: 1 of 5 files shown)


TheSavior:
Bump?

TheSavior:
@markelog @mikesherov @hzoo 

Bump...again. We seem to be way past a patch version and other PRs have been merged. What should I be waiting for on this one?

PR #1341, opened 2015-05-05
validateAlignedFunctionParameters: Fix doc

adrianheine:


markelog:
Oh, that's unpleasant, thank you.

markelog:
I see you are from wikimedia, do you need that rule in "wikimedia" preset?

adrianheine:
I'm from Wikimedia Deutschland, not the foundation – we follow a slightly different coding style. However, the [MediaWiki coding conventions](https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Line_length) suggests this rule, and the MediaWiki code base passes it. So yeah, I'd say the wikimedia preset should have it. @Krinkle do you agree?

PR #1345, opened 2015-05-06
requirePaddingNewlinesAfterBlocks no longer uses global state

### LINKED ISSUE #1343: requirePaddingNewLinesAfterBlocks uses global state
`requirePaddingNewLinesAfterBlocks` uses global state for the exceptions, which means you cannot have multiple instances with different settings. Moreover, when `requirePaddingNewLinesAfterBlocks` is configured with `true`, it does not overwrite the exceptions, so you cannot run with `true` after you ran with exceptions in one nodejs run. **Edit:** It actually never resets exceptions.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13220897-requirepaddingnewlinesafterblocks-uses-global-state?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TheSavior:
Fixes #1343

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -113,13 +113,7 @@ var excludes = {
     'FunctionExpression': ['.'],
 };
 
-var exceptions = {
-    'CallExpression': false,
-    'ArrayExpression': false,
-    'Property': false
-};
-
-function isException(parent) {
+function isException(parent, exceptions) {
     var grandpa = parent.parentNode;
 
     // Check if this block is used in call or array expression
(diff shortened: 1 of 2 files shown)


adrianheine:
Thanks, looks great!

zxqfox:
"grandpa" :smirk_cat: 

:+1:

markelog:
Thank you!

PR #1350, opened 2015-05-08
New Rule: requireShorthandArrowFunctions (ES6)

### LINKED ISSUE #1340: New Rule: requireShorthandArrowFunctions (ES6)
[airbnb/javascript#8.2](https://github.com/airbnb/javascript/tree/master#8.2)

Names: 
`requireArrowFunctionShorthand`
`requireImplicitReturnInArrowFunction`
`disallowExplicitReturnInArrowFunction`

If the arrow function body is only a single expression?

```js
// good
evens.map(v => v + 1);
// bad
evens.map(v => { return v + 1; });
```

Maybe..?
```js
file.iterateNodesByType('ArrowFunctionExpression', function(node) {
    if (body.type === 'BlockStatement' &&
        body.body.length === 1 &&
        body.body[0].type === 'ReturnStatement') {
        errors.add..
    }
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13120182-new-rule-requireshorthandarrowfunctions-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1340.

### DIFF
--- a/lib/rules/require-shorthand-arrow-functions.js
+++ b/lib/rules/require-shorthand-arrow-functions.js
@@ -1,0 +1,67 @@
+/**
+ * Require arrow functions to use an expression body when returning a single statement
+ * (no block statement, implicit return).
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireShorthandArrowFunctions": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * // single expression
+ * evens.map(v => v + 1);
+ * // multiple statments require a block
+ * evens.map(v => {
+ *     v = v + 1;
+ *     return v;
+ * });
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * evens.map(v => { return v + 1; });
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireShorthandArrowFunctions';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('ArrowFunctionExpression', function(node) {
+            var body = node.body;
+            if (body.type === 'BlockStatement' &&
+                body.body.length === 1 &&
+                body.body[0].type === 'ReturnStatement') {
+                errors.add(
+                    'Use the shorthand arrow function form',
+                    node.body.loc.start
+                );
+            }
+        });
+    }
+
+};
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -728,6 +728,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/validate-newline-after-array-elements'));
 
     this.registerRule(require('../rules/disallow-not-operators-in-conditionals'));
+
+    this.registerRule(require('../rules/require-shorthand-arrow-functions'));
 };
 
 /**
--- a/test/specs/rules/require-shorthand-arrow-functions.js
+++ b/test/specs/rules/require-shorthand-arrow-functions.js
@@ -1,0 +1,24 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/require-shorthand-arrow-functions', function() {
+    var checker;
+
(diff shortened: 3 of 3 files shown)


zxqfox:
Name suggestion: `requireArrowFunctions`?

hzoo:
I suggested `requireShorthandArrowFunctions` because for the 2 examples for the rules both are still arrow functions right? The other one just doesn't have the block and return statements.

And also I was thinking we could do another rule called `requireArrowFunctions` which would ask for an arrow function when possible instead of resassigning `this` or `bind` (basically figure out whenever it's a possible replacement which should be all anon functions / function expressions I think).

lib/rules/require-shorthand-arrow-functions.js
```diff
@@ -0,0 +1,64 @@
...
+ * ```js
+ * "requireShorthandArrowFunctions": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * evens.map(v => v + 1);
+ * evens.map(v => {
+ *     v = v + 1;
```
markelog: ?

markelog: Confusion for me, i don't understand why would anyone would wants this

hzoo: I wasn't sure if I needed this example but it's just to show it doesn't error when the function has multiple statements in it since you need the braces and return in that case.

hzoo: Do you mean this specific example or the rule? I can remove this example if it's confusing.

lib/rules/require-shorthand-arrow-functions.js
```diff
@@ -0,0 +1,64 @@
...
+ *     return v;
+ * });
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * evens.map(v => { return v + 1; });
```
markelog: That is not a "single statement", which is hard to get your head around it, i'd say this rule is strange, is there any use-cases for it?

markelog: Oh, so it prohibits use of `return` if there is a single expression?

hzoo: Well what I'm thinking is:
one form is `v => v + 1`.
The other is `{ return v => v + 1 }`.
The shorter form is only allowed when you return a single statement/expression - it becomes a implicit return.

lib/rules/require-shorthand-arrow-functions.js
```diff
@@ -0,0 +1,64 @@
...
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('ArrowFunctionExpression', function(node) {
+            var body = node.body;
+            if (body.type === 'BlockStatement' &&
+                body.body.length === 1 &&
+                body.body[0].type === 'ReturnStatement') {
+                errors.add(
+                    'Use the shorthand arrow function form.',
```
markelog: No dot at the end

lib/rules/require-shorthand-arrow-functions.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * Require shorthand arrow function expressions when returning a single statement.
```
markelog: Maybe i was the only one confused by this, but if you could clarify this a bit more, that would be great

hzoo: No I think you're right about it being confusing. I'm not sure what the best wording should be. I think it's since arrow functions are actually the shorthand of normal functions anyway.

One distinction is that one way is an expression body and the one with braces is a statement body (block)

```js
// Expression bodies
var odds = evens.map(v => v + 1);
// Statement bodies
nums.forEach(v => {
  if (v % 5 === 0)
    fives.push(v);
});
```

Maybe even the rule name should be changed but I'm not sure.

zxqfox: Are you sure you can use `fives` (outer variables) inside arrow functions?

hzoo: Yeah since arrow functions are just a shorter syntax for function expressions. This is copied from airbnb styleguide. Although that isn't exactly the point of the example so 

```js
// expression body (single expression so don't need curly braces and can do implicit return)
nums.map(v => v + 1);
// statement body (two statement requires curly braces + return statement
nums.map(v => {
  v = v + 1;
  return v;
});
```

Issue #1358, opened 2015-05-11
Style issues


markelog:
> expects valid type instead of Promise * Error[] at lib/checker.js :
   138 | * @param {String} path
   139 | * @param {Function} fileHandler
   140 | * @returns {Promise * Error[]}
---------------------^
   141 | */
   142 |Checker.prototype._processPath = function(path, fileHandler) {

/cc @zxqfox

zxqfox:
`Promise * Error[]` is not valid jsdoc afaik. 
https://github.com/jsdoc3/jsdoc/issues/509

markelog:
Yeah, i read it, as i understand this is not part of the "standart" yet?

mrjoelkemp:
What's the best way to handle this via the jscs-jsdoc plugin? Ignore that style of type or disable specific instances? Or do we have to change them until the spec is updated?

markelog:
I'd say we need understand if "jsdoc" plugin is correct here, if it is, then we should fix the issues, if not, we should correct the check.

zxqfox:
@markelog Yep, there is no standard. If we parsing it — we can stick to something parsable.

Tell me what it means and I tell you if it doesn`t correct.

Personally, I don't see why `*` (star) should be a sign to use for promises when Google Closer and jsdoc3 standard uses `<...>` for that.

Promise is an Object that contains Result or Reject reason, several callbacks and internal state. So why we can't use a standard: `Promise.<?Result, ?Reject>`?

mrjoelkemp:
I'm fine with something like `Promise.<String>` if that's standard. To be clear, this implies that the promise resolves with a string, correct?

zxqfox:
@mrjoelkemp Yep. Just there is a standard for *any* generic objects like `Array.<Object.<String, function(Array.<String>)>>` driven by Google Closure and integrated to jsdoc toolkit 3. But there is no something specific for Promises right now.

btw, I'm using online tool that parses jsdoc: http://kuniwak.github.io/jsdoctypeparser/ and since 1.2.0 it uses PEG, but `*` is still unknown thing. We can shut this exception in our jsdoc plugin, but still it's invalid (or unknown).

zxqfox:
I'd say `{Promise * Error[]}` should be a `Promise.<Error[]>` for a while.

markelog:
> driven by Google Closure and integrated to jsdoc toolkit 3

Enough for me, could you send a PR?

PR #1357, opened 2015-05-11
Preset: require spaces in anonumous fe for crockford style

markelog:
Fixes #1351

PR #1361, opened 2015-05-11
Add intergration tests for "autofix" feature

### LINKED ISSUE #1086: Integration tests
Since auto-fix might became more complex then it is now, it might introduce some unpleasant bugs, we would need to do our best to avoid them.

It relatively easy catch syntactic errors, since esprima would notify us about them, but algorithmic errors is another story. 

In order to catch them, it would makes sense to add integration tests: include couple popular libraries with big code-base and high test coverage (maybe even create a different repo for it) then run auto-fix with all available presets and on every auto-fix execution - run their tests and see if something breaks.

Presumably, these tests might take good time to run, i would create their own `npm` command for it, so they wouldn't execute on `npm test` command.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8709429-integration-tests?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Fixes #1086

### DIFF
--- a/test/scripts/integration.js
+++ b/test/scripts/integration.js
@@ -35,15 +35,20 @@ function execPresets(presets) {
         unitError: 'Unit tests failure'
     };
 
-    (function(messages, args) {
+    (function test(messages, args) {
+
+        // Execute jscs -x on jscs source code for the given presets
         var child = spawn('node', args, { timeout: Infinity });
 
         process.stdout.write(messages.start + '\n\n');
         process.stdout.write(chalk.magenta(' - ') + messages.fixStart + '\n');
 
+        // For some reason this makes subprocess to be more effective
         child.stdout.on('data', function() {});
 
-        child.on('close', function(code) {
+        // Wait until autofix process is done
+        child.on('close', function executeTests(code) {
+
             // exit code == 2 means there is some unfixable style violations
             if (code !== 0 && code !== 2) {
                 process.stderr.write(chalk.red(' ! ') + messages.fixError + '\n\n');
@@ -52,12 +57,14 @@ function execPresets(presets) {
                 return;
             }
 
+            // Execute jscs tests on modified source code
             var sChild = spawn('mocha', ['--color']);
             var counter = 1;
 
             process.stdout.write(chalk.magenta(' - ') + messages.unitStart + '\n');
 
-            sChild.stdout.on('data', function() {
+            // Show output, basically show that something is going on
+            sChild.stdout.on('data', function log() {
 
                 // Do not be so verbose
                 if (counter % 5 === 0) {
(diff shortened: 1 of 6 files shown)


test/scripts/integration.js
```diff
@@ -0,0 +1,97 @@
...
+        unitError: 'Unit tests failure'
+    };
+
+    (function(messages, args) {
+        var child = spawn('node', args, { timeout: Infinity });
+
+        process.stdout.write(messages.start + '\n\n');
+        process.stdout.write(chalk.magenta(' - ') + messages.fixStart + '\n');
+
+        child.stdout.on('data', function() {});
```
mrjoelkemp: Otherwise what happens?

mrjoelkemp: I guess the general criticism is to comment why you needed to do this.

mrjoelkemp:
Great work here. Not really following what the approach is from reading the code. Can you refactor the code to use more named functions to clarify the high-level algorithm?

markelog:
@mrjoelkemp addressed your remarks

test/specs/config/generator.js
```diff
@@ -113,7 +113,7 @@ describe('lib/config/generator', function() {
         });
     });
 
-    it('generates a .jscsrc file with the user\'s violation choices', function(done) {
+    it.skip('generates a .jscsrc file with the user\'s violation choices', function(done) {
```
TheSavior: why are we turning off a test?

mrjoelkemp: See https://github.com/jscs-dev/node-jscs/pull/1357#discussion_r30009897

mrjoelkemp:
The comments actually helped a ton. You're running autofix on a subdirectory of the codebase and then making sure the tests still pass. Got it. 

Should we also run non auto-fixed jscs on that subdirectory to make sure the style isn't violated post-autofix?

zxqfox:
@markelog #1358 is landed, no more need to disable jsdoc checks.

markelog:
> Should we also run non auto-fixed jscs on that subdirectory to make sure the style isn't violated post-autofix?

You mean run autofix with our config?

mrjoelkemp:
No, your integration test runs autofix then makes sure the tests pass (ensuring that autofix didn't break the code). I'm saying after doing that, make sure that the style isn't violated by running `jscs -p presetName lib test/specs`.

Autofix for a rule, if not implemented correctly, can either break code (by producing syntax errors) or break style (by modifying the ast incorrectly). Your integration test only handles the first case, as far as I can tell.

markelog:
There is examples like `crockford` style that prevents use of `continue` keyword for example, such presets will always return broken styles.

mrjoelkemp:
Ah, plus we don't have autofix for every rule in a preset, so we'll always have style errors until we have all rules as fixable.

mrjoelkemp:
LGTM

PR #1364, opened 2015-05-11
disallowSpaceAfterKeywords: fix else if case

### LINKED ISSUE #1346: "disallowSpaceAfterKeywords" rule does not distinguish between 'else' and 'else if'
I have this jscs rule:
`disallowSpaceAfterKeywords: ["if", "else", "for", "while", "do", "switch", "try", "catch"]`

An error this produces is:
`Unexpected whitespace between else and if`

That error is incorrect - there *must* be a space between else and if for an 'else if' condition.
There should be an exception added for 'else if'.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13343769-disallowspaceafterkeywords-rule-does-not-distinguish-between-else-and-else-if?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1346

### DIFF
--- a/lib/rules/disallow-space-after-keywords.js
+++ b/lib/rules/disallow-space-after-keywords.js
@@ -63,9 +63,15 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         file.iterateTokensByTypeAndValue('Keyword', this._keywords, function(token) {
+            var nextToken = file.getNextToken(token);
+
+            // Make an exception if the next token is also a keyword (a space would be required).
+            if (nextToken.type === 'Keyword') {
+                return;
+            }
             errors.assert.noWhitespaceBetween({
                 token: token,
-                nextToken: file.getNextToken(token)
+                nextToken: nextToken
             });
         });
     }
--- a/test/specs/rules/disallow-space-after-keywords.js
+++ b/test/specs/rules/disallow-space-after-keywords.js
@@ -42,4 +42,14 @@ describe('rules/disallow-space-after-keywords', function() {
 
         assert(checker.checkString('for(var i in a){}').isEmpty());
     });
+
+    it('should not report illegal space after else when using else if #1346', function() {
+        checker.configure({ disallowSpaceAfterKeywords: ['else'] });
+        assert(checker.checkString('if(x){} else if(x){} else{}').isEmpty());
+    });
+
+    it('should not report illegal space when the next token is also a keyword #1346', function() {
+        checker.configure({ disallowSpaceAfterKeywords: ['return'] });
+        assert(checker.checkString('return void(0);').isEmpty());
+    });
 });


lib/rules/disallow-space-after-keywords.js
```diff
@@ -63,9 +63,13 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         file.iterateTokensByTypeAndValue('Keyword', this._keywords, function(token) {
+            var nextToken = file.getNextToken(token);
+            if (token.value === 'else' && nextToken.value === 'if') {
```
mikesherov: Perhaps use something like what is found here: fef808ea42809fb7ea8d9244913bfcb478c55313

mikesherov: That approach will also catch other cases like: `return void(0);`

hzoo: Ok, so something like `if (nextToken.type === 'Keyword') { return; }`?

mikesherov: right.

hzoo:
Ok added an comment for this and extra test.

test/specs/rules/disallow-space-after-keywords.js
```diff
@@ -1,7 +1,7 @@
 var Checker = require('../../../lib/checker');
 var assert = require('assert');
 
-describe('rules/disallow-space-after-keywords', function() {
+describe.only('rules/disallow-space-after-keywords', function() {
```
zxqfox: only?

hzoo: man.. I keep forgetting this

mikesherov: By the way, perhaps we should disallow this identifier name... If only we knew of a linter that provided this functionality.... ;-)

hzoo: Lol so I thought of that too since it's annoying but forgot that we could
just use the rule!

On Tuesday, May 12, 2015, Mike Sherov <notifications@github.com> wrote:

> In test/specs/rules/disallow-space-after-keywords.js
> <https://github.com/jscs-dev/node-jscs/pull/1364#discussion_r30147217>:
>
> > @@ -1,7 +1,7 @@
> >  var Checker = require('../../../lib/checker');
> >  var assert = require('assert');
> >
> > -describe('rules/disallow-space-after-keywords', function() {
> > +describe.only('rules/disallow-space-after-keywords', function() {
>
> By the way, perhaps we should disallow this identifier name... If only we
> knew of a linter that provided this functionality.... ;-)
>
> —
> <https://github.com/jscs-dev/node-jscs/pull/1364/files#r30147217>.
>


-- 
-Henry Zhu
Georgia Institute of Technology
Industrial Engineering '14
Computer Science Minor

hzoo: I don't think we currently have a way to support disallow `describe.only` together - only `describe` and `.only`. That would be an enhancement?

PR #1369, opened 2015-05-13
(disallow|require)SpacesInFunctionDeclaration: check export default f…

hzoo:
…unction

WIP? If we don't want to do this now that's cool since it's before it's needed.

Basically `export default function() {}` is a `FunctionExpression` right now but when we update `esprima` it's going to be `FunctionDeclaration`. After this I think I can push `babel-jscs`?

So thus it would have no `node.id` and should be skipped in this rule.

I commented out some test's since they won't pass without updating `esprima` at the moment. Or we could just remove them.

For #1376

### DIFF
--- a/lib/rules/disallow-spaces-in-function-declaration.js
+++ b/lib/rules/disallow-spaces-in-function-declaration.js
@@ -75,6 +75,10 @@ module.exports.prototype = {
         var beforeOpeningCurlyBrace = this._beforeOpeningCurlyBrace;
 
         file.iterateNodesByType(['FunctionDeclaration'], function(node) {
+            // Exception for `export default function` #1376
+            if (!node.id) {
+                return;
+            }
 
             if (beforeOpeningRoundBrace) {
                 var functionToken = file.getFirstNodeToken(node.id);
--- a/test/specs/rules/disallow-spaces-in-function-declaration.js
+++ b/test/specs/rules/disallow-spaces-in-function-declaration.js
@@ -21,6 +21,10 @@ describe('rules/disallow-spaces-in-function-declaration', function() {
             assert(checker.checkString('function abc (){}').getErrorCount() === 1);
         });
 
+        it('should not report missing space before round brace in export default function', function() {
+            checker.configure({ esnext: true });
+            assert(checker.checkString('export default function(){}').isEmpty());
+        });
     });
 
     describe('beforeOpeningCurlyBrace', function() {
@@ -36,8 +40,9 @@ describe('rules/disallow-spaces-in-function-declaration', function() {
             assert(checker.checkString('function abc() {}').getErrorCount() === 1);
         });
 
-        it('should not report missing space before round brace without option', function() {
-            assert(checker.checkString('function abc (){}').isEmpty());
+        it('should not report missing space before curly brace in export default function', function() {
+            checker.configure({ esnext: true });
+            assert(checker.checkString('export default function(){}').isEmpty());
         });
     });
 });
--- a/test/specs/rules/require-spaces-in-function-declaration.js
+++ b/test/specs/rules/require-spaces-in-function-declaration.js
@@ -21,6 +21,10 @@ describe('rules/require-spaces-in-function-declaration', function() {
             assert(checker.checkString('function abc (){}').isEmpty());
         });
 
+        it('should not report space before round brace in export default function', function() {
+            checker.configure({ esnext: true });
+            assert(checker.checkString('export default function (){}').isEmpty());
+        });
     });
 
     describe('beforeOpeningCurlyBrace', function() {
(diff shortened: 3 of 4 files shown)


mrjoelkemp:
So that I understand, we're saying that exporting default functions will be the only time that a nameless function declaration is proper syntax? Otherwise, it's a syntax error.

lib/rules/disallow-spaces-in-function-declaration.js
```diff
@@ -75,6 +75,9 @@ module.exports.prototype = {
         var beforeOpeningCurlyBrace = this._beforeOpeningCurlyBrace;
 
         file.iterateNodesByType(['FunctionDeclaration'], function(node) {
+            if (!node.id) {
```
mrjoelkemp: Please move the description of this PR to an issue (the explanation of why this needs to happen) and then reference the issue number in a comment above this line. The reasoning as to why we need to do this should be captured in the code.

test/specs/rules/disallow-spaces-in-function-declaration.js
```diff
@@ -40,4 +40,30 @@ describe('rules/disallow-spaces-in-function-declaration', function() {
             assert(checker.checkString('function abc (){}').isEmpty());
         });
     });
+
+    describe('es6', function() {
+        beforeEach(function() {
+            checker.configure({ esnext: true });
+        });
+
+        it('should not report missing space before round brace in export default function', function() {
```
mrjoelkemp: Maybe move to: 

```js
describe('before round brace', function() {
  beforeEach(function() {
    checker.configure({ disallowSpacesInFunctionDeclaration: { beforeOpeningRoundBrace: true } });
  });

  it('should not report missing space in export default function', ...

})
```

test/specs/rules/disallow-spaces-in-function-declaration.js
```diff
@@ -40,4 +40,30 @@ describe('rules/disallow-spaces-in-function-declaration', function() {
...
+            checker.configure({ disallowSpacesInFunctionDeclaration: { beforeOpeningRoundBrace: true } });
+            assert(checker.checkString('export default function() {}').isEmpty());
+        });
+
+        it.skip('should report missing space before round brace in export default function', function() {
+            checker.configure({ disallowSpacesInFunctionDeclaration: { beforeOpeningRoundBrace: true } });
+            assert(checker.checkString('export default function(){}').getErrorCount() === 1);
+        });
+
+        it('should not report missing space before curly brace in export default function', function() {
```
mrjoelkemp: Same comment as above: move to its own describe and use a beforeEach for what's common

test/specs/rules/require-spaces-in-function-declaration.js
```diff
@@ -40,4 +40,30 @@ describe('rules/require-spaces-in-function-declaration', function() {
             assert(checker.checkString('function abc() {}').isEmpty());
         });
     });
+
+    describe('es6', function() {
+        beforeEach(function() {
+            checker.configure({ esnext: true });
+        });
+
+        it.skip('should report missing space before round brace in export default function', function() {
```
mrjoelkemp: Same comment as in previous file. Move to describe with beforeEach

markelog:
Okay, lets wait for @mrjoelkemp "LGTM"

lib/rules/disallow-spaces-in-function-declaration.js
hzoo: Ok makes sense - made the issue.

test/specs/rules/disallow-spaces-in-function-declaration.js
hzoo: Should I keep the skipped tests?

mrjoelkemp:
LGMT

zxqfox:
> Coverage decreased (-0.06%) to 99.23%

I'm aware of this. ;-(

hzoo:
Well I can't write the tests for the added code since no code will run that
part of the code until we update esprima.

On Thursday, May 14, 2015, Alexej Yaroshevich <notifications@github.com>
wrote:

> Coverage decreased (-0.06%) to 99.23%
>
> I'm aware of this. ;-(
>
> —
> <https://github.com/jscs-dev/node-jscs/pull/1369#issuecomment-102095468>.
>


-- 
-Henry Zhu
Georgia Institute of Technology
Industrial Engineering '14
Computer Science Minor

zxqfox:
@hzoo Np, let's just left an issue for that.

hzoo:
Sounds good! Yeah even the tests now aren't being tested correctly because of the same issue - but the missing tests would be checking for an error so I just left them out (instead of skipping).

mrjoelkemp:
Merge or no?

hzoo:
Can merge - we have an issue to add a test for coverage in #1377 for after esprima is updated to return `FunctionDeclaration` with a null id. Esprima issue is https://github.com/jquery/esprima/issues/311

hzoo:
Actually looking at it now - I think I should add the same check to `disallowSpacesInFunction` and `disallowAnonymousFunctions` (and for require)

PR #1380, opened 2015-05-15
requireSpaceBeforeBlockStatements: Fix doc

adrianheine:
`true` does not assert exactly a single space, but at least a single space – just as `1` does.

### DIFF
--- a/lib/rules/require-space-before-block-statements.js
+++ b/lib/rules/require-space-before-block-statements.js
@@ -5,7 +5,7 @@
  *
  * Values:
  *
- * - `true` require a single space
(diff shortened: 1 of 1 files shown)


hzoo:
:+1: - At some point /2.0 we should look at the consistency of all these rules in enforcing exactly 1 or atLeast 1. I think the reason why we did this was because of people putting comments in between but now we have that helper method for that - `file.getPrevToken(lastToken, {includeComments: true});`.

markelog:
@hzoo could create an issue for that, so we wouldn't forget?

markelog:
Thank you!

PR #1388, opened 2015-05-18
(require|disallow)SpacesIn(Function|FunctionExpression): support clas…

### LINKED ISSUE #1066: Error for es6 class method's on a single line
An error for `disallowSpacesInAnonymousFunctionExpression` for `"beforeOpeningRoundBrace": true` with es6 class.
So it's only when it's a single line. (with `esnext` and `esprima-fb`).
```
Component = class {
  render() { return <div />; }
------------^
};
Component = class {
  render() { return func(); }
};
```

This is fine
```
Component = class {
  render() {
    return <div />;
  }
};
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8585191-error-for-es6-class-method-s-on-a-single-line?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
…s methods (ES6)

Fixes #1066.

```js
Component = class {
  render() { return <div />; }
------------^
```

Issue was that it wasn't checking `render` and `(` spacing but `{` and `return` or `)` and `{`.

### DIFF
--- a/lib/rules/require-spaces-in-function-expression.js
+++ b/lib/rules/require-spaces-in-function-expression.js
@@ -89,7 +89,14 @@ module.exports.prototype = {
 
             if (beforeOpeningRoundBrace) {
                 // for a named function, use node.id
-                var functionToken = file.getFirstNodeToken(node.id || node);
+                var functionNode = node.id || node;
+
+                // for a class method
+                if (parent.type === 'MethodDefinition' && parent.key.type === 'Identifier') {
+                    functionNode = parent.key;
+                }
+
+                var functionToken = file.getFirstNodeToken(functionNode);
                 errors.assert.whitespaceBetween({
                     token: functionToken,
                     nextToken: file.getNextToken(functionToken),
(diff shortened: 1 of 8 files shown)


lib/rules/require-spaces-in-function-expression.js
```diff
@@ -89,7 +89,14 @@ module.exports.prototype = {
 
             if (beforeOpeningRoundBrace) {
                 // for a named function, use node.id
-                var functionToken = file.getFirstNodeToken(node.id || node);
+                var functionNode = node.id || node;
```
markelog: Hm, maybe we need to add helper method for that?

markelog:
I will merge this as-is, but it seems

```js
var functionNode = node.id || node;

// for a class method
if (parent.type === 'MethodDefinition' && parent.key.type === 'Identifier') {
    functionNode = parent.key;
}
```

this could be really moved to the helper

PR #1397, opened 2015-05-21
New Rule: requireNumericLiterals

### LINKED ISSUE #1386: New Rule: prefer using binary/octal/hex literals (ES6)
And opposite rule too?

```js
// good
0b111110111 === 503;
0o767 === 503;
0x1F7 === 503;
// bad
parseInt("111110111", 2) === 503;
parseInt("767", 8) === 503;
parseInt("1F7", 16) === 255;
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15543154-new-rule-prefer-using-binary-octal-hex-literals-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1386 

Not sure about the name again

### DIFF
--- a/lib/rules/require-numeric-literals.js
+++ b/lib/rules/require-numeric-literals.js
@@ -1,0 +1,73 @@
+/**
+ * Requires use of binary, hexadecimal, and octal literals instead of parseInt.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireNumericLiterals": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * 0b111110111 === 503;
+ * 0o767 === 503;
+ * 0x1F7 === 503;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * parseInt("111110111", 2) === 503;
+ * parseInt("767", 8) === 503;
+ * parseInt("1F7", 16) === 255;
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+
+        this._radixMap = {
+            2: 'binary',
+            8: 'octal',
+            16: 'hexadecimal'
+        };
+    },
+
+    getOptionName: function() {
+        return 'requireNumericLiterals';
+    },
+
+    check: function(file, errors) {
+        var radixMap = this._radixMap;
+        file.iterateNodesByType(['CallExpression'], function(node) {
+            // don't check for parseInt(1)
+            if (node.arguments.length !== 2) {
+                return;
+            }
+
+            // only error if the radix is 2, 8, or 16
+            var radixName = radixMap[node.arguments[1].value];
+            if (node.callee.type === 'Identifier' &&
+                node.callee.name === 'parseInt' &&
+                radixName
+            ) {
+                errors.add('Use ' + radixName + ' literals instead of parseInt', node.loc.start);
+            }
+        });
+    }
+};
--- a/test/specs/rules/require-numeric-literals.js
+++ b/test/specs/rules/require-numeric-literals.js
@@ -1,0 +1,40 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/require-numeric-literals', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+
+        checker.configure({
+            esnext: true,
+            requireNumericLiterals: true
+        });
+    });
+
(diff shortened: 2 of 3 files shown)


lib/rules/require-numeric-literals.js
```diff
@@ -0,0 +1,72 @@
...
+    getOptionName: function() {
+        return 'requireNumericLiterals';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType(['CallExpression'], function(node) {
+            var callee = node.callee;
+            var args = node.arguments;
+
+            if (args.length !== 2) {
```
mrjoelkemp: Please comment about when this would be the case. A code sample would suffice

lib/rules/require-numeric-literals.js
```diff
@@ -0,0 +1,72 @@
...
+                return;
+            }
+
+            var stringArg = args[0];
+            var radix = args[1];
+
+            if (callee.type === 'Identifier' &&
+                callee.name === 'parseInt' &&
+                typeof stringArg.value === 'string') {
+                if (radix.value === 2) {
```
mrjoelkemp: Create a map of bases:

```js
{
  2: 'binary',
  8: 'octal',
  16: 'hexadecimal'
}
```

Then this entire conditional gets replaced by:

```js
errors.add('Use ' + bases[radix.value] + ' literals instead of parseInt', node.loc.start);`
```

lib/rules/require-numeric-literals.js
hzoo: I guess this is just a return early check since you can have many `CallExpressions` like `a(), a(b), a(b,c,d)` - it's not specifically needed but we do need to know at least 2 exist. Should I just put something like that there?

test/specs/rules/require-numeric-literals.js
```diff
@@ -0,0 +1,37 @@
...
+        assert(checker.checkString('0b111110111 === 503;').isEmpty());
+        assert(checker.checkString('0o767 === 503;').isEmpty());
+        assert(checker.checkString('0x1F7 === 503;').isEmpty());
+    });
+
+    it('should not report on other call expressions', function() {
+        assert(checker.checkString('a();').isEmpty());
+        assert(checker.checkString('a(1);').isEmpty());
+        assert(checker.checkString('a(1, 2);').isEmpty());
+        assert(checker.checkString('parseInt(1, 2);').isEmpty());
```
mrjoelkemp: It's confusing as to why this is valid. I thought the rule (based on its name) was to ban the use off `parseInt` entirely. 

Now, I see that it's only for predicates/conditions. The rule's name should be adjusted to mention this.

hzoo: Sorry that is a bad example - I think it's because I'm checking that the first argument is a string and in that case it's not but will become a string anyway so that should error. It would be simplier to just disallow parseInt but I guess when I was thinking of the rule I was thinking it would only discourage use when it could be a binary, hex, octal literal instead (so it wouldn't error for different radix?

mrjoelkemp: > I was thinking it would only discourage use when it could be a binary, hex, octal literal instead

Sounds good

lib/rules/require-numeric-literals.js
mrjoelkemp: Yeah, a comment along the lines of `// Prevent calls like parseInt(1), ...` would help clarify this condition

test/specs/rules/require-numeric-literals.js
hzoo: Disallowing `parseInt` could actually just be done with `disallowIdentifiers`

mrjoelkemp: Without proper use cases from the community, it feels like we're guessing as to what people need. 

Are you saying that this rule isn't worth including?

hzoo: No I was just saying that as a solution to your point before. I think this rule is still valid to enforce the literal style rather than parseInt.

hzoo:
Ok made the changes and removed the temp variables since I think its fine without them.

lib/rules/require-numeric-literals.js
```diff
@@ -0,0 +1,70 @@
...
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireNumericLiterals';
+    },
+
+    check: function(file, errors) {
+        var radixMap = {
```
mrjoelkemp: Unnecessary object creation per file checked. Move to this.radixMap in configure.

mrjoelkemp:
Lgtm

mdevils:
LGTM, but test failed, take a look please.

mrjoelkemp:
Thanks @hzoo. Nice work

Issue #1385, opened 2015-05-17
New Rule: shorter object syntax in arrow function (ES6)


hzoo:
No idea what the name for this is, but can have both rules.

```js
// bad
  [1, 2, 3].map((x) => {
    return { number: x };
  });

// good
  [1, 2, 3].map((x) => ({ number: x });
```
an example from http://tc39wiki.calculist.org/es6/arrow-functions/

```js
// Parenthesize the body to return an object literal expression
let key_maker = val => ({key: val});
```

alawatthe:
Do you only want to disallow the return statement when returning an object or in all single expression arrow functions? I.e something like

```js
// bad
x => {return 42;}

// good
x => 42
```

Name suggestion for the second case: disallowReturnInSingleExpressionArrowFunctions Does this win the price of the longest rule name so far? ;-)

mrjoelkemp:
(require|disallow)ImplicitReturn or the same with InArrowFunction for
specificity (if necessary)?

hzoo:
Related to #1340 - `requireShorthandArrowFunctions`

```js
// good
evens.map(v => v + 1);
// bad
evens.map(v => { return v + 1; });
```

This is specifically for returning a single object. 

Actually maybe we should just merge it into a single rule.

ljharb:
Whatever you end up with, I'd want to be able to configure rules so that I could:
 - allow implicit return of non-object-literals
 - disallow implicit return of paren-wrapped object literals
 - disallow implicit return of blocks with labelled statements

and, potentially, disallow implicit return entirely, or, allow implicit return entirely. (in case that helps make "picking a rule name" harder :-) )

hzoo:
Could some of those options could become their own rules? Or only apply within arrow functions?

Maybe we should have a `disallowLabelStatments` rule.. - (had to look at up haha).

ljharb:
We absolutely should. Those should only be used intentionally imo, and that's what linters should catch :-)

hzoo:
We released [3.0](https://github.com/jscs-dev/node-jscs/releases/tag/v3.0.0) recently! We also wrote a [post](https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.oot3wvgpu) about how 3.0 will be our last major release (since we are deprecating the project) and that we are [merging with ESLint](http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint)!

[...]

PR #1398, opened 2015-05-21
New Rule: disallowNodeTypes

### LINKED ISSUE #1389: New Rule: disallowNodeTypes
Ref https://github.com/jscs-dev/node-jscs/issues/1385#issuecomment-102881542.
> Maybe we should have a disallowLabelStatements rule

So instead of a one off rule (`disallowLabelStatements`), we can do the same as `disallowIdentifierNames`, and `disallowKeywords`? Although I don't know what other statements we would want to disallow that is like this (since most would be keywords)

Need a better name - @ljharb?

Would take an array of strings from: https://github.com/jquery/esprima/blob/master/esprima.js#L108-169

`"disallowASTStatements": ["LabelStatement"]`
`"disallowNodeTypes": ["LabelStatement"]`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15742538-new-rule-disallownodetypes?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1389 

Need some help here.

Esprima node types: https://github.com/jquery/esprima/blob/758196a1c5dd20c3ead6300283a1112428bc7045/esprima.js#L108-L169

Acorn/babel node types:
https://github.com/babel/babel/blob/65a44a1e131c8bc336b218991bae7548176de283/src/babel/types/alias-keys.json#L2-L121

Espree node types: https://github.com/eslint/espree/blob/a23b65db10648daf7d4b0761b0ce09eea864550a/lib/ast-node-types.js#L41-L115

Do we want to check if the node types given in the array are supported by the parser or just let it error?

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -719,6 +719,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-spaces-in-for-statement'));
     this.registerRule(require('../rules/disallow-spaces-in-for-statement'));
 
+    this.registerRule(require('../rules/disallow-node-types'));
+
     this.registerRule(require('../rules/disallow-keywords-in-comments'));
 
     this.registerRule(require('../rules/disallow-identifier-names'));
--- a/lib/rules/disallow-node-types.js
+++ b/lib/rules/disallow-node-types.js
@@ -1,0 +1,64 @@
+/**
+ * Disallow use of certain node types (from Esprima/ESTree).
+ * Esprima node types
+ *  - [list](https://github.com/jquery/esprima/blob/758196a1c5dd20c3ead6300283a1112428bc7045/esprima.js#L108-L169)
+ *
+ * Type: `Array`
+ *
+ * Value: Array of parser node types to be disallowed.
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowNodeTypes": ['LabeledStatement']
+ * ```
+ *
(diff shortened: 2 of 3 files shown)


ljharb:
\( ﾟ◡ﾟ)/ 👍

mrjoelkemp:
This feels like a JSHint rule.

hzoo:
Yeah that's true. However, we do have disallowIdentifiers, and disallowKeywords that are used with the intent to disallow parts as well (`with`, certain methods like `eval` or something) so this rule is created with a similar purpose?

mrjoelkemp:
Sure then. DisallowKeywords was likely discussed at length and accepted so
this is fine by me.

markelog:
mm, where broad rule, couldn't hurt though

hzoo:
@markelog was from https://github.com/jscs-dev/node-jscs/issues/1385#issuecomment-102881542 - originally to disallow `LabeledStatements`

markelog:
Yeah, i read it, there might be use-cases for that, sure, it feels as advanced rule though

hzoo:
Yeah that's definitely true, since you have to know what node types are/esprima/ASTs.

PR #1401, opened 2015-05-23
Reporters: make inline and inlinesingle machine readable

andreastt:
Changes the formatting of the reporters `inline` and `inlinesingle`
from

	file.js: line 7, col 7, Unexpected token {

to

	file.js:7:7: Unexpected token {

which is much like the output from other compilers and preprocessors.

The benefit is integration with existing tools that parse similar
output.  For example Emacs allows navigation by these “hyperlinks”
through the rules in `compilation-error-regexp-alist`.  Other
editors, like acme, lets you right-click on such tokens to jump to
the file, line, and column.

**Please let me know if you’d rather see this implemented as a separate, new reporter.  Unless there are any backwards compatibility concerns, I do however think that this approach is uncontroversial.**

### DIFF
--- a/lib/reporters/inline.js
+++ b/lib/reporters/inline.js
@@ -1,9 +1,11 @@
 var util = require('util');
+
 /**
  * @param {Errors[]} errorsCollection
  */
 module.exports = function(errorsCollection) {
     var errorCount = 0;
+
     /**
      * Formatting every error set.
      */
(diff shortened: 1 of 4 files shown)


mdevils:
Hello @andreastt. Thank you for contribution.

I'm afraid `inline` and `inlinesingle` was created in compliance with `jshint` default reporter and changing their output would break backwards compatibility. But the idea of having a new reporter is discussable.

andreastt:
@mdevils Do you have any preference on a name for such a new reporter?

andreastt:
Closing this in favour of #1403.

PR #1403, opened 2015-05-23
Reporters: add new machine readable unix-style reporter

andreastt:
Similar to `inline` and `inlinesingle`, but instead of outputting

	file.js: line 7, col 7, Unexpected token {

it outputs

	file.js:7:7: Unexpected token {

which is much like the output from other compilers and preprocessors.

The benefit is integration with existing tools that parse similar
output.  For example Emacs allows navigation by these “hyperlinks”
through the rules in `compilation-error-regexp-alist`.  Other editors
such as acme, lets you right-click on such tokens to jump to the
file, line, and column.

Closes gh-1403

### DIFF
--- a/lib/reporters/unix.js
+++ b/lib/reporters/unix.js
@@ -4,13 +4,10 @@ var util = require('util');
  * @param {Errors[]} errorsCollection
  */
 module.exports = function(errorsCollection) {
-    var errorCount = 0;
-
     errorsCollection.forEach(function(errors) {
         var file = errors.getFilename();
         if (!errors.isEmpty()) {
             errors.getErrorList().forEach(function(error) {
-                errorCount++;
                 console.log(util.format('%s:%d:%d: %s', file, error.line, error.column, error.message));
             });
         }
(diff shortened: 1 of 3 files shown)


andreastt:
I should add that jshint has a similar [“unix” reporter](https://github.com/jshint/jshint/blob/master/src/reporters/unix.js) done by me.

mdevils:
Good job @andreastt!
Looks good to me.
/cc @markelog

lib/reporters/unix.js
```diff
@@ -0,0 +1,18 @@
+var util = require('util');
+
+/**
+ * @param {Errors[]} errorsCollection
+ */
+module.exports = function(errorsCollection) {
+    var errorCount = 0;
```
mrjoelkemp: Unused

andreastt:
@mrjoelkemp Removed the unused variable.

mrjoelkemp:
Lgtm

markelog:
I rarely land PRs without looking at the code, but when i do, it is always with two "LGTM" comments.

Thank you!

Issue #1159, opened 2015-03-11
Test targets explicitly provided which are listed in `.jscsignore`


zxqfox:
1. `git` doesn't add files by default even if they targeted explicitly, paths to which are listed in `.gitignore`: `echo 'something' >> .gitignore; git add something`;
2. `jshint` (more importantly) act the same as git and doesn't check explicitly provided targets if they are listed in `.jshintignore`;
3. `jscs` atm has `excludedFiles` (as analogue `.jshintignore`/`.gitignore`), but it checks explicit targets since `v1.5.6` — inconsistency at my pov.

Can we rethink that and go `jshint`/`git` way?

/cc @markelog @mikesherov @mdevils @mrjoelkemp @hzoo @indexzero

hzoo:
Sounds good to me - JSHint also lets you do `jshint . --exclude-path .gitignore`

PR #1405, opened 2015-05-24
Make check on "fileExtensions" and "excludeFiles" for specified path.

aliosv:


mrjoelkemp:
Thanks @aliosv. Looks like a few tests broke with this PR.

markelog:
@aliosv could you elaborate your issues and please add tests to this PR

aliosv:
@markelog, pre-commit hook run this script https://github.com/bem/bem-components/blob/v2/.githooks/pre-commit/lint, but there is a problem: jscs will check all staged files, not taking into account own config.

markelog:
Still don't understand the issue, you saying `excludeFiles` doesn't work?

Could provide tests for this pull? With them, it would be easier to grasp exactly what do you trying to fix.

aliosv:
excludeFiles and fileExtesions doesn't work for specified path, i.e. ````jscs <path>```` will check the file regardless of the config.
I will add tests as soon as i can.

aliosv:
added tests

test/specs/checker.js
```diff
@@ -43,9 +43,29 @@ describe('modules/checker', function() {
...
+
+            it('should return empty array of errors for excluded file', function() {
+                checker.configure({ excludeFiles: ['./test/**'] });
+
+                return checker.checkPath('./test/data/checker/file.js').then(function(errors) {
+                    assert(errors.length === 0);
+                });
+            });
+
+            it('should return empty array of errors for file with not registered extension', function() {
```
markelog: Without changes in the source, only this test is falling whereas this behaviour is by design, we open to discuss this issue, but it seems point of this pull should be about another matter?

markelog:
@aliosv still interested?

aliosv:
@markelog, yes, but i do not see what to discuss. PR adds new feature: check specified path to file by config settings, it's broke backward compatibility, but you still have possibility to check any file using stdin. Maybe you know how to solve my issue in other way?

markelog:
I don't understand what is the issue, tests your added are passing without source changes

aliosv:
@markelog, https://github.com/bem/bem-components/blob/v2/.githooks/pre-commit/lint#L27
на пре-коммит хуке jscs может валидировать файлы, которые запрещены в его конфиге, этого не хочется.

markelog:
Please use english, otherwise others can't participate in the discussion.

I'm afraid you are not to descriptive, link to the repo doesn't explain anything, please add failing tests to this pr or create a repo where possible bug could be reproduced.

aliosv:
@markelog, test 'should return empty array of errors for excluded file' should fail.

markelog:
So can you add a falling test?

aliosv:
@markelog, https://github.com/aliosv/node-jscs-test project to test. Also i find my error: each file is checked on excluding, so problem only in cheking on allowed extension.

markelog:
> fails the test that you commented(https://github.com/jscs-dev/node-jscs/pull/1405/files#diff-9511adab08ed83a63c8bb84eb0d54617R63). 

Just like i said - this is by design, for example if you do `eslint test.json`, or `jshint test.json` without specifying additional extensions for either of those tools, they will still print the errors, but if you specify `test.json` either in `.jshintignore` or in `.eslintignore` they will be ignored.

We follow common practise here.

zxqfox:
@markelog If file specified in `.someignore` file it should be ignored even if specified as argument. Right? Our `excludeFiles` doesn't skip files if you pass them directly as arguments.

There was an issue about that #1159

aliosv:
@zxqfox, 
>Our excludeFiles doesn't skip files if you pass them directly as arguments.

in my test - skip(https://github.com/aliosv/node-jscs-test/tree/excludeFiles), check please.

markelog:
@zxqfox 

> Our excludeFiles doesn't skip files if you pass them directly as arguments.

Could provide a test-case?

@aliosv Thank you for understanding.

PR #1410, opened 2015-05-29
disallowCommaBeforeLineBreak: added ignoreFunction, lineBreak options

dreef3:
Hi,

I needed to handle two corner cases in my project with disallowCommaBeforeLineBreak rule:
1. I wanted to have a line break before comma only in simple objects without methods
2. Fixed version would place a comma and both tokens before and after it on a single line while I wanted to have comma at the beginning of the next line

So I've added two options to the rule:
```javascript
{
    mode: ["all", "ignoreFunction"],
    lineBreak: ["none", "beforeComma"]
}
```

By default the rule would behave the same way as before (that is, accept `true` instead of options). Example for beforeComma:
*Invalid:*
```javascript
var y = { three: 3, four: 4};
```
*Valid:*
```javascript
var y = {
    three: 3
    , four: 4
};
```

### DIFF
--- a/lib/rules/disallow-comma-before-line-break.js
+++ b/lib/rules/disallow-comma-before-line-break.js
@@ -5,12 +5,10 @@
  *
  * Values:
  *  - `true` for default behavior (strict mode, comma on the same line)
- *  - `"mode"` options:
- *    - `"all"` for strict mode
- *    - `"ignoreFunction"` ignores objects if one of the property values is a function expression
- *  - `"lineBreak"` options:
- *    - `"none"` comma can be on the same line with both operands
- *    - `"beforeComma"` comma must have a line break after it's first operand
+ *  - `Object`:
+ *    - `'lineBreak'` requires a line break between object key and comma
+ *    - `'allExcept'` array of exceptions:
+ *       - `'function'` ignores objects if one of their values is a function expression
  *
  * JSHint: [`laxcomma`](http://www.jshint.com/docs/options/#laxcomma)
  *
@@ -20,7 +18,7 @@
  * "disallowCommaBeforeLineBreak": true
  * ```
  *
- * ##### Valid
+ * ##### Valid for `true`
  *
  * ```js
  * var x = {
@@ -39,13 +37,16 @@
  * };
  * ```
  *
+ * ##### Valid for `{"allExcept": ["function"]}`
+ *
  * ```js
- * "disallowCommaBeforeLineBreak": {
- *     "lineBreak": "beforeComma"
- * }
+ * var x = {
+ *     one: 1,
+ *     two: function() {}
+ * };
  * ```
  *
- * ##### Valid
+ * ##### Valid for `{"lineBreak": true}`
  *
  * ```js
  * var x = {
@@ -54,7 +55,7 @@
  * };
  * ```
  *
- * ##### Invalid
+ * ##### Invalid for `{"lineBreak": true}`
  *
  * ```js
  * var y = { three: 3, four: 4};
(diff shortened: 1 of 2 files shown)


test/specs/rules/disallow-comma-before-line-break.js
```diff
@@ -60,4 +60,46 @@ describe('rules/disallow-comma-before-line-break', function() {
     it('should not report comma placement in single line object declaration', function() {
         assert(checker.checkString('var a = {a:1, c:3};').isEmpty());
     });
+
+    describe('mode option', function() {
+        describe('ignoreFunction value', function() {
+            beforeEach(function() {
+                checker.configure({
+                    disallowCommaBeforeLineBreak: {
+                        mode: 'ignoreFunction'
```
zxqfox: mmm, `allExcept: 'function'`?

lib/rules/disallow-comma-before-line-break.js
```diff
@@ -40,10 +40,35 @@ module.exports = function() {};
 module.exports.prototype = {
```
zxqfox: Need to update docs.

And I'm not sure I understand what beforeComma option really means ;-( Perhaps there are better names.

test/specs/rules/disallow-comma-before-line-break.js
hzoo: Example of rule using allExcept: https://github.com/jscs-dev/node-jscs/blob/f55bbae8bfbd8d57a4f756193d824e6f398ff7a7/lib/rules/require-space-after-line-comment.js#L10

dreef3:
@zxqfox ,

As I understand `allExcept` would skip only a single property in the object. I took this `ignoreFunction` from requireAlignedObjectValues, where the same option used to skip the object entirely. Not sure if this option would be useful to others. I could also add `ignoreLineBreak` from requireAlignedObjectValues to make it look consistent.

`beforeComma` does what shown in examples: it places a line break before comma when fix applied. By default this rule places all items separated by `,` into single line. Example:
```javascript
// before
var y = require('foo'),
      z = require('bar');
``` 
```javascript
// after
var y = require('foo'), z = require('bar');
```

With `beforeComma` it would become:
```javascript
// after
var y = require('foo')
    , z = require('bar');
```

Could please help me to find a better name for this option?

dreef3:
Updated to use `allExcept`.

lib/rules/disallow-comma-before-line-break.js
```diff
@@ -51,12 +90,47 @@ module.exports.prototype = {
...
+            var node = file.getNodeByRange(token.range[0]);
+
+            if (node.type !== 'ObjectExpression' ||
+                node.loc.start.line === node.loc.end.line ||
+                node.properties < 2) {
+                return false;
+            }
+
+            return node.properties.some(function(property) {
+                if (exceptFunction && property.value.type === 'FunctionExpression') {
```
mrjoelkemp: `return exceptFunction && property.value.type === 'FunctionExpression'`

lib/rules/disallow-comma-before-line-break.js
```diff
@@ -51,12 +90,47 @@ module.exports.prototype = {
...
+        var exceptFunction = this._exceptFunction;
+        var lineBreak = this._lineBreak;
+
+        function canSkip(token) {
+            if (!exceptFunction) {
+                return false;
+            }
+            var node = file.getNodeByRange(token.range[0]);
+
+            if (node.type !== 'ObjectExpression' ||
```
mrjoelkemp: What is this entire condition describing? Can you leave a comment with an example that would satisfy this condition?

lib/rules/disallow-comma-before-line-break.js
```diff
@@ -31,6 +36,30 @@
...
+ * ##### Valid for `{"allExcept": ["function"]}`
+ *
+ * ```js
+ * var x = {
+ *     one: 1,
+ *     two: function() {}
+ * };
+ * ```
+ *
+ * ##### Valid for `{"lineBreak": true}`
```
mrjoelkemp: Seems like this could be `allExcept: ['singleLine']`

dreef3: @mrjoelkemp,
Do you think this `lineBreak` could be a default behavior for multiline ObjectExpression? For example from current docs:

##### Valid
```js
var x = {
    one: 1
    , two: 2
};
```

Right now fixed variant of this rule is always single line: `{one: 1, two: 2}`

lib/rules/disallow-comma-before-line-break.js
dreef3: This expression verifies that we check only multiline objects with more that one property:
* `allExcept: ['function']` does lookup only inside ObjectExpression
* No point checking singleline objects for line breaks
* No comma if there's only a single property

dreef3:
Fixed review comments.

Also I've changed logic a little bit, please correct me if I made wrong assumptions:
* No point to check for line breaks in singleline expression so I added skip check for these
* I made what previously was a `lineBreak` option a default behavior for multiline objects. Looks like it satisfies examples in [current](http://jscs.info/rule/disallowCommaBeforeLineBreak.html) version of the docs

markelog:
It seems we good here?

zxqfox:
Yeah, looks good to me.

mrjoelkemp:
Great work here @dreef3. Thanks again for contributing.

PR #1415, opened 2015-05-30
Add "fix" field in rule declaration

markelog:
Addresses https://github.com/jscs-dev/node-jscs/issues/1201 https://github.com/jscs-dev/node-jscs/pull/1202 https://github.com/jscs-dev/node-jscs/pull/1265, but most importantly #1113.

Example with `validateQuoteMarks` is just that, example.

Since this pull incorporates #1150, diff is pretty big, i will create 2.0 branch merge #1150 and change target branch of this pull, it should make everything more clearer, meanwhile you could use  <a href="https://github.com/markelog/node-jscs/compare/8192eab...fb450">this</a> link instead.

This pull represents two different ideas for autofix infrastructure - "fix" field and assertion interface. 

Which is bad, we need to make them complimentary to each other or unite approaches, but we need to keep ball rolling need to release 2.0 soon, we can improve/refactor later.

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -220,6 +220,17 @@ Configuration.prototype.getConfiguredRules = function() {
 };
 
 /**
+ * Returns configured rule.
+ *
+ * @returns {Rule | null}
+ */
+Configuration.prototype.getConfiguredRule = function(name) {
+    return this._configuredRules.filter(function(rule) {
+        return rule.getOptionName() === name;
+    })[0] || null;
+};
+
+/**
  * Returns the list of unsupported rule names.
  *
  * @return {String[]}
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -60,23 +60,8 @@ Configuration.prototype.load = function(config) {
     copyConfiguration(overrides, currentConfig);
 
     var ruleSettings = this._processConfig(currentConfig);
-    var processedSettings = {};
 
-    Object.keys(ruleSettings).forEach(function(optionName) {
-        var rule = this._rules[optionName];
-        if (rule) {
-            var optionValue = ruleSettings[optionName];
-            if (optionValue !== null) {
-                rule.configure(ruleSettings[optionName]);
-                this._configuredRules.push(rule);
-                processedSettings[optionName] = ruleSettings[optionName];
-            }
-        } else {
-            this._unsupportedRuleNames.push(optionName);
-        }
-    }, this);
-
-    this._ruleSettings = processedSettings;
+    this._loadRules(ruleSettings);
 };
 
 /**
@@ -285,19 +270,6 @@ Configuration.prototype._processConfig = function(config) {
         options.plugins.forEach(this._loadPlugin, this);
     }
 
-    // Apply presets
-    var presetName = options.preset;
-    if (presetName) {
-        this._presetName = presetName;
-        assert(typeof presetName === 'string', '`preset` option requires string value');
-        var presetData = this._presets[presetName];
-        assert(Boolean(presetData), 'Preset "' + presetName + '" does not exist');
-        var presetResult = this._processConfig(presetData);
-        Object.keys(presetResult).forEach(function(key) {
-            ruleSettings[key] = presetResult[key];
-        });
-    }
-
     // File extensions
     if (options.fileExtensions) {
         assert(
@@ -367,6 +339,11 @@ Configuration.prototype._processConfig = function(config) {
         this._loadVerbose(options.verbose);
     }
 
+    // Apply presets
+    if (options.hasOwnProperty('preset')) {
+        this._loadPreset(options.preset);
+    }
+
     // NOTE: rule setting must come last in order to
     // override any rules that are loaded from a preset
     Object.keys(config).forEach(function(key) {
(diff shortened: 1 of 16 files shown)


markelog:
Intergration tests for the win! Since they helped to find to discover <a href="https://github.com/sindresorhus/to-single-quotes/issues/6">this</a> issue, tests are failing because of that

lib/errors.js
```diff
@@ -2,6 +2,14 @@ var assert = require('assert');
 var chalk = require('chalk');
 var TokenAssert = require('./token-assert');
 
+function assertChecker(line, column) {
```
mrjoelkemp: `assertPosition` is a clearer name

lib/errors.js
```diff
@@ -51,6 +55,23 @@ Errors.prototype = {
     },
 
     /**
+     * Adds style error to the list
+     *
+     * @param {Object} errorInfo
+     */
+    cast: function(errorInfo) {
```
mrjoelkemp: Not seeing the need for this additional public function. Why not put the checks in `_addError`?

If you want to remove that logic to avoid cluttering `_addError`, then maybe put it in a `_validateError` function called within `_addError`

lib/string-checker.js
```diff
@@ -126,6 +133,41 @@ StringChecker.prototype = {
...
     /**
+     * Fix provided error.
+     *
+     * @param {JsFile} file
+     * @param {Errors} errors
+     * @protected
+     */
+    _fixJsFile: function(file, errors) {
+        var list = errors.getErrorList();
+        var configration = this.getConfiguration();
```
mrjoelkemp: typo: `configuration`

test/specs/config/node-configuration.js
```diff
@@ -54,6 +90,76 @@ describe('modules/config/node-configuration', function() {
     });
 
     describe('load', function() {
+        it('should load existed preset', function() {
```
mrjoelkemp: `existing`

lib/string-checker.js
```diff
@@ -126,6 +133,41 @@ StringChecker.prototype = {
...
+
+            var instance = configration.getConfiguredRule(error.rule);
+
+            if (instance && instance.fix) {
+                try {
+
+                    // "error.fixed = true" should go first, so rule can
+                    // decide for itself (with "error.fixed = false")
+                    // if it can fix this particular error
+                    error.fixed = true;
```
mrjoelkemp: 1. What about adding this to the Error constructor?
2. The default should be `false` and made `true` after the call to `instance.fix(error)`.

lib/string-checker.js
```diff
@@ -126,6 +133,41 @@ StringChecker.prototype = {
...
+                    instance.fix(error);
+
+                } catch (e) {
+                    error.fixed = undefined;
```
mrjoelkemp: `false`?

package.json
```diff
@@ -73,6 +73,8 @@
     "pathval": "~0.1.1",
     "prompt": "~0.2.14",
     "strip-json-comments": "~1.0.2",
+    "to-double-quotes": "^1.0.0",
```
mrjoelkemp: The majority of our dependencies use tildes. What's the heuristic to determine if we should use caret or tilde?

lib/string-checker.js
markelog: For situation when rule decides it can't fix the error, if case is too complicated, for exampe. Then in the `fix` method it can do `error.fixed = false`.

lib/string-checker.js
markelog: Original value is `undefined`

package.json
markelog: `npm i --save`

lib/string-checker.js
mrjoelkemp: > if case is too complicated for example

My alternative implementation would be to have the rule throw an error in the fix method for unfixable errors. I guess an error being unfixable isn't an exception? What I'm getting at is avoiding rules knowing about the `fixed` property of the error.

Of course, `fixed` is public, so it doesn't matter so much.

package.json
mrjoelkemp: Ah, I changed my default to tilde. I guess we'll opt for caret first and then discuss whether or not it should be a tilde during PR review?

mrjoelkemp:
:+1:

package.json
markelog: Sounds good

lib/string-checker.js
markelog: > I guess an error being unfixable isn't an exception?

Yeah, otherwise we would show that big message, inviting user to submit a bug to this repo.

> What I'm getting at is avoiding rules knowing about the fixed property of the error.

Alternative for it would be to make a return value of the `fix` method count, if returns `undefined | true` -> `error.fixed = true` if returns `false` -> `error.fixed = undefined`, or `false` value, which, btw could signify failure (either through exception or explicit return value) of fix attempt.

mrjoelkemp: I prefer that suggestion. Nice idea!

lib/config/configuration.js
```diff
@@ -25,22 +25,148 @@ var BUILTIN_OPTIONS = {
...
+     */
     this._excludedFileMasks = [];
+
+    /**
+     * List of existing files that falls under exclusion masks.
+     *
+     * @protected
+     * @type {Array}
+     */
     this._excludedFileMatchers = [];
```
zxqfox: I think it should be moved to NodeConfiguration if we gonna use `.jscsignore` to extend with internal ignores.

mikesherov:
I'd prefer another minor release before we jump to 2.0. I'll discuss more over gitter, but I wanted to mention it here since it was mentioned in the dexdription.

lib/string-checker.js
markelog: None of the other rule method does it, so i'm not totally sure :(

markelog:
Replaced by #1430

PR #1418, opened 2015-05-31
maximumLineLength: allow function declarations to exceed limit

### LINKED ISSUE #1412: new maximumLineLength option
It would be nice if [maximumLineLength](http://jscs.info/rule/maximumLineLength.html) had an option `allowMethodDeclarations` which would exempt lines like the following. It would default to `false` for backwards compatibility.

```
function myCoolFunctionWith(descriptiveArgumentName1, descriptiveArgumentName2, descriptiveArgumentName3) {
```
or
```
  _removeSingleLeaf(rootModel, rootProp, rootInstance, leafModel, leafProp, leafInstance, watcher) {
```

I can take a pass at this in the near future.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18743673-new-maximumlinelength-option?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

clayreimann:
As is this will break existing configurations. It sounded like that was the desired change from the issue, but I can put backwards compatibility back in (with an appropriate warning message).

Fixes #1412

### DIFF
--- a/lib/rules/maximum-line-length.js
+++ b/lib/rules/maximum-line-length.js
@@ -8,7 +8,7 @@
  *  - `Object`:
  *     - `value`: (required) lines should be at most the number of characters specified
  *     - `tabSize`: (default: `1`) considered the tab character as number of specified spaces
- *     - `allExcept`: (default: `[]`) an array of conditions that will exempt a line
+ *     - `ignore`: (default: `[]`) an array of conditions that will exempt a line
  *        - `regex`: allows regular expression literals to break the rule
  *        - `comments`: allows comments to break the rule
  *        - `urlComments`: allows comments with long urls to break the rule
@@ -38,7 +38,7 @@
 var assert = require('assert');
 var chalk  = require('chalk');
 var DEPRECATED = '[ ' + chalk.yellow('WARN') + ' ] ${option}.${setting} is deprecated.' +
-                 'Please use the corresponding `allExcept` setting instead';
+                 'Please use the corresponding `ignore` setting instead';
 
 module.exports = function() {};
 
@@ -63,7 +63,7 @@ module.exports.prototype = {
                 this._tabSize += ' ';
             }
 
-            var exceptions = maximumLineLength.allExcept || [];
+            var exceptions = maximumLineLength.ignore || [];
             this._allowRegex = (exceptions.indexOf('regex') !== -1);
             this._allowComments = (exceptions.indexOf('comments') !== -1);
             this._allowUrlComments = (exceptions.indexOf('urlComments') !== -1);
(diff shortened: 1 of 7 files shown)


lib/rules/maximum-line-length.js
```diff
@@ -8,9 +8,11 @@
  *  - `Object`:
  *     - `value`: (required) lines should be at most the number of characters specified
  *     - `tabSize`: (default: `1`) considered the tab character as number of specified spaces
- *     - `allowComments`: (default: `false`) allows comments to break the rule
```
markelog: Could you not remove but deprecate them?

lib/rules/maximum-line-length.js
```diff
@@ -58,9 +63,28 @@ module.exports.prototype = {
...
-            this._allowComments = (maximumLineLength.allowComments === true);
-            this._allowUrlComments = (maximumLineLength.allowUrlComments === true);
+            var exceptions = maximumLineLength.ignore || [];
+            this._allowRegex = (exceptions.indexOf('regex') !== -1);
+            this._allowComments = (exceptions.indexOf('comments') !== -1);
+            this._allowUrlComments = (exceptions.indexOf('urlComments') !== -1);
+            this._allowFunctionSignature = (exceptions.indexOf('functionSignature') !== -1);
+
+            if (maximumLineLength.hasOwnProperty('allowRegex')) {
+                console.log(DEPRECATED.replace('${option}', this.getOptionName())
```
markelog: The thing is, output might not be terminal, but anything user want, this is why we have reporters.

So only mentioning them in the docs as deprecated is fine.

lib/rules/maximum-line-length.js
```diff
@@ -76,6 +100,7 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
+        // console.log('check file: ', file._source);
```
markelog: .

lib/rules/maximum-line-length.js
```diff
@@ -8,9 +8,11 @@
  *  - `Object`:
  *     - `value`: (required) lines should be at most the number of characters specified
  *     - `tabSize`: (default: `1`) considered the tab character as number of specified spaces
- *     - `allowComments`: (default: `false`) allows comments to break the rule
- *     - `allowUrlComments`: (default: `false`) allows comments with long urls to break the rule
- *     - `allowRegex`: (default: `false`) allows regular expression literals to break the rule
+ *     - `ignore`: (default: `[]`) an array of conditions that will exempt a line
```
markelog: Why `ignore`?

markelog:
btw, github doesn't send notification when you push new changes, so it is preferable to leave a comment when you do

lib/rules/maximum-line-length.js
clayreimann: `ignore` because I found the wording `allExcept` unclear and not very descriptive. Reverting because the project as a whole appears to prefer `allExcept`.

clayreimann:
Addressed deprecations in documentation rather than logging errors, also updated some of the test files for presets (since `yandex.js` was the only preset test to address maximum line length).

markelog:
Nicely done! Thank you

PR #1423, opened 2015-06-01
requireArrowFunctions: don't check AssignmentExpression or Property

### LINKED ISSUE #1413: requireArrowFunctions: create more exceptions
Ref original: https://github.com/jscs-dev/node-jscs/issues/1383

In es6 strict mode/babel, Arrow functions refer to the outer level `this`. 

In ES6 modules the top level this is `undefined`.  If you want a function with its own this value, you cannot use an arrow function. (maybe that could even be a rule? can't use arrow function with `this` on top level scope)

Two issues for this (I think)?

- shouldn't error when setting prototype or method (not an issue I suppose if you only use classes)

```js
function MyFunc(a) {
  this.a = a;
}
MyFunc.prototype.methodHere = function(a) {};
MyFunc.method2 = function(a) {}
```

- shouldn't error in object literal (should use method shorthand syntax instead, because otherwise this becomes undefined)
```js
var a = {
  b: 1,
  normalFunction: function() { return this.b; },
  arrowFunc: () => { return this.b; },  // undefined.b;
  methodShorthand() { return this.b; }
};
var a.arrow = () => { return this.b }; // undefined.b;
```

So really this rule would allow you to replace all function expressions that don't have this as well as whenever you are using .bind(this) or var x = this?

https://goo.gl/sOxDk1

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18749407-requirearrowfunctions-create-more-exceptions?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1413 

Current idea is don't check for those.

```js
// should use method shorthand syntax instead, otherwise `this` becomes undefined.
var a = {
  b: 1,
  normalFunction: function() { return this.b; },
  arrowFunc: () => { return this.b; },  // undefined.b;
  methodShorthand() { return this.b; }
};
var a.arrow = () => { return this.b }; // undefined.b;
```

Or we figure out if `this` is being used or some other idea.

### DIFF
--- a/lib/rules/require-arrow-functions.js
+++ b/lib/rules/require-arrow-functions.js
@@ -33,9 +33,16 @@
  * ##### Invalid
  *
  * ```js
+ * // function expression in a callback
  * [1, 2, 3].map(function (x) {
  *     return x * x;
  * });
+ * // function expression in a return statement
+ * function a(x) {
+ *     return function(x) { return x };
+ * };
+ * // function expression in a variable declaration
+ * var a = function(x) { return x };
  * ```
  */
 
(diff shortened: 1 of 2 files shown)


mrjoelkemp:
This is a good catch. +1 for ignoring. Computing valid  scopes sounds like a JSHint feature more than a style choice.

hzoo:
Yeah the best test was to take the rule and run it through some code to find these issues!

PR #1432, opened 2015-06-05
Errors: Ability to suppress a single line

### LINKED ISSUE #1372: Support ignoring of a single line with a trailing comment
Hello!

The main thing that prevents us from being able to use jscs at work is that jscs doesn't have the ability to ignore a single line with a trailing comment, like jshint does:

```javascript
var x = 1
var y = 2 // jshint ignore:line
var z = 3
```

Here the missing semicolon on the first and third line will be reported, but the second will not.

It would be great to have the same feature in jscs.

Thanks,
Louis

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14652726-support-ignoring-of-a-single-line-with-a-trailing-comment?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lpil:
This allows the user to suppress errors on a single by appending a trailing comment in the format `// jscs: ignore requireSemicolons`

Fixes #1372 - issue

Possible edgecase that I've not handled:

```javascript
var a = 1
// jscs:disable requireSemicolons
var b = 2
var c = 3
var d = 4 // jscs:ignore requireSemicolons
var e = 5
var f = 6 // should it be enabled or disabled here?
```

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -79,7 +79,7 @@ JsFile.prototype = {
 
         var comments = this.getComments();
         var blockRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
-        var lineRe = /(jscs\s*:\s*ignore)(.*)/;
+        var lineRe = /(jscs\s*:\s*ignore)(\s+\w.*)/;
 
         comments.forEach(function(comment) {
             var enabled;
@@ -93,16 +93,26 @@ JsFile.prototype = {
                 this._addToDisabledRuleIndex(enabled, blockParsed[3], line);
 
             } else if (lineParsed && lineParsed.index === 0) {
-
-                if (!this.isEnabledRule(lineParsed[2], line)) {
-                    return;
-                }
-
-                this._addToDisabledRuleIndex(false, lineParsed[2], line);
-                this._addToDisabledRuleIndex(true, lineParsed[2], line + 1);
+                this._disableRulesAt(lineParsed[2], line);
             }
 
         }, this);
+    },
+
+    /**
+     * Disables a rules for a single line, not re-enabling any disabled rules
+     *
+     * @private
+     */
+    _disableRulesAt: function(rules, line) {
+        rules = rules.split(/\s*,\s*/);
+        for (var i = 0; i < rules.length; i++) {
+            if (!this.isEnabledRule(rules[i], line)) {
+                continue;
+            }
+            this._addToDisabledRuleIndex(false, rules[i], line);
+            this._addToDisabledRuleIndex(true, rules[i], line + 1);
+        }
     },
 
     /**
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -78,7 +78,10 @@ JsFile.prototype = {
         this._disabledRuleIndex = [];
 
         var comments = this.getComments();
+        // Matches a comment enabling or disabling rules.
         var blockRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
+        // Matches a comment disbling a rule for one line.
+        // A rule name must be given for it to match.
         var lineRe = /(jscs\s*:\s*ignore)(\s+\w.*)/;
 
         comments.forEach(function(comment) {
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -81,8 +81,7 @@ JsFile.prototype = {
         // Matches a comment enabling or disabling rules.
         var blockRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
         // Matches a comment disbling a rule for one line.
-        // A rule name must be given for it to match.
-        var lineRe = /(jscs\s*:\s*ignore)(\s+\w.*)/;
+        var lineRe = /(jscs\s*:\s*ignore)(.*)/;
 
         comments.forEach(function(comment) {
             var enabled;
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -78,18 +78,25 @@ JsFile.prototype = {
         this._disabledRuleIndex = [];
 
         var comments = this.getComments();
-        var commentRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
+        var blockRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
+        var lineRe = /(jscs\s*:\s*ignore)(.*)/;
 
         comments.forEach(function(comment) {
             var enabled;
-            var parsed = commentRe.exec(comment.value.trim());
+            var value = comment.value.trim();
+            var blockParsed = blockRe.exec(value);
+            var lineParsed = lineRe.exec(value);
+            var line = comment.loc.start.line;
 
-            if (!parsed || parsed.index !== 0) {
-                return;
+            if (blockParsed && blockParsed.index === 0) {
+                enabled = blockParsed[2] === 'en';
+                this._addToDisabledRuleIndex(enabled, blockParsed[3], line);
+
+            } else if (lineParsed && lineParsed.index === 0) {
+                this._addToDisabledRuleIndex(false, lineParsed[2], line);
+                this._addToDisabledRuleIndex(true, lineParsed[2], line + 1);
             }
 
-            enabled = parsed[2] === 'en';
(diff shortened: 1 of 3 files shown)


mrjoelkemp:
Awesome. Thanks for contributing! Without reviewing your PR just yet and looking at the edge case, the `jscs:disable` (without a matching `jscs:enable` should disable jscs for the entire file. The `jscs:ignore`, in that case, is redundant. As for the question on `var f = 6`, it should be disabled due to the `jscs:disable` comment.

lpil:
OK. In which case I think I need to know whether the rule or rules in question is enabled or disabled at that point in time.

I could split the list of rules and then check each with `isEnabledRule`, but that doesn't seem quite right somehow. Do you have any suggestions? Otherwise I'll keep playing about and see if I can find something that feels nicer.

lpil:
I've made the trailing ignore comment require a rule always. This is largely due to the behaviour of enabling/disabling all and specific rules in the same area being poorly defined.

It's also nice as it discourages the user from being heavy handed with ignoring rules.

What are your thoughts?

test/specs/js-file.js
```diff
@@ -162,6 +162,61 @@ describe('modules/js-file', function() {
...
+                assert(file.isEnabledRule('validateQuoteMarks', 1));
+                assert(!file.isEnabledRule('validateQuoteMarks', 2));
+                assert(!file.isEnabledRule('validateQuoteMarks', 3));
+                assert(!file.isEnabledRule('validateQuoteMarks', 4));
+            });
+
+            it('should also work with multiple rules', function() {
+                var file = createJsFile([
+                    '// jscs: disable validateQuoteMarks',
+                    'var a = "1"; // jscs: ignore validateQuoteMarks, anotherRule',
```
markelog: Just in case, could add `//jscs: ` test?

lib/js-file.js
```diff
@@ -102,6 +124,8 @@ JsFile.prototype = {
      */
     isEnabledRule: function(ruleName, line) {
         var enabled = true;
+        ruleName = ruleName.trim();
```
markelog: Why?

lib/js-file.js
```diff
@@ -78,22 +78,44 @@ JsFile.prototype = {
         this._disabledRuleIndex = [];
 
         var comments = this.getComments();
-        var commentRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
+        var blockRe = /(jscs\s*:\s*(en|dis)able)(.*)/;
+        var lineRe = /(jscs\s*:\s*ignore)(\s+\w.*)/;
```
markelog: Could add comments for those variables?

test/specs/js-file.js
lpil: Sorry, I'm not sure what you mean?

lib/js-file.js
lpil: Of course. What specifically would you like me to document here?

lib/js-file.js
lpil: (Off the top of my head) This was because something was returning rule names with surrounding whitespace, and I think that rule names should never have whitespace.

I do not know if this is actually required for correct behaviour currently, it could have been a hang over from a previous state. I'll check when I next have access to a computer, which is probably Monday.

test/specs/js-file.js
hzoo: I think @markelog means `//jscs: ` in that it doesn't have a space after the `//` like in `// jscs: ignore validateQuoteMarks`

lib/js-file.js
markelog: If you could add a test for it, it would be great

test/specs/js-file.js
markelog: Yeah, exactly

lib/js-file.js
markelog: Just why those variables are needed

lpil: Well they're not really- they could just be literals within the loop. I don't know why the original author decided to make it a variable.

Would you prefer them not to be variables?

test/specs/js-file.js
lpil: Certainly. :)

lib/js-file.js
lpil: Sure thing.

markelog:
Also, the most important thing :-), could add a note to the docs about this feature?

lib/js-file.js
markelog: In some guides it is preferable to move them in special variables with prefixes, it seems author followed that guide.

> Would you prefer them not to be variables?

It is fine with me, just a comment would suffice

lpil:
There you go.

I wasn't entirely sure what to say about the regexes, let me know what you think. :)

test/specs/js-file.js
```diff
@@ -162,6 +162,94 @@ describe('modules/js-file', function() {
...
+                assert(!file.isEnabledRule('validateQuoteMarks', 3));
+                assert(file.isEnabledRule('anotherRule', 1));
+                assert(!file.isEnabledRule('anotherRule', 2));
+                assert(file.isEnabledRule('anotherRule', 3));
+            });
+
+            it('requires a rule to be specified', function() {
+                var file = createJsFile([
+                    '// jscs: disable validateQuoteMarks',
+                    'var a = "1"; // jscs: ignore', // invalid
```
markelog: Could that be valid? To ignore single line for all rules?

lpil: Not easily with how disabling all rules is currently represented in the disabled rule index. There's no good way to know which rules to reenable after the line.

The existing comments's behaviour is also a little funny here... I think this would be a good place for refactoring- having the 'all' comments actually add an entry for every rule rather than this special `'*'` rule.

markelog: Hm, not sure if i get it, changing from `/(jscs\s*:\s*ignore)(\s+\w.*)/` to `/(jscs\s*:\s*ignore)(.*)/` and redoing test to 

```js
it('requires a rule to be specified', function() {
    var file = createJsFile([
        'var a = "1"; // jscs:ignore',
        'var b = "1";'
    ].join('\n'));
    assert(!file.isEnabledRule('validateQuoteMarks', 1));
    assert(file.isEnabledRule('validateQuoteMarks', 2));
});
```

Works for me

lpil: Ah, that's not quite the problem I'm afraid!

It's easy to make a regex that will permit comments without any rule names, and this `isEnabledRule` based test will pass in that situation, the problem is with toggling on and off all rules *when* there is already specific rules enabled. This is due to there being one special rule that represents all other rules, `'*'`.

If we have 1 rule turned off, and then want to toggle on and off all rules, how do we do this? When we toggle back on all rules after the single line trailing comment we will override the previously turned off rule as there's no good way to identify it was already on, and that it needs to be disabled again on the same line we enable all rules.

This same problem happens with the existing block comments, which I think might qualify as a bug?

Either way, the intended behaviour is slightly unclear, and implementing this feature will take a sizeable refactoring of the disabled rule index system, which my colleagues and I did not feel comfortable undertaking.

markelog: Yeah, that makes sense, it seems as edge case though, we have never saw someone having issues with it.

Could you then change regexp and create issue for described problem?

lpil: To be clear, do you want me to allow disabling all rules for a single line with a trailing comment?

markelog: If there is no objections, then yeah.

lpil: You're the boss. I'll do this tomorrow. :)

lpil: Done. How does that look?

markelog:
For some reason coverage decreased, but changed `js-file` is still 100%, so i landed it regardless. 
Anyway, awesome! Thank you :-)

lpil:
Thanks! :D :dancer:

Issue #1214, opened 2015-03-28
Removal of `JsFile.prototype.getTokenPosByRangeStart`


rwjblue:
`JsFile.prototype.getTokenPosByRangeStart` was removed in https://github.com/jscs-dev/node-jscs/pull/1076. Since `JsFile` is exposed as public API and this function was not marked as `@private` (or underscored), I would have expected it to be available (but possibly deprecated) throughout the 1.x timeframe.  I also did not see this removal mentioned in the [release notes](https://github.com/jscs-dev/node-jscs/releases/tag/v1.12.0).

We were using this function in a custom rule in Ember ([here](https://github.com/emberjs/ember.js/blob/master/lib/jscs-rules/require-spaces-after-closing-parenthesis-in-function-declaration.js)), and an issue was reported in https://github.com/emberjs/ember.js/issues/10743.

I have updated the rule in Ember to allow builds to work properly (in https://github.com/emberjs/ember.js/pull/10747), but I am uncertain what y'alls normal process is here for this type of stuff so I figured I'd open an issue to let y'all know.

rwjblue:
Also, the newer `getFirstNodeToken` method is **much** better/easier to use,  thank you for adding it.

hzoo:
There was an issue about this - I think https://github.com/jscs-dev/node-jscs/issues/1084? Sorry about that

mrjoelkemp:
Thanks @hzoo. Sorry about that oversight @rwjblue; we removed the function since it wasn't being used internally. We should have deprecated it and removed it in 2.0. Thanks for reporting the issue; it's great to get some certainty that folks are using the public apis.

@mikesherov Should we consider reverting the commit that removes the JSFile method and including it in a patch release?

markelog:
I would even move those methods in different repo as its own module. 

Since this no longer fictional, i would re-added, who knows when 2.0 will came up.

mikesherov:
Not interested in readding the removed functions. We have never documented them. Yes, it's unfortunate that people had relied on them but we'd need to document their existence before considering them part of the API.

mrjoelkemp:
> it's unfortunate that people had relied on them

Doesn't that potentially affect all users of `additionalRules`? Not that everyone used this particular function, but they use JSFile's methods.

This feels like a delicate area to be so firm. By this logic, we could merge the rule of silence PR since we never documented that the error/success messages couldn't change – well knowing 3rd party integrations probably rely on that output. 

[...]

PR #1433, opened 2015-06-06
Improve JsFile constructor for better encapsulation

mdevils:
* Removed `JsFile.parse`.
* Moved parse routines to `JsFile` constructor.
* Added `JsFile:: getParseErrors` method.
* Fixed tests.
* Added tests.

I need this to implement abstraction on top of different file formats and then implement processing `JS` inlined to `HTML` pages.

/cc @markelog @zxqfox @mikesherov @mrjoelkemp

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -12,24 +12,37 @@ var KEYWORD_OPERATORS = {
  * File representation for JSCS.
  *
  * @name JsFile
+ * @param {Object} params
+ * @param {String} params.filename
+ * @param {String} params.source
+ * @param {Object} params.esprima
+ * @param {Object} [params.esprimaOptions]
+ * @param {Boolean} [params.es3]
+ * @param {Boolean} [params.es6]
  */
-var JsFile = function(filename, source, tree, options) {
-    options = options || {};
+var JsFile = function(params) {
+    params = params || {};
+    this._parseErrors = [];
+    this._filename = params.filename;
+    this._source = params.source;
+    this._tree = {tokens: [], comments: []};
 
-    this._filename = filename;
-    this._source = source;
-    this._tree = tree || {tokens: [], comments: []};
-
-    this._es3 = options.es3 || false;
-    this._es6 = options.es6 || false;
+    this._es3 = params.es3 || false;
+    this._es6 = params.es6 || false;
 
     this._lineBreaks = null;
-    this._lines = source.split(/\r\n|\r|\n/);
+    this._lines = this._source.split(/\r\n|\r|\n/);
+
+    try {
+        this._tree = parseJavaScriptSource(this._source, params.esprima, params.esprimaOptions);
+    } catch (e) {
+        this._parseErrors.push(e);
+    }
 
     this._tree.tokens = this._fixEs6Tokens(this._tree.tokens);
     this._tokens = this._buildTokenList(this._tree.tokens, this._tree.comments);
     this._addEOFToken();
-    this._applyWhitespaceData(this._tokens, source);
+    this._applyWhitespaceData(this._tokens, this._source);
 
     var tokenIndexes = this._buildTokenIndex(this._tokens);
     this._tokenRangeStartIndex = tokenIndexes.tokenRangeStartIndex;
@@ -651,6 +664,15 @@ JsFile.prototype = {
     },
 
     /**
+     * Returns list of parse errors.
+     *
+     * @returns {Error[]}
+     */
+    getParseErrors: function() {
+        return this._parseErrors;
+    },
+
+    /**
      * Builds token list using both code tokens and comment-tokens.
      *
      * @returns {Object[]}
@@ -805,7 +827,7 @@ JsFile.prototype = {
  * @param {Object} [esprimaOptions]
  * @returns {Object}
  */
-JsFile.parse = function(source, esprima, esprimaOptions) {
+function parseJavaScriptSource(source, esprima, esprimaOptions) {
     var finalEsprimaOptions = {
         tolerant: true
     };
@@ -859,6 +881,6 @@ JsFile.parse = function(source, esprima, esprimaOptions) {
     }
 
     return tree;
-};
+}
 
 module.exports = JsFile;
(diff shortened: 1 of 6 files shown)


lib/js-file.js
```diff
@@ -805,7 +827,7 @@ JsFile.prototype = {
  * @param {Object} [esprimaOptions]
  * @returns {Object}
  */
-JsFile.parse = function(source, esprima, esprimaOptions) {
+function parseJavaScriptSource(source, esprima, esprimaOptions) {
```
mrjoelkemp: This is considered a breaking change no?

mdevils: Nope, it's not part of public API.

mrjoelkemp: See https://github.com/jscs-dev/node-jscs/issues/1214. It's not private because it's not closed over, not using `@private`, and is not called `_parse`.

mdevils: It's a static method of `JsFile`, plugins cannot access it (without hacks).

mrjoelkemp: Great point. Missed that. :+1:

lib/js-file.js
```diff
@@ -12,24 +12,37 @@ var KEYWORD_OPERATORS = {
...
+ * @param {Object} options
+ * @param {String} options.filename
+ * @param {String} options.source
+ * @param {Object} options.esprima
+ * @param {Object} [options.esprimaOptions]
+ * @param {Boolean} [options.es3]
+ * @param {Boolean} [options.es6]
  */
-var JsFile = function(filename, source, tree, options) {
+var JsFile = function(options) {
```
gibson042: Forgive the intrusion, but `source` is definitely not optional and should probably remain an independent parameter rather than entering the `options` object.

mdevils: Most of the params are not optional. I introduces object because signature became to complicated. Probably the problem is the name: `options`. I can rename it to `params`.

markelog:
I'd say this PR should point to different target, i think lintining js in html code should be a 2.0 issue.

mdevils:
> I'd say this PR should point to different target, i think lintining js in html code should be a 2.0 issue.

Did not understand you. What do you mean by different target? What extracting from HTML has to do with 2.0?

markelog:
> What do you mean by different target?

Not `master` but `2.0`

> What extracting from HTML has to do with 2.0?

Nothing, but this is a structural change, which would mean we would have problems merging it to 2.0, while if we land this in 2.0 right away, we wouldn't have these issues.

So i'd propose landing only iterative changes for the 1.x, like new rules, bugfixes, etc. i.e. changes that wouldn't affect merging process with 2.0 and allow us to release 1.x if needs to, everything else -  2.x

lib/string-checker.js
```diff
@@ -225,28 +220,34 @@ StringChecker.prototype = {
...
-            this._addParseError(parseErrors, parseError);
-            return {output: source, errors: parseErrors};
+        var file = createFileInstance(source);
+        var errors = new Errors(file, this._verbose);
+
+        var parseErrors = file.getParseErrors();
+
+        if (parseErrors.length > 0) {
+            parseErrors.forEach(function(parseError) {
+                this._addParseError(errors, parseError);
```
mrjoelkemp: No need to check for whether or not the max errors have been exceeded?

lib/string-checker.js
```diff
@@ -93,27 +93,22 @@ StringChecker.prototype = {
...
-        var sourceTree;
-        var parseError;
-
-        try {
-            sourceTree = JsFile.parse(source, this._esprima, this._configuration.getEsprimaOptions());
-        } catch (e) {
-            parseError = e;
-        }
-
-        var file = this._createJsFileInstance(filename, source, sourceTree);
```
mrjoelkemp: Are you trying to remove all usage of this? Why wouldn't you just update it to use the new object format and then use it within this function to generate `file` (as it used to) and remove your local duplicate: https://github.com/jscs-dev/node-jscs/pull/1433/files?diff=unified#diff-c4ec64ef009f51578bddc596b5b6ddc2R225?

A little confused here.

test/specs/js-file.js
```diff
@@ -8,27 +8,43 @@ var fs = require('fs');
...
-            JsFile.parse(sources, esprima),
-            options || {}
-        );
+        var params = {
+            filename: 'example.js',
+            source: sources,
+            esprima: esprima
+        };
+
+        for (var i in options) {
```
mrjoelkemp: We already have lodash.assign (https://github.com/jscs-dev/node-jscs/blob/mdevils.jsfile-constructor/package.json#L71), maybe use it here?

lib/js-file.js
```diff
@@ -12,24 +12,37 @@ var KEYWORD_OPERATORS = {
...
 
     this._es3 = options.es3 || false;
     this._es6 = options.es6 || false;
 
     this._lineBreaks = null;
-    this._lines = source.split(/\r\n|\r|\n/);
+    this._lines = this._source.split(/\r\n|\r|\n/);
+
+    try {
+        this._tree = parseJavaScriptSource(this._source, options.esprima, options.esprimaOptions);
```
mrjoelkemp: :+1: for bringing this into JSFIle

mrjoelkemp:
Looks good aside from the confusion around what's happening in `lib/string-checker` and the failed test. Nice work here.

lib/string-checker.js
mdevils: In fixing we do not support max errors.

lib/string-checker.js
mdevils: Yeah, thank you for noticing. I will figure this out.

test/specs/js-file.js
mdevils: Good point, thanks!

mdevils:
@markelog Undestood. OK, I will merge this to `2.0` after review.

mdevils:
@mrjoelkemp Thank you for your review!

lib/string-checker.js
mdevils: Fixed this: reused `_createJsFileInstance`.

mrjoelkemp:
This particular PR is a harmless refactor. I'd say merge to 2.0 then cherrypick to master. No reason it shouldn't get released. 

If anything, a community PR that modifies JSFile (like grit processing) would need these changes or 2.0 will have a bad time with the merge conflict (if any).

mdevils:
OK, I will port it to master.

PR #1435, opened 2015-06-06
Refactor integration tests

mdevils:
After integration tests silently failed, I decided to figure out why, and during that process, I made some refactoring:

* More human-readable integration tests code.
* Use promises instead of callbacks: bulletproof execution flow and readability.
* Show test output (currently output is hidden, it hides errors).
* Fix STDIO-error (the actual cause of the problem).

/cc @markelog

### DIFF
--- a/test/scripts/integration.js
+++ b/test/scripts/integration.js
@@ -3,7 +3,6 @@ var chalk = require('chalk');
 var vow = require('vow');
 var vowFs = require('vow-fs');
 
-var Promise = vow.Promise;
 var SOURCES = ['lib', 'test/specs'];
 var MOCHA = 'node_modules/.bin/mocha';
 
@@ -36,8 +35,8 @@ vowFs.listDir('./presets').then(function(presetFilenames) {
         );
     });
 })
-.progress(function() {})
-.done();
+    .progress(function() {})
+    .done();
 
 /**
  * Creates log error handler to output error message and pass error further.
@@ -142,7 +141,6 @@ function runProcess(executable, args, successCodes) {
  * Exit handler. Restores file changes before exit.
  *
  * @param {Error} error
- * @returns {Promise}
  */
 function exit(error) {
     var errorCode = 0;
--- a/test/scripts/integration.js
+++ b/test/scripts/integration.js
@@ -8,7 +8,7 @@ var MOCHA = 'node_modules/.bin/mocha';
 
 /**
  * Applying every available preset to JSCS sources and tests, then executing tests.
- * It is
+ * So we can make sure nothing breaks during these reformatting actions.
  */
 vowFs.listDir('./presets').then(function(presetFilenames) {
     var presets = presetFilenames.map(function(presetFilename) {
--- a/test/scripts/integration.js
+++ b/test/scripts/integration.js
@@ -10,31 +10,32 @@ var MOCHA = 'node_modules/.bin/mocha';
  * Applying every available preset to JSCS sources and tests, then executing tests.
  * So we can make sure nothing breaks during these reformatting actions.
  */
-vowFs.listDir('./presets').then(function(presetFilenames) {
-    var presets = presetFilenames.map(function(presetFilename) {
-        return presetFilename.replace('.json', '');
-    });
-    console.log('\n' + chalk.green('> ') + 'Autofix ingeration tests');
-    return promiseQueue(presets, function(presetName) {
-        console.log('\nPreset "' + chalk.green(presetName) + '"');
-        logStep('Autofix execution');
-        return applyPreset(presetName).then(
-            function() {
-                logStep('Unit tests');
-                console.log('');
-                return runTests().then(
-                    function() {
-                        console.log('');
-                        logStep('Unit test on "' + chalk.green(presetName) + '" preset are finished');
-                    },
-                    logErrorHandler('Unit tests failure'),
-                    passthroughProgress
-                );
-            },
-            logErrorHandler('Autofix failure')
-        );
-    });
-})
+vowFs.listDir('./presets')
+    .then(function(presetFilenames) {
+        var presets = presetFilenames.map(function(presetFilename) {
+            return presetFilename.replace('.json', '');
+        });
+        console.log('\n' + chalk.green('> ') + 'Autofix ingeration tests');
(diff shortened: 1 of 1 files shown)


markelog:
> Show test output (currently output is hidden, it hides errors).
> Fix STDIO-error (the actual cause of the problem).

Those are fixed in 2.0 branch :)

> Use promises instead of callbacks: bulletproof execution flow and readability.

Well, `promise().then(fn)` is basically the same as `read(..., fn)` if there was something like `.then().then()` i would all for it, otherwise i don't see the point. Besides there is still code that uses callbacks which makes the code heterogeneous, i.e. uses two different concepts.

mdevils:
@markelog Just see the code. I could not understand anything in the previous version — it was not properly abstracted — process routines messed up with control flow. With all respect, I was totally confused when I was trying to fix my build.

With this code you have structure and abstraction.

mdevils:
> Those are fixed in 2.0 branch :)

Why not port this to 1.0 and fix PRs? Where can I read about 2.0 branch?

markelog:
> @markelog Just check out the code. I could not understand anything in the previous version — it was not properly abstracted — process routines messed up with control flow. With all respect, I was totally confused when I was trying to fix my build.

> With this code you have structure and abstraction.

This is subjective, matter of taste even. And that is okay, we all see code that we don't like and have different opinions on. 

For example i don't see how this code is better - additinal abstraction like promise 
don't give me anything, except as what it is - an abstraction, which could be considered as needless 
abstraction altogether since there is no advantages to use promises, at least, i don't see them, since promises usually dealing with different kind of code flow.

But `vowfs` and `vow` are not "popular" packages, without any real community, API that you have to know, know common issues too and bunch of other things.

Those might be obvious to you, but for anyone else this is a blackbox. Whereas i operated with standart API.

Yes, you introduced couple functions and couple constants, but that could be as seen as what i described above -  additional needless abstraction, like `passthroughProgress` helper, it just proxies to `stdout` does nothing else, why it need to be there? Because you might switch to `console.log` there? Or because it just feels right? Or both? You might see it as "proper abstraction" someone else could see this is "needless abstraction".

I mean phrase like 
> it was not properly abstracted — process routines messed up with control flow

is subjective and not productive, i can say the same thing about these changes, and mine phrase would mean nothing to you, just like yours statement means nothing to me. If you say something like that - prove it, or better yet, just lay the evidences, not the conclusion.

I'm not gonna call your code as "bad" or something else negative, no, but i'd propose you would give me the same courtesy and operate only on objective terms - expandability, cyclomatic complexity, modularity, etc.

> Why not port this to 1.0 and fix PRs? Where can I read about 2.0 branch?

There is only roadmap. I'd propose landing only iterative changes for the 1.x, like new rules, bugfixes, etc. i.e. changes that wouldn't affect merging process with 2.0, everything else - 2.x that is my view though, didn't yet shared it with anyone else.

mdevils:
> is subjective and not productive, i can say the same thing about these changes, and mine phrase would mean nothing to you, just like yours statement means nothing to me. If you say something like that - prove it, or better yet, just lay the evidences, not the conclusion.

How is it subjective?

Look at this code:
```js
        // Execute jscs -x on jscs source code for the given presets
        var child = spawn('node', args, { timeout: Infinity });

        process.stdout.write(messages.start + '\n\n');
        process.stdout.write(chalk.magenta(' - ') + messages.fixStart + '\n');

        // For some reason this makes subprocess to be more effective
        child.stdout.on('data', function() {});

        // Wait until autofix process is done
        child.on('close', function executeTests(code) {
```

You have low-level routines messed (controlling child processes) up with control routines and user interface (console).

And look at this:

```js
        console.log('\nPreset "' + chalk.green(presetName) + '"');
        logStep('Autofix execution');
        return applyPreset(presetName).then(
            function() {
                logStep('Unit tests');
                console.log('');
                return runTests().then(
                    function() {
                        console.log('');
                        logStep('Unit test on "' + chalk.green(presetName) + '" preset are finished');
                    },
                    logErrorHandler('Unit tests failure'),
                    passthroughProgress
                );
            },
            logErrorHandler('Autofix failure')
        );
```

As you can see, this is pure high-level part: only control flow and user interface. All low-level code is encapsulated into functions with names which are abstracted over implementation.

So I conclude that this is not the matter of tastes.

markelog:
> You have low-level routines (controlling child processes) messed up with control routines and user interface (console).

I don't see who they "messed up" with it, but i see dramatic increase of cyclomatic complexity, which is has direct correlation with code perception.

Encapsulation by itself is not always great thing, if you put something in function it doesn't mean it is became encapluslated. "high-level" code is not always great too, especially, with lack of definition of it and context dependency. 

Also this kind be interpreted as half way still, since constructions like this -

```js
console.log('');
logStep(...);
```

or `console.log('\n' + chalk.green('> ') + 'Autofix ingeration tests');`, while in `logStep` you "encapsulated" `console.log(chalk.magenta(' - ') + step);`.

Do not strike me as properly encapsulated.

Those things are needed only if you need them, while you increased the complexity, level of encapsulation, added new interfaces, new APIs, for what?

You had the problem - you didn't see error output of failed tests, in 2.0 branch you can now do that. Whereas here, you don't introduce new features or improvements, you exercising your level of "rightness", you refactor because of the refactoring, uroborus of some sort.

This discussion could go on for the long time, we will start to define what is a "low-level" and why cyclomatic complexity not the right metric and so on. 

That would be controproductive. You want to make this "right", fine, i'm not oppose to it, i don't see the point, i don't see who this would be better or why you paradigms are better then mine. But fine, this is your call anyhow.

There is only one admonition from me, i would prefer if you don't use promises.

mdevils:
I don't understand what are we discussing here. In the initial version you cannot understand what is going on reading the code. We had several process instances, console writes all over around. In the new version you can easily understand the flow just reading the code from top to bottom.

Again, it took me 1 hour to debug this file. Mainly because I could not understand the flow and what's going on. That's the only reason I rewrote this file.

> There is only one admonition from me - i would prefer if you don't use promises.

I don't care about "promises vs callbacks". We already use promises in JSCS, so consistency is more important than preferences.

mdevils:
We finished out discussion in IM.

Issue #169, opened 2014-01-14
disallowEmptyBlocks: Don't warn `Empty block found` for empty functions


lahmatiy:
Empty functions may be used in various cases, that's not a mistake.
Valid examples:

``` js
var noop = function(){};

function doSomeStuff(){
  // TODO: implement code in future
}
```

ptarjan:
What ever happened to this? I'd like to to allow comments in if blocks

```
if (foo) {
  // Clearly foo is ok
} else {
  ...
}
```

hzoo:
How about you make another issue since it's for blocks with only comments rather than functions

ptarjan:
@hzoo https://github.com/jscs-dev/node-jscs/issues/1767

PR #1442, opened 2015-06-08
disallowEmptyBlocks: should not report empty arrow blocks

jakezatecky:
Similar to #169 and #174, this just extends the logic to arrow functions.

### DIFF
--- a/lib/rules/disallow-empty-blocks.js
+++ b/lib/rules/disallow-empty-blocks.js
@@ -49,7 +49,8 @@ module.exports.prototype = {
             if (node.body.length === 0 &&
                 node.parentNode.type !== 'CatchClause' &&
                 node.parentNode.type !== 'FunctionDeclaration' &&
-                node.parentNode.type !== 'FunctionExpression') {
+                node.parentNode.type !== 'FunctionExpression' &&
+                node.parentNode.type !== 'ArrowFunctionExpression') {
                 errors.add('Empty block found', node.loc.end);
             }
         });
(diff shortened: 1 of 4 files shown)


hzoo:
Looks good.

Also do we want to make a helper for functions (`FunctionDeclaration`, `FunctionExpression`, `ArrowFunctionExpression`)? Maybe some other rules need to be updated with ArrowFunctionExpression

jakezatecky:
It probably makes sense to add a helper, but I'll leave that for the more experienced. It is also probably the case that other rules also need to be updated with `ArrowFunctionExpression` blocks (such as `validateAlignedFunctionParameters`).

hzoo:
@jakezatecky If it's not hard to add we might as well include it here too (the other rules), otherwise we can just do another PR later (if you want to, or I'l do it).

jakezatecky:
I certainly can go and add the arrow function-related modifications to other rules. I should be able to get some time closer toward the end of this week.

zxqfox:
@jakezatecky Thanks! Good luck with that ;-)

markelog:
@jakezatecky still interested?

jakezatecky:
Sorry for the lack of progress. Yes, I'm still interested and will (soon) work on updating the other rules to account for arrow functions.

markelog:
No worries :-)

markelog:
Oh, i see you added a commit, but coverage decreased, would you mind adding more tests and rebase?

markelog:
Hm, didn't find a problem with the coverage, anyway, thank you!

PR #1445, opened 2015-06-10
allExcept: inCallExpression should also include object instantiation

matoilic:
Currently when setting the inCallExpression exception for requirePaddingNewLinesAfterBlocks object instantiation is not included in that exception. In the end it's the same scenario. So, when setting the inCallExpression exception the following should also be valid.

```javascript
//should be valid
new MyClass(
    'param1',
    function() {
        //some code
    }
);
```

```javascript
//is already valid
myFunction(
    'param1',
    function() {
        //some code
    }
);
```

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -7,7 +7,8 @@
  * - `true`: always require a newline after blocks
  * - `Object`:
  *      - `"allExcept"`: `Array`
- *          - `"inCallExpressions"` Blocks don't need a line of padding in argument lists
+ *          - `"inCallExpressions"` Blocks don't need a line of padding in function argument lists
+ *          - `"inNewsExpressions"` Blocks don't need a line of padding in constructor argument lists
  *          - `"inArrayExpressions"` Blocks don't need a line of padding in arrays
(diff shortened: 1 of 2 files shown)


hzoo:
@matoilic Do you want to do the fix as well? Also `new MyClass` is a `NewExpression` not a `CallExpression` so we would need to add another option (I guess `inNewExpression`) for that.

matoilic:
I've added a new option `inNewExpressions` as suggested and adapted the documentation.

lib/rules/require-padding-newlines-after-blocks.js
```diff
@@ -7,7 +7,8 @@
  * - `true`: always require a newline after blocks
  * - `Object`:
  *      - `"allExcept"`: `Array`
- *          - `"inCallExpressions"` Blocks don't need a line of padding in argument lists
+ *          - `"inCallExpressions"` Blocks don't need a line of padding in function argument lists
+ *          - `"inNewsExpressions"` Blocks don't need a line of padding in constructor argument lists
```
zxqfox: New vs News

zxqfox:
LGTM (except one nit in doc)

markelog:
Thank you! It was easy isn't it :)

PR #1446, opened 2015-06-10
(WIP) Misc: use babel-jscs for the esnext option

### LINKED ISSUE #1353: Supporting ES7/Babel
For #1352, #1344, #1270.

My plan is to make `babel` a possible parser to use in the `esprima` option.
I could do a repo for this so you can do `"esprima": "babel-jscs"` or something like that.
Nothing would need to be changed in `jscs` itself.
`babel-jscs` would modify the acorn parser to output into an esprima format and patch `estraverse` (just modifying `babel-eslint` somewhat).

I got it working on an es7 code snippet (the issues linked) and a small project but running it with more rules leads to some errors that still need to be fixed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13724331-supporting-es7-babel?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1353 @markelog 

Don't know if we just want to replace the `esprima-harmony-jscs`?

Also since babel-jscs has babel-core as a dependency, adding babel-jscs would make jscs bigger. It's not a problem for people using babel already - just unneeded for everyone else?

---

I think the tests are failing because of the local/global npm issue? Not sure why.

All the tests with `esprima: 'babel-jscs',` like

```js
it('should use a custom esprima provided at CLI', function() {
            return assertNoCliErrors(cli({
                args: ['test/data/cli/esnext.js'],
                esprima: 'babel-jscs', // was fine with 'esprima-harmony-jscs'
                config: 'test/data/cli/cli.json'
            }));
        });
```

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -109,7 +109,7 @@ jscs path[ path[...]] --reporter=./some-dir/my-reporter.js
 ```
 
 ### `--esnext` (`-e`)
-Attempts to parse your code as ES6 using the harmony version of the esprima parser. Please note that this is currently experimental, and will improve over time.
+Attempts to parse your code as ES6+ using the babel-jscs package as the parser.Please note that this is currently experimental, and will improve over time.
 
 ### `--esprima` (`-s`)
 Attempts to parse your code with a custom Esprima version.
@@ -236,7 +236,7 @@ Default: Infinity
 
 ### esnext
 
-Attempts to parse your code as ES6 using the harmony version of the esprima parser.
+Attempts to parse your code as ES6+ using the babel-jscs package as the parser.Please note that this is currently experimental, and will improve over time.
 
 Type: `Boolean`
 
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -109,7 +109,7 @@ jscs path[ path[...]] --reporter=./some-dir/my-reporter.js
 ```
 
 ### `--esnext` (`-e`)
-Attempts to parse your code as ES6+ using the babel-jscs package as the parser.Please note that this is currently experimental, and will improve over time.
+Attempts to parse your code as ES6+, JSX, and Flow using the babel-jscs package as the parser. Please note that this is experimental, and will improve over time.
 
 ### `--esprima` (`-s`)
 Attempts to parse your code with a custom Esprima version.
@@ -236,7 +236,7 @@ Default: Infinity
 
 ### esnext
 
-Attempts to parse your code as ES6+ using the babel-jscs package as the parser.Please note that this is currently experimental, and will improve over time.
+Attempts to parse your code as ES6+, JSX, and Flow using the babel-jscs package as the parser. Please note that this is experimental, and will improve over time.
 
 Type: `Boolean`
 
--- a/lib/string-checker.js
+++ b/lib/string-checker.js
@@ -1,5 +1,5 @@
 var defaultEsprima = require('esprima');
-var harmonyEsprima = require('esprima-harmony-jscs');
+var babelJscs = require('babel-jscs');
 var Errors = require('./errors');
 var JsFile = require('./js-file');
 var Configuration = require('./config/configuration');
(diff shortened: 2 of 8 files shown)


OVERVIEW.md
```diff
@@ -236,7 +236,7 @@ Default: Infinity
 
 ### esnext
 
-Attempts to parse your code as ES6 using the harmony version of the esprima parser.
+Attempts to parse your code as ES6+ using the babel-jscs package as the parser.Please note that this is currently experimental, and will improve over time.
```
markelog: "parser.Please not" -> "parser. Please not"

markelog: Also, we should add note about "jsx"

markelog:
> I think the tests are failing because of the local/global npm issue?

Nope, legitimate failure, "couldn't resolve jscs"

> make jscs bigger

By how much?

OVERVIEW.md
zxqfox: `JSX` (uppercase?)
Not sure currently is needed.
"will improve" → "will be improved"?

markelog: "uppercase" sure

hzoo:
Yeah I think I need to modify the whole monkeypatch module thing (replace `estraverse.VisitorKeys`) to work correctly. `jscsLoc = Module._resolveFilename("jscs", module.parent.parent);`

@markelog Not sure - if I look at `babel-jscs/node_modules/babel-core` it's 18.8 mb?

markelog:
It seems you solved all open issues in `babel-jscs` repo, can we land this?

hzoo:
Yeah - I can't figure out why it can't find `jscs` or `babel-jscs` to make the tests fail path though - some path issue

markelog:
Okay, i finish this up

hzoo:
I wasn't able to fix the tests (even though it works) - and it looks like it's still failing (on master) @markelog?

hzoo:
Also do we want to keep the option as `esnext` or use a different name?

mrjoelkemp:
What's the benefit of changing the name?

hzoo:
Was just thinking about how `esnext` isn't really a thing anymore but it's not a big deal or anything

markelog:
> Was just thinking about how esnext isn't really a thing anymore

Well, Esprima still doesn't support all es6 features and es7 features and jsx could be considered as "esnext", doesn't you think?

mikesherov:
esnext was a term used to describe es6 back in the "harmony" days. It is no longer a useful description of any language feature. If anything, we should either switch to a `version` flag (considering new version every year), or have a `stage` flag which allows the user to opt into features at different stages of acceptance (this is what babel does).

mrjoelkemp:
The entire flag, the added code and tests for esnext all seem unnecessary.
We use esprima and if it lags behind what others need out of a parser, then
they should just specify a parser to use with custom options for that
parser. If we want to avoid them having to do that then we should adopt the
bleeding edge parser instead of esprima.

During the push to stabilize babel-jscs, I always wondered why we don't
just use that as the primary parser instead of Esprima.

2 options in my mind (both involve dropping esnext):

1. Adopt babel-jscs as the primary parser
2. Use esprima and encourage the use of babel-jscs as an alternative

Thoughts?

hzoo:
Ok I'd go with 2 then for now. Right now babel-jscs is slow because of some of the transforms needed (not that the babel parsing itself is slow but the added stuff makes it slower)

mrjoelkemp:
@hzoo How much slower are we talking?

hzoo:
It's just slow enough people will ask about it (with babel-eslint which is basically the same). It's only for people who want to lint on keystroke/save very often. It's not that big of a problem or else we would of gotten a lot of issues for it already. https://github.com/mdevils/cst/pull/8#issuecomment-116091494

PR #1453, opened 2015-06-13
Update package.json

paladox:
Updating per https://david-dm.org/jscs-dev/node-jscs#info=devDependencies and https://david-dm.org/jscs-dev/node-jscs


mrjoelkemp:
Thanks for contributing. The upgrade combination of esprima and estraverse is breaking the test suite. Since we don't yet have integration tests for the browser bundle, I'd rather not bump a major version of browserify (seeing what the breaking change is through their commit history is hard). 

Updates to lodash.assign and jshint can go in though.

Issue #1447, opened 2015-06-11
New rule for , and ; to not have preceding space?


ShashankaNataraj:

markelog:
Would be very interesting rule, would you like to tackle this one?

markelog:
You also already can do that for the comma btw - http://jscs.info/rule/disallowSpaceAfterBinaryOperators.html

But we would rather move it to it own rule(s), see discussion in https://github.com/jscs-dev/node-jscs/issues/845#issuecomment-111121060

hzoo:
`dockyard/ember-suave` has the semicolon one done. https://github.com/dockyard/ember-suave/blob/master/lib/rules/disallow-space-before-semicolon.js if we want to use that. And regarding commas I think there was an issue about that

markelog:
Cool, if only it would use `assert` module and no regexps

markelog:
Okay, we landed semicolon rule, should be pretty easy to create rule for the comma

markelog:
@ShashankaNataraj would you like to tackle this?

ShashankaNataraj:
@markelog SUre would! Will raise a PR.

markelog:
@ShashankaNataraj it seems you almost finish with it?

ShashankaNataraj:
@markelog yeah, but I'd like to test it a couple of times before submitting.

PR #1457, opened 2015-06-16
New Rule: disallowSpaceBeforeSemicolon

rimunroe:
Disallow any spaces before semicolons (related to #1447)

```js
// "disallowSpaceBeforeSemicolon": true

// bad
var a ;

// good
var b;
```

The tests probably need some work.

### DIFF
--- a/test/specs/rules/disallow-space-before-semicolon.js
+++ b/test/specs/rules/disallow-space-before-semicolon.js
@@ -1,0 +1,31 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/disallow-space-before-semicolon', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    it('should not accept a false value as an option', function() {
+        assert.throws(function() {
+            checker.configure({ disallowSpaceBeforeSemicolon: false });
+        });
+    });
+
+    it('does not allow spaces before semicolons', function() {
+        checker.configure({ disallowSpaceBeforeSemicolon: true });
+
+        assert(checker.checkString('; ;').getErrorCount() === 1);
+        assert(checker.checkString('var a = 1 ;').getErrorCount() === 1);
+        assert(checker.checkString('var a = 2  ;').getErrorCount() === 1);
+    });
+
+    it('does allow semicolons with no spaces', function() {
+        checker.configure({ disallowSpaceBeforeSemicolon: true });
+
+        assert(checker.checkString('var a = 1;').isEmpty());
+    });
+});
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -744,6 +744,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-not-operators-in-conditionals'));
 
     this.registerRule(require('../rules/require-matching-function-name'));
+
+    this.registerRule(require('../rules/disallow-space-before-semicolon'));
 };
 
 /**
--- a/lib/rules/disallow-space-before-semicolon.js
+++ b/lib/rules/disallow-space-before-semicolon.js
@@ -1,0 +1,56 @@
+/**
+ * Disallows spaces before semicolons.
+ *
+ * Types: `Boolean`
+ *
+ * Values: `true` to disallow any spaces before any semicolon.
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowSpaceBeforeSemicolon": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var a = 1;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var a = 1 ;
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(option) {
+        assert(
+          option === true,
+          this.getOptionName() + ' option requires true value'
+        );
+    },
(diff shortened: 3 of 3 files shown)


rimunroe:
Also, I wasn't sure about the commit structure exactly because this doesn't directly fix an issue, just half of it.

hzoo:
You mean the commit message? You can just reference it (`Ref #1447`) instead of fix `Fixes #1447`)

However I think that issue should be about the semicolon spacing anyway since the comma discussion should of already been talked about in another issue.

rimunroe:
Yeah, I noticed that at least on my install I don't have any issues with commas anymore.

lib/rules/require-space-before-semicolon.js
```diff
@@ -0,0 +1,60 @@
...
+ * Requires space before semicolon.
+ *
+ * Types: `Boolean`
+ *
+ * Values: `true` to require a single space before any semicolon.
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireSpaceBeforeSemicolon": true
```
markelog: I wonder who would want that rule, yeah, we usually ask for the opposite rule, but i think we should be more soft on that.

markelog:
> Coverage decreased (-0.57%)

hm?

Otherwise LGTM.

rimunroe:
I have a bit of an ethical objection to allowing that rule to be honest, but I did know a person who coded with that style.

markelog:
You can check the coverage by running `npm run coverage-html && open coverage.html`

rimunroe:
It looks like the tests aren't covering anything in the two new rules, although I'm not exactly sure why.

markelog:
Oh, right :-), check this out - https://github.com/jscs-dev/node-jscs/blob/8de3850aff54e1519bd196a0627ee1cc3b6a96a0/lib/config/configuration.js#L563

You would need to change tests for disallow rule too

hzoo:
Yeah I don't see anyone asking for the opposite rule - I would be in favor of leaving it out

rimunroe:
@markelog Thanks! I'll fix that.

@hzoo I'd be happy to leave it out.

There are a few other tiny fixes as well that I'll push up in a few.

rimunroe:
It's passing fine now, and I changed the rule slightly so that it no longer looks at comments--or at least that's what I think the change does.

hzoo:
I think you can add a test for `var a = 1;` (no errors) as well.

rimunroe:
Done! Anything else?

hzoo:
@rimunroe LGTM - will need to squash/rebase your commits. Nice job!

rimunroe:
@hzoo What's the procedure for doing that at this point? Am I supposed to force push to my branch or is there some nicer way of doing it?

hzoo:
Sorry yeah you would rebase into a single commit and force push the branch.
It's ok since it's your own pr branch and not on master and you have to
rewrite the history

On Tuesday, June 16, 2015, Richard Munroe <notifications@github.com> wrote:

> @hzoo <https://github.com/hzoo> What's the procedure for doing that at
> this point? Am I supposed to force push to my branch or is there some nicer
> way of doing it?
>
> —
> <https://github.com/jscs-dev/node-jscs/pull/1457#issuecomment-112549204>.
>


-- 
Henry Zhu

rimunroe:
Spiffy. Anything else?

hzoo:
Don't think so - just wait on @markelog or someone else.

rimunroe:
Thanks for all the help!

test/specs/rules/disallow-space-before-semicolon.js
```diff
@@ -0,0 +1,31 @@
...
+            checker.configure({ disallowSpaceBeforeSemicolon: false });
+        });
+    });
+
+    it('does not allow spaces before semicolons', function() {
+        checker.configure({ disallowSpaceBeforeSemicolon: true });
+
+        assert(!checker.checkString('; ;').isEmpty());
+        assert(!checker.checkString('var a = 1 ;').isEmpty());
+        assert(!checker.checkString('var a = 2  ;').isEmpty());
```
markelog: Empty?

hzoo: Yeah it's `!` so not empty - but to be consistent I guess it should be changed to .getErrorCount() === 1) like the other tests @rimunroe

rimunroe: Ah okay. I had clicked on a few tests to see how the assertions were written, and just happened on one or two that check for non-emptiness. Is there a reason that this is used in some tests or is it just old code that hasn't been changed yet?

hzoo: Maybe it's old, but I think it makes sense to be specific when possible (even though I guess it's not a big deal)? It doesn't seem to be used that often and only when there is a large group of tests. Maybe we should change all those? 

Don't know the reason other than it's easier to type haha.

markelog: Just missed the `!` :-), but yeah, specific is better, although if we would be real sticklers we would ask you to put assertions in their very own `test` functions since it easier to isolated them with `it.only`.

markelog:
Thank you!

rimunroe:
@markelog did the old version get merged by accident?

hzoo:
Oh yeah it has `requireSpaceBeforeSemicolon` and no configuration file change

markelog:
Yeah, i had bad connection and failed to see `git fetch` has failed, merging local old version instead. 
Anyway, fixed, thank you for keeping an eye on it.

rimunroe:
@markelog @hzoo Thanks!

PR #1460, opened 2015-06-17
requirePaddingNewlinesBeforeKeywords: Modify special scenarios

### LINKED ISSUE #1092: Treat do while loops special for requirePaddingNewlinesBeforeKeywords
When I enable `requirePaddingNewlinesBeforeKeywords` for the while keyword that makes totally sense for the following case:

```
var a = true;

while (a) {
  something();
}
```

Where it make not really sense is in the do while case:

```
do {
  something();
} while (a);
```

Having this in place will make jscs complain about the fact that there is no newline before `while`.

It would be great to turn off that rule for a do while loop.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8742366-treat-do-while-loops-special-for-requirepaddingnewlinesbeforekeywords?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oredi:
Make special scenario for "elseif" perform index searching against a special case object.
Allow "while" that is part of a "dowhile" to succeed

Fixes #1092 

@mikesherov Please let me know if you think this is suitable.

### DIFF
--- a/lib/rules/require-padding-newlines-before-keywords.js
+++ b/lib/rules/require-padding-newlines-before-keywords.js
@@ -85,14 +85,16 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         var excludedTokens = [':', ',', '(', '='];
-
+        var specialCases = { 'if': 'else' };
         file.iterateTokensByTypeAndValue('Keyword', this._keywords, function(token) {
             var prevToken = file.getPrevToken(token);
-
-            // Handle special case of 'else if' construct.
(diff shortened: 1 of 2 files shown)


markelog:
Thank you!

PR #1477, opened 2015-06-22
requirePaddingNewlinesInBlocks: Add detail object to options

oredi:
Took a try to address this issue. Reviewers please take note that I did a fair amount of refactoring in the Test Suite to attempt to reuse test code and make the test more consistent throughout the various options that will be supported during configuration. I can always revert to the old way of testing hard code the individual tests for each supported option if that is the preferred practice.

### DIFF
--- a/lib/rules/require-padding-newlines-in-blocks.js
+++ b/lib/rules/require-padding-newlines-in-blocks.js
@@ -10,9 +10,13 @@
  *
  * ```js
  * "requirePaddingNewlinesInBlocks": true
+ * "requirePaddingNewlinesInBlocks": 1
+ * "requirePaddingNewlinesInBlocks": { "open": true, "close": true }
+ * "requirePaddingNewlinesInBlocks": { "open": false, "close": true }
+ * "requirePaddingNewlinesInBlocks": { "open": true, "close": false }
  * ```
  *
- * ##### Valid for mode `true`
+ * ##### Valid for mode `true` or `{ "open": true, "close": true }`
  *
  * ```js
  * if (true) {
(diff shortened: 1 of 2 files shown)


markelog:
> Took a try to address this issue

What issue, with the refactoring it is hard to see

oredi:
@markelog 
Sorry for not being clear in the PR description. The main issue that this PR fixes is the request to add an object option {open:(boolean), close: (boolean)} to the rule requirePaddingNewlinesInBlocks.

But when working on it I found the following:
    1. Similar tests had to be duplicated for each option supported. For example: "should report missing padding newline after opening brace"
    2. Some tests applied to value true but not the value 1

So I refactored the tests to do the following:
    1. applied to all options except for special cases "should not report single statement functions" that is applicable to only the number option
    2. reuse same code by making the test statement "abc();" or "abc();abc();" dynamic.

I wonder if this makes the PR too complex?

markelog:
It might be better to break down this commit on two, one that does test refactoring and the either that adds new option (is there an issue btw?), also, could add a docs for it?

oredi:
@markelog . Sure to make this cleaner I am thinking of doing the following:
1. Create a new issue for refactoring the tests for requirePaddingNewlinesInBlocks.
2. Create a new PR to for the fix of the new issue
3. Rework this change after PR in 2 is merged

FYI the issue for the new option is 529, it was dated sometime back

markelog:
Sounds good, although it is not necessary to create separate pr, it might be better, but it might not worth the fuzz

oredi:
@markelog I have created #1485 for tracking the change for the unit test. I will break the single commit into 2, first for refactoring the unit tests than another for the addition of a new option and update this PR.

oredi:
Hi @markelog,

I have separated the changes in the PR into 2 commits, I did a rebase to clean up the changes as well. The new option and its usage has been added in the rule file as well.

markelog:
Awesome! Thank you

PR #1479, opened 2015-06-23
Autofix trailing whitespaces and missing commas

Burnedo:
I've added autofix option for requireTrailingComma and disallowTrailingWhitespace rules. Please merge if you find it useful.


markelog:
Try `reportAndFix` test helper for autofixing - https://github.com/jscs-dev/node-jscs/search?utf8=%E2%9C%93&q=reportAndFix

Also, check this out https://github.com/jscs-dev/node-jscs/commit/f8d0360e39b6928a4323671d18e61be9e039022b

it is better move it in `fix` field, providing that you would change pr target to `2.0`

Burnedo:
@markelog 
Do you mean that I should move new tests to ./test/specs/rules/ and use reportAndFix instead of files?

What is the _fix field?

markelog:
> Do you mean that I should move new tests to ./test/specs/rules/ and use reportAndFix instead of files?

Yep, this is how we usually check autofix feature

> What is the _fix field?

See https://github.com/jscs-dev/node-jscs/blob/ff7a6a1f4366bfc3c9004d79946bfec0ef570259/lib/rules/validate-quote-marks.js in 2.0 branch

Burnedo:
I've refactored the code and the tests.
Regarding 2.0 version I'll try to fix it on another branch and create a new pull request for it.
Can this be merged into current master?

markelog:
> Regarding 2.0 version I'll try to fix it on another branch and create a new pull request for it.

With the `_fix` field? Promise :-)?

Burnedo:
@markelog It doesn't work by fast and simple copy-pasting the code :)
I have to merge my changes into 2.0 and dive deeper into the code. I'll definitely do this, but can't tell exact ETA for now.

test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 test/specs/**/*.js
 -R dot
--u bdd
+-u bdd
```
markelog: ?

test/data/autofixing/spaces.js
```diff
@@ -1,3 +1,3 @@
 var y=2;
 var x=y+1;
-if(x);
+if(x);
```
markelog: ?

lib/errors.js
```diff
@@ -30,8 +30,9 @@ Errors.prototype = {
      * @param {String} message
      * @param {Number|Object} line
      * @param {Number} [column] optional if line is an object
+     * @param {Booleab} [fixed] optional if error is fixed
```
markelog: `Booleab `

Burnedo: Fixed, sorry about that.

test/mocha.opts
Burnedo: I edited this while developing and accidentally pushed the change, then restored original content.

test/data/autofixing/spaces.js
Burnedo: Same as in mocha.opts file.

markelog:
@zz85 could you check this out?

since this is autofix ticket /cc @mdevils

zz85:
Some comments with regards to the autofixing trailing spacing rule in this PR and #1480.

- this is an example where it's interesting the code handling ASTs has increased complexity compared to a solution with employs only regex
- @Burnedo approach here seems decent. This was the first approach I took working on this issue. The difference between this PR and #1480 is that there's a little of backtracing here while it's a slightly simpler loop with no backtracking in other PR. 
- I've looked at instance._fix() and errors.cast(). If we accept this pr errors.add() which takes in an fixed flag, then maybe _fix isn't required for this rule. I would understand having fixes and error checking in different passes could have performant, however the implementations for trailing _fix seems like it would just be additional overheads
- It is not clear whether we should target 2.0 or master branch, and since github doesn't allow switch of branches after opening a PR, might have to open another
- This PR does not take into account trailing spaces in block comments, therefore the following these would fail here
```js
reportAndFix({
    name: 'should fix trailing space in block comments',
    rules: { disallowTrailingWhitespace: true },
    input: '/*\nbla\n\t\n*/',
    output: '/*\nbla\n\n*/'
});

reportAndFix({
    name: 'should ignore empty lines in block comments',
    rules: { disallowTrailingWhitespace: 'ignoreEmptyLines' },
    input: '/* \nbla\n\t\n*/',
    output: '/*\nbla\n\t\n*/'
});
```
- apart from this, other tests from the other PR passes here and so all tests in this PR passes in other PR.
- I think the jscs team should pick an implementation they find works the best, I'll be happy autofix for trailing spaces lands quickly so I could close https://github.com/zz85/mrdoobapproves/issues/38

Burnedo:
@zz85 First test case is passing, but you are right about second. IgnoreEmptyLines won't work for comments. Will fix.

Burnedo:
I've covered tests mentioned by @zz85

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
     check: function(file, errors) {
         var ignoreEmptyLines = this._ignoreEmptyLines;
-
         var lines = file.getLines();
+
         for (var i = 0, l = lines.length; i < l; i++) {
+
             if (lines[i].match(/\s$/) && !(ignoreEmptyLines && lines[i].match(/^\s*$/))) {
-                errors.add('Illegal trailing whitespace', i + 1, lines[i].length);
+                var fixed = false;
```
zz85: what is fixed use for?

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+                var startLineNumber;
+                var precendingToken;
+                var targetToken;
+
+                while (!precendingToken && currentLineNumber > 0) {
+                    precendingToken = file.getLastTokenOnLine(currentLineNumber, {includeComments: true});
+                    currentLineNumber--;
+                }
+
+                if (typeof precendingToken === 'undefined') {
```
zz85: should it be sufficient to do `precendingToken === 'undefined'`?

lib/rules/disallow-trailing-whitespace.js
zz85: ok, just saw its use below...

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+                                    if (!blockLines[k].match(/^\s*$/) || k === 0) {
+                                        blockLines[k] = blockLines[k].split(/\s$/).join('');
+                                    }
+                                }
+                                precendingToken.value = blockLines.join('\n');
+                            } else {
+                                precendingToken.value = precendingToken.value.split(/\s\n/).join('\n');
+                            }
+                        } else {
+                            precendingToken.value = precendingToken.value.split(/\s$/).join('');
```
zz85: `.split().join()` could be replaced with `replace()`

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+
+                                for (var k = 0; k < blockLines.length; k++) {
+
+                                    if (!blockLines[k].match(/^\s*$/) || k === 0) {
+                                        blockLines[k] = blockLines[k].split(/\s$/).join('');
+                                    }
+                                }
+                                precendingToken.value = blockLines.join('\n');
+                            } else {
+                                precendingToken.value = precendingToken.value.split(/\s\n/).join('\n');
```
zz85: should `file.getLineBreakStyle()` be used instead of `\n`?

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+
+                if (typeof precendingToken === 'undefined') {
+                    targetToken = file.getFirstToken();
+                    startLineNumber = 1;
+                } else {
+                    targetToken = file.getNextToken(precendingToken, {includeComments: true});
+                    startLineNumber = precendingToken.loc.end.line;
+
+                    if (precendingToken.isComment &&
+                        precendingToken.loc.start.line <= (i + 1) &&
```
zz85: is ` precendingToken.loc.start.line <= (i + 1) &&  precendingToken.loc.end.line >= (i + 1)` essentially `precendingToken.type === 'Block'`?

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+                    var targetLine = lines[targetToken.loc.start.line - 1];
+                    for (var j = 0, whitespace = targetLine.charAt(j);
+                        whitespace.match(/\s/);
+                        j++, whitespace = targetLine.charAt(j)) {
+                        targetIndent += whitespace;
+                    }
+                    targetToken.whitespaceBefore = Array(eolCount).join('\n') + targetIndent;
+                    fixed = true;
+                }
+                precendingToken = undefined;
```
zz85: this could be set at where `var precendingToken` is.

lib/rules/disallow-trailing-whitespace.js
```diff
@@ -64,13 +64,70 @@ module.exports.prototype = {
...
+                        }
+                        fixed = true;
+                    }
+                }
+
+                if (typeof targetToken !== 'undefined' && !fixed) {
+                    var eolCount = targetToken.loc.start.line - startLineNumber + 1;
+                    var targetIndent = '';
+                    var targetLine = lines[targetToken.loc.start.line - 1];
+                    for (var j = 0, whitespace = targetLine.charAt(j);
```
zz85: this can be rewritten as `for (var j = 0, whitespace ; whitespace = targetLine.charAt(j), whitespace.match(/\s/); j++)`

lib/rules/disallow-trailing-whitespace.js
Burnedo: It will never be true. It could be written like ``` precendintToken === undefined ```, but I believe it's a better practice to use typeof.

lib/rules/disallow-trailing-whitespace.js
Burnedo: In case of regexp you are right.

lib/rules/disallow-trailing-whitespace.js
Burnedo: ``` file.getLineBreakStyle() ``` is not very strict, so I decided to just use hardcoded new line. Using ``` file.getLineBreakStyle() ``` would enforce me to use RegExp constructor and make code even less readable.

lib/rules/disallow-trailing-whitespace.js
Burnedo: This condition is to determine if we should fix this token, not if it's a block comment. precendintToken in some cases could be not at the line we are trying to fix in current iteration.

lib/rules/disallow-trailing-whitespace.js
Burnedo: No. We have to set precendingToken to undefined at the end of iteration because of a strange (not intuitive) javascript scope behavior in loops.

lib/rules/disallow-trailing-whitespace.js
Burnedo: Is it matter of choice or significantly more performant?

zz85: it is a matter of choice as well as being more succinct and generating less code

markelog:
So @zz85, can i get "LGTM"?

markelog:
One more thing, is it possible for you to move autofix logic to the `token-assert` module? Otherwise we will have three place where we fix things.

zz85:
:+1: if this gets cleaned up.

markelog:
@Burnedo will you be able to make the necessary changes, until monday, since we're planning to release then.

If not, it is okay, i think we can move this logic to `token-assert` module ourselves.

Burnedo:
@markelog I'm sorry, but I have no time at all this week. It would be great if you could take care of this issue. Thank you.

markelog:
np, you did an awesome thing here!

markelog:
Landed, thank you!

hzoo:
Nice!

PR #1480, opened 2015-06-23
Support autofix for disallow-trailing-whitespace

zz85:
- uses AST tokens by lines instead of just line strings
- uses errors.assert to emit errors
- add disallowTrailingWhitespace rule to mdcs preset
- fixes https://github.com/zz85/mrdoobapproves/issues/38

This seems to work fine - I suspect however some of the code could be refactored to use `TokenAssert`.

cc @gero3 @mikesherov @markelog

### DIFF
--- a/lib/rules/disallow-trailing-whitespace.js
+++ b/lib/rules/disallow-trailing-whitespace.js
@@ -64,13 +64,89 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         var ignoreEmptyLines = this._ignoreEmptyLines;
+        var lineTokens = file._tokensByLineIndex;
+        var lines = file.getLines().length;
+        var lastLine = 1;
 
-        var lines = file.getLines();
-        for (var i = 0, l = lines.length; i < l; i++) {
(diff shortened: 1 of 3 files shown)


zz85:
Just realize that there is also another PR on trailing spaces https://github.com/jscs-dev/node-jscs/pull/1479

zz85:
looks like I've some code coverage to fix. i should also add some autofixing tests, but i'm not where that would go...

image

markelog:
See the comment i left in another pr

zz85:
@markelog thanks. will try to add the tests

zz85:
Updated with tests.

zz85:
oops clumsy fingers

markelog:
How about 

> it is better move it in _fix field, providing that you would change pr target to 2.0

?

zz85:
@markelog do you mean that we should use _fix and target 2.0 instead master?

markelog:
Yeah

zz85:
added comments to #1479. also, this travis build seems to be broken only because i closed and re-opened this PR?

markelog:
Thank you @zz85 for this pull and for the #1479 review!

PR #1492, opened 2015-06-26
maximumLineLength: Add exception for long require expressions

phlip9:
Some of our longer `require` expressions break our 80 character line limit. We'd like to keep this limit enforced everywhere, except on the lines that `require` modules.

This pull request adds an option to the `maximumLineLength` rule, in the `allExcept` array, with name `require`. If added, any line with a `require` call expression is ignored when checking for line-length violations.

### DIFF
--- a/lib/rules/maximum-line-length.js
+++ b/lib/rules/maximum-line-length.js
@@ -13,6 +13,7 @@
  *        - `comments`: allows comments to break the rule
  *        - `urlComments`: allows comments with long urls to break the rule
  *        - `functionSignature`: allows function definitions to break the rule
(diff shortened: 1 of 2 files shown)


mrjoelkemp:
Nice work here @phlip9. Thanks for contributing! Looks good, but a test showing that a require call reports without the option on would assert that the rule still works properly.

phlip9:
@mrjoelkemp Gotcha! I added another test as you requested--everything looks like it works correctly.

mrjoelkemp:
Thanks again @phlip9.

PR #1494, opened 2015-06-26
requireArrowFunctions: only error for callbacks

### LINKED ISSUE #1483: requireArrowFunctions: only check for FunctionExpression as a callback?
Similar to #1413 

So I think we can only warn when using `FunctionExpression` as a callback since in all the other cases we would have to do work to check for `this`/scope.

So in my next PR I would make an exception for `node.parent.type === 'VariableDeclarator'` or `ReturnStatement`

Or maybe since its only one case now it should somehow only check if it's a callback (`!node.parent`)?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22936641-requirearrowfunctions-only-check-for-functionexpression-as-a-callback?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1483 

This ok - only checks callbacks since we can't be sure otherwise without a lot of other checks?

### DIFF
--- a/lib/rules/require-arrow-functions.js
+++ b/lib/rules/require-arrow-functions.js
@@ -1,5 +1,5 @@
 /**
- * Requires that arrow functions are used instead of anonymous function expressions.
+ * Requires that arrow functions are used instead of anonymous function expressions in callbacks.
  *
  * Type: `Boolean`
  *
@@ -28,6 +28,12 @@
  * var x = { bar() {} }
  * // class method
  * class Foo { bar() {} }
+ * // function expression in a return statement
+ * function a(x) {
+ *     return function(x) { return x };
+ * };
+ * // function expression in a variable declaration
+ * var a = function(x) { return x };
  * ```
  *
  * ##### Invalid
@@ -37,12 +43,6 @@
  * [1, 2, 3].map(function (x) {
  *     return x * x;
  * });
- * // function expression in a return statement
- * function a(x) {
- *     return function(x) { return x };
- * };
- * // function expression in a variable declaration
- * var a = function(x) { return x };
  * ```
  */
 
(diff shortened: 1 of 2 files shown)


lib/rules/require-arrow-functions.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Requires that arrow functions are used instead of anonymous function expressions.
+ * Requires that arrow functions be used instead of anonymous function expressions in callbacks.
```
mrjoelkemp: Keep the `are`. `Requires that arrow functions are used instead of anonymous function expressions in callbacks`

mrjoelkemp:
The weakening of the rule makes sense. We'll see how folks use it. 

LGTM

hzoo:
Ok fixed although getting the coveralls issue again

markelog:
Hm, since fat function is statically binds `this`, is it that hard to check the scope? We don't want it report if parent is `Program` node? Can we just check it?

hzoo:
Don't know how we can check the scope (don't think we have any other rules doing this)

```js
// Example with an object (VariableDeclaration, VariableDeclarator, ObjectExpression, Property, FunctionExpression)
var a = {
  b: 1,
  normalFunction: function() { return this.b; },
  arrowFunc: () => { return this.b; }, // undefined
  methodShorthand() { return this.b; },
  arrowFuncInsideFunc() {
    return () => { return this.b; } // not undefined
  }
};
```

We can't just check `node.parent` is `Program` because it could be like `node.parent.parent.parent...` - or maybe there's a way

markelog:
> (don't think we have any other rules doing this)

They don't, that is an interesting part :-).

If we would go that route, we would need to find first scope, which should be pretty easy, since there is only handful constructions that creates closure - `let`, `with`, `function`, `class` and `const`. 

Checked syntax should have one of those fairly early in the upper stack, if it is not it means it in the global context, i.e. parent = `Program`

mikesherov:
We could also use `escope` if we're interested in scope analysis.

hzoo:
> They don't, that is an interesting part :-).

Good point haha. Although maybe we should land this first and then see if we want to do that later?

hzoo:
Should we make a different issue for scope tracking if we want to do that?

markelog:
Done, could you create an issue based on this discussion?

PR #1498, opened 2015-06-29
Reporters: add new summary reporter

### LINKED ISSUE #1047: add -l option. List violating files.
I'm currently trying to add jscs to an existing codebase. I'm fiddling with options and would really love to know which files are problematic.

Feature Request: `jscs . -l` prints out a list of violating files. The number of infringements could be listed next to the filename as well or be optional.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8502631-add-l-option-list-violating-files?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oredi:
This reporter provides a summary of all files with errors found listing each filename against the error count. The following shows a sample of the output
    file.js: 1 error found.
    file2.js: 2 errors found.

Fixes #1047

### DIFF
--- a/test/specs/reporters/summary.js
+++ b/test/specs/reporters/summary.js
@@ -1,0 +1,110 @@
+var assert = require('assert');
+var sinon = require('sinon');
+var rewire = require('rewire');
+
+var Checker = require('../../../lib/checker');
+var summaryReporter = rewire('../../../lib/reporters/summary');
+
+describe('reporters/summary', function() {
+    var checker;
+
+    var scenarios = [
+        {
+            statement: 'with (x) {}',
+            results: {
+                byFile: {
+                    file: 'input',
+                    totalError: 1
+                },
+                byRule: [
+                    {
+                        name: 'disallowKeywords',
+                        totalError: 1,
+                        fileError: 1
+                    },
+                    {
+                        name: 'All',
+                        totalError: 1,
+                        fileError: 1
+                    }
+                ]
+            }
+        },
+        {
+            statement: 'with(x){} with(x){}',
+            results: {
+                byFile: {
+                    file: 'input',
+                    totalError: 2
+                },
+                byRule: [
+                    {
+                        name: 'disallowKeywords',
+                        totalError: 2,
+                        fileError: 1
+                    },
+                    {
+                        name: 'All',
+                        totalError: 2,
+                        fileError: 1
+                    }
+                ]
+            }
+        }
+    ];
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    describe('valid code syntax', function() {
+        it('should correctly reports no errors', function() {
+            checker.configure({ disallowKeywords: ['with'] });
+            sinon.stub(console, 'log');
+            summaryReporter([checker.checkString('a++;')]);
+            assert.equal(console.log.getCall(0).args[0], 'No code style errors found.');
+            assert(console.log.calledOnce);
+            console.log.restore();
+        });
+    });
+
+    describe('invalid code syntax', function() {
+        scenarios.forEach(function(scenario) {
+            describe(JSON.stringify(scenario.statement), function() {
+                beforeEach(function() {
+                    checker.configure({ disallowKeywords: ['with'] });
+                    sinon.stub(console, 'log');
+                    summaryReporter([checker.checkString(scenario.statement)]);
+                });
+
+                afterEach(function() {
+                    console.log.restore();
+                });
+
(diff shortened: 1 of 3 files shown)


mrjoelkemp:
LGTM. Nice work!

markelog:
Could you post screenshot of the summary output?

markelog:
Is this analogue of the https://github.com/jscs-dev/node-jscs/pull/601?

oredi:
@markelog I think #601 is doing a more detailed report than this. I could change the name if required

The reporter implemented by this PR will only show the list of file names with the error count for files with errors found by the rules. The following shows a sample of the output

    file.js: 1 error found.
    file2.js: 2 errors found.

I will do a screenshot of a run when I get back on my PC in another 8 hours.

markelog:
How about we unite two approaches? first we could show paths with error count and then rule table - http://imgur.com/FBX0C75

hzoo:
@markelog You can add images to github from the clipboard - and lol I noticed that's what the message below says (although maybe you knew that already)

oredi:
@markelog  I can look into that I believe we could implement this reporter to cover both. I see cli-table as one of the project dependencies so I will use that to create 2 tables showing different summaries. I will update this PR with the new code but it might take some time though.

oredi:
@markelog I have modified the PR making 2 commits for fixing #453 and #1047.
This report will now contain 2 tables as shown below, I did not style the tables but it can be done if required.

image

test/specs/reporters/summary.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    it('should correctly reports no errors', function() {
+        summaryReporter([checker.checkString('a++;')]);
+        assert.equal(console.log.getCall(0).args[0], 'No code style errors found.');
+        assert(console.log.calledOnce);
+    });
+
+    it('should correctly report 1 error', function() {
+        var expectedByFile = [
```
markelog: Here, it seems you testing, although quite elegantly, logic of the external tool?

markelog:
> for fixing #453

No connections to the new reporter

> I did not style the tables but it can be done if required.

Yeah, that would be preferable, it seems, wow effect of the #601 partly was achieved because of the styling thing.

Also, could align width of those tables?

oredi:
Hi @markelog,

I have merged the 2 commits into a single one since the relevance to the old issue is no longer there. 

Styling to the table has also been applied and I made the widths of both tables similar as well. Please take note that the wow factor introduces an unwanted side effect. In order to make the tables align I had to make the columns fixed width summing to a total of 80 characters (taking 80 characters as the default for terminals)). For error count columns they are usually fine but the rule (49) and path (62) columns might be truncated as shown in the image.  

By checking the length of the rule filenames, I think the longest rule in jscs is now 45 characters so unless someone comes up with a very long rule the path column might be the one with a higher chance of being truncated.

image

I have written the original tests to mimic the output since this is a test suite for a reporter so I thought have the formatted output might make it clearer about what the reporter does. But as per your comments and while working on the styling I found that it make the tests brittle and easily broken as it needs a fair amount of inside knowledge on cli-table. So I have also modified the tests to use regular expressions to reduce the correlation of the tests with cli-table.

Please let me know if you feel that more changes are required.

markelog:
> I found that it make the tests brittle

Yeah, that is one of the reason i was asking about that, `console` checks could be really tricky, just like regexps, perhaps you could change those tests to `indexOf` instead?

Also, red looks kinda aggressive, i understand that perhaps you chose that color because those are errors and errors are usually red. But i would propose to switch to more friendly color, like green one.

oredi:
For the styling I actually copied from the original code of #601, I have made the easy change and I have also modified the headers to make more space for the path or rule column
image

For the tests I was testing out the use of the indexOf and found that the console output actually contains a lot of unicode use for styling so the checking of value for "Errors" or "No of Files" becomes very tricky.

As an alternative I found in package.json the rewire module, with that I was able to access the raw data used in cli-table to test the values rather than messing with the formatting and console related output. It has some drawbacks listed below although I think it is better as the tests are more accurate as only the data is tested instead of data meshed up with display content.

Drawbacks
    1. Introduces variables outside the exported module function (I think this is fine as the file is meant to be used in node
    2. Test file needs to know inners of cli-table and also internal variables of the reporter.

markelog:
Awesome! Thank you

PR #1500, opened 2015-06-29
Changed the right alignment

kolomiichenko:
Changed the right alignment. 

**_Valid**_

``` javascript
var x = {
    a:   1,
    bcd: 2,
    ef:  'str'
};
```

**_Invalid**_

``` javascript
var x = {
    a   : 1,
    bcd : 2,
    ef  : 'str'
};
```

Fixed a conflict between [disallowSpaceAfterObjectKeys](http://jscs.info/rule/disallowSpaceAfterObjectKeys.html) and [requireAlignedObjectValues](http://jscs.info/rule/requireAlignedObjectValues.html) rules.

### DIFF
--- a/lib/rules/require-aligned-object-values.js
+++ b/lib/rules/require-aligned-object-values.js
@@ -100,13 +100,18 @@ module.exports.prototype = {
                 return;
             }
 
-            var space = minColonPos - maxKeyEndPos;
             tokens.forEach(function(pair) {
                 errors.assert.spacesBetween({
                     token: pair.key,
                     nextToken: pair.colon,
-                    exactly: maxKeyEndPos - pair.key.loc.end.column + space,
-                    message: 'Alignment required'
+                    exactly: 0,
+                    message: 'Space after colon forbidden'
+                });
+                errors.assert.spacesBetween({
+                    token: pair.colon,
+                    nextToken: file.getNextToken(pair.colon),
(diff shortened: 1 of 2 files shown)


markelog:
Could you elaborate? `requireAlignedObjectValues` should enforce 

```js
var x = {
    a   : 1,
    bcd : 2,
    ef  : 'str'
};
```

kolomiichenko:
When you set `"disallowSpaceAfterObjectKeys": true` and `"requireAlignedObjectValues": "all"` rules will conflict because first disallow spaces before collons. 
So I think that pattern alignment is fundamentally wrong and I offer my alignment template.
But maybe need to modify, and make the choice of alignment template...

markelog:
There is a lot of rules that conflict with each other, like <a href="http://jscs.info/rule/requireSemicolons.html">`requireSemicolons`</a> and <a href="http://jscs.info/rule/disallowSemicolons.html">`disallowSemicolons`</a>.

Current style of `requireAlignedObjectValues` is fairly popular, it seems it would better to introduce `disallowAlignedObjectValues` rule for your particular use-case.

hzoo:
I think you would just make an option for `requireAlignedObjectValues` to align by the key or colon since some people would like either?

markelog:
Still interested?

markelog:
Okay, ping us up if you change your mind

PR #1503, opened 2015-07-01
yodaConditions: allow literals on the left side of comparison operators (<, >, <=, >=)

segv:
There are a number of reasons for disallowing `LITERAL === VARIABLE`
in code, however there is no similar reason to disallow `LITERAL <
VARIABLE` and, especially for array bounds or other limit checks, the version where the
constant (generally 0) comes first is more readable.

example code (which caused this pull request in the first place):

    if (0 <= index) {
      ...
    }

### DIFF
--- a/lib/rules/disallow-yoda-conditions.js
+++ b/lib/rules/disallow-yoda-conditions.js
@@ -43,11 +43,7 @@ module.exports.prototype = {
             '==': true,
             '===': true,
             '!=': true,
-            '!==': true,
-            '>': true,
-            '<': true,
-            '>=': true,
-            '<=': true
+            '!==': true
         };
     },
 


zxqfox:
We've discussed that in #1031

I'm strongly agree that `<=`, `>=`, etc should be enabled by default, but we need an agreement of all to change this.

@jscs-dev/jscs-maintainer Can we rethink that?

markelog:
>I'm strongly agree that <=, >=, etc should be enabled by default

So, you against these changes?

zxqfox:
Wow, sorry. Shouldn't!
5 июля 2015 г. 21:56 пользователь Oleg Gaidarenko <notifications@github.com> написал:

markelog:
Okay then :-), it is a breaking change so merged in 2.0, since this pr was incomplete, i did it myself, @segv i hope you don't mind.

Issue #1359, opened 2015-05-11
Make jsDoc plugin part of the jscs dist


markelog:
It could be in another repo, but a dependency of jscs.

/cc @zxqfox

zxqfox:
Mmm?

markelog:
Purpose of this ticket is to discuss inclusion of jsdoc rules to the jscs distribution

mrjoelkemp:
"new rule"? Can you further describe what the inclusion entails? Is this for the website?

markelog:
"new rule" in sense of lack of such rule(s) in the current jscs distro

mrjoelkemp:
So you want users to be able to use jscs-jsdoc from jscs directly. Got it. 

What is painful about the current solution of it being an external plugin? i.e. What are the benefits? The separation of concern is clear as it stands.

markelog:
> What is painful about the current solution of it being an external plugin?

Nothing painful, it would improve convenience though, and improve popularity of such rule(s), since we would document it on the site and could use them in the presets.

> The separation of concern is clear as it stands.

Separation would stay, since rule(s) would be in another repo.

zxqfox:
> since we would document it on the site and could use them in the presets.

Mmm, but we can document them without inclusion into jscs dist, no?
I'm just trying to find a way to write other plugins with documenting at jscs.info.

markelog:
> Mmm, but we can document them without inclusion into jscs dist, no?

That would be a complicated task.

> I'm just trying to find a way to write other plugins with documenting at jscs.info.

That might be unnecessary

mrjoelkemp:
Plugins are in dedicated repos though, so they could also have dedicated github pages that we could link to within our overview.

What we're saying, in a more general sense, is that org-owned plugins are likely to be brought into the core (unifying the config, not the code). 

Do we intend on keeping the `jsDoc` key as part of the configuration or flattening it to the main jscsrc level with this integration?

zxqfox:
I'd say, if we want to show docs on jscs.info and/or build dist with it we should be sure in
- rules inside plugin are the first class rules (there are some dirty things that couldn't be done without changes in jscs core. e.g. jscs-dev/jscs-jsdoc#61);
- docs formatted as well as in core;
- the same support as in core.

E.g. atm it's not easy to use plugins for projects via samba/afp. IDE plugins looking for a global jscs and failing on searching project plugins. And also it's hard to use different versions of jscs for different projects. Etc.

[...]

PR #1508, opened 2015-07-05
Misc: set default value of maxErrors option to 50

### LINKED ISSUE #681: Change the default maxError to 50
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5275604-change-the-default-maxerror-to-50?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
* Set default value of maxErrors option to 50
* Do not consider maxErrors value at autofix execution
* Rename testsuites of root modules
  (since they should point to the path of module their testing)

Fixes #681

/cc @mikesherov @mdevils

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -226,7 +226,7 @@ Set the maximum number of errors to report
 
 Type: `Number`
 
-Default: Infinity
+Default: 50
 
 #### Example
 
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -111,7 +111,7 @@ function Configuration() {
      * @protected
      * @type {Number}
      */
-    this._maxErrors = Infinity;
+    this._maxErrors = 50;
 
     /**
      * JSCS CWD.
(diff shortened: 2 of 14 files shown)


mrjoelkemp:
Not seeing new logic about how this prevents autofix from being affected. A test would suffice for reassurance. 

Lgtm, but this is for sure a breaking change. So merge in the 2.0 branch? Still haven't heard if we're doing 2.0 first or 1.14.

markelog:
> A test would suffice for reassurance.

For sure, will add one

> Lgtm, but this is for sure a breaking change. So merge in the 2.0 branch? Still haven't heard if we're doing 2.0 first or 1.14.

That is the question, isn't it :-). It is good thing that we have two branches, there is three things need to be done, in order to release 2.0

1. https://github.com/jscs-dev/node-jscs/issues/986
2. #1359
3. https://github.com/jscs-dev/node-jscs/issues/1353

#986 requires more discussion, so it might closed as won't fix, others, mostly depend on free time of the 
@zxqfox and @hzoo.

It seems #1359 is pretty easy to do and @hzoo solved most of complicated issues of #1353, one thing is left is one little step.

So i'm thinking, if we could do those three things quickly, like week at most, we might as well, release 2.0.

hzoo:
https://github.com/jscs-dev/node-jscs/pull/1494 left on the actual 1.14 milestone but I'm assuming there might be a few more?

mrjoelkemp:
Thanks for the explanation @markelog. 2.0 on 7/14. Sounds like a plan.

PR #1515, opened 2015-07-08
add support for babel-jscs

hzoo:
monkeypatch (specifically use of Module._resolveFilename) fails
on linux without symlink

removed monkeypatch method on babel-jscs and moved patching estraverse
to jscs itself.

Ref #1446

Pinning babel-jscs to PR to make it pass.

### DIFF
--- a/lib/tree-iterator.js
+++ b/lib/tree-iterator.js
@@ -1,4 +1,7 @@
 var estraverse = require('estraverse');
+var assign     = require('lodash.assign');
+var types      = require('babel-core').types;
+assign(estraverse.VisitorKeys, types.VISITOR_KEYS);
 
 module.exports = {
     iterate: iterate
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
     "node": ">= 0.10.0"
   },
   "dependencies": {
-    "babel-jscs": "^1.0.1",
+    "babel-core": "^5.6.15",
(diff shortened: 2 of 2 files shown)


hzoo:
Ok wow it's all passing now - @markelog @zxqfox @mrjoelkemp

markelog:
Thank you @hzoo Hooray!

hzoo:
I can merge https://github.com/jscs-dev/babel-jscs/pull/10/files so we can change `jscs-dev/babel-jscs#remove-monkeypatch`. Would that be a major version so 2.0.0 or just 1.0.4?

mrjoelkemp:
It's a major bump since folks using babel-jscs will hit errors without the
latest jscs master since you removed the monkeypatch.

hzoo:
I was thinking that although I put in the readme to use master for jscs anyway since no release yet - ok thanks major then.

hzoo:
Also @markelog - can we update estraverse now?

https://github.com/jscs-dev/node-jscs/blob/master/lib/tree-iterator.js#L65 So when you remove add `handler` as well some tests fail.

We won't need https://github.com/jscs-dev/node-jscs/blob/master/lib/tree-iterator.js#L39-L49 since it would be covered in babel keys already. I don't know if we need the XJS ones since that's deprecated.

PR #1519, opened 2015-07-10
Misc: remove all deprecated rules/tests

### LINKED ISSUE #993: validateJSDoc: Drop deprecated rule
Ref #778

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8271312-validatejsdoc-drop-deprecated-rule?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
- remove (require|disallow)(Left|Right)stickedOperators
- remove validateJsdoc

Fixes #993

Do we just remove the files and add a note in the changelog?


mrjoelkemp:
Forgot about this item. Nice!

With the docs generated from the rules, there are no readme changes to make (thankfully). This should be good enough.

LGTM

lib/config/configuration.js
```diff
@@ -828,11 +828,6 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-parentheses-around-iife'));
 
     /* deprecated rules */
-    this.registerRule(require('../rules/require-left-sticked-operators'));
-    this.registerRule(require('../rules/disallow-left-sticked-operators'));
-    this.registerRule(require('../rules/require-right-sticked-operators'));
-    this.registerRule(require('../rules/disallow-right-sticked-operators'));
-    this.registerRule(require('../rules/validate-jsdoc'));
     /* deprecated rules (end) */
```
zxqfox: What about this comments? ;-)

hzoo: I just left it in since its comment and we'll prob have more deprecated stuff

markelog: I would remove it, if we will have something else, we will add it again :-)

PR #1521, opened 2015-07-10
New Rule: requireVarDeclFirst

### LINKED ISSUE #99: Add option to enforce variable declaration at the top of declaring scope

oredi:
Enforce variable declaration at the top of declaring scope

Fixes #99

### DIFF
--- a/lib/rules/require-var-decl-first.js
+++ b/lib/rules/require-var-decl-first.js
@@ -21,6 +21,11 @@
  *     y = 2;
  * ```
  * ```js
+ * 'use strict;'
+ * var x = 1,
+ *     y = 2;
+ * ```
+ * ```js
  * var x = 1;
  * var y = 2;
  * ```
@@ -54,6 +59,13 @@
  * // comments
  * // comments 2
  *  var z;
+ * };
+ * ```
+ * ```js
+ * var w = 1;
+ * function x() {
+ *  "use strict";
+ *  var y;
  * };
  * ```
  * ```js
@@ -115,6 +127,24 @@ function getOffsetForBlockStatement(enclosingScope, varDecl, commentTokens) {
         offset += getCommentOffsetBetweenNodes(parentNode, varDecl, commentTokens);
     }
     return offset;
+}
+
+function getUseStrictDeclFirst(enclosingScope) {
+    var firstNode;
+    if (enclosingScope.type === 'Program') {
+        firstNode = enclosingScope.body[0];
+    } else {
+        firstNode = enclosingScope.body.body[0];
+    }
+
+    if (firstNode.type === 'ExpressionStatement' &&
+        firstNode.hasOwnProperty('expression') === true &&
+        firstNode.expression.hasOwnProperty('value') === true &&
+        firstNode.expression.value === 'use strict') {
+        return firstNode;
+    }
+
+    return null;
 }
 
 function isFirstVarDeclInScope(enclosingScope, varDecl, whitespaceOffsetBeforeVarDecl, commentTokens) {
@@ -208,6 +238,7 @@ module.exports.prototype = {
             var enclosingScope;
             var scopeContents;
             var previousNode;
+            var useStrictDirective;
             var isVarDeclFirst = false;
 
             var whitespaceOffsetBeforeVarDecl = file.getFirstNodeToken(varDecl).whitespaceBefore.length;
@@ -215,12 +246,22 @@ module.exports.prototype = {
             enclosingScope = getVariableScope(varDecl.parentNode);
             if (!scopesFoundInFile.hasOwnProperty(enclosingScope.range[0])) {
                 scopesFoundInFile[enclosingScope.range[0]] = { hasNonVarDecl: false, varDecl: [] };
+                // placing the handling 'use strict' declared as the first statement of scope here to improve
+                // performance to run once per scope discovered in file
+                useStrictDirective = getUseStrictDeclFirst(enclosingScope);
+                if (useStrictDirective !== null) {
+                    // Special case to make varDecl stack contain the use strict as first node
+                    // this reduces the complexity of the isFirstVarDecInScope and reuses
+                    // isPreviousNodeAVarDecl to handle this special scenario
+                    scopesFoundInFile[enclosingScope.range[0]].varDecl.push(useStrictDirective);
+                }
             }
+
             scopeContents = scopesFoundInFile[enclosingScope.range[0]];
 
             if (scopeContents.varDecl.length === 0) {
                 isVarDeclFirst = isFirstVarDeclInScope(
-                    enclosingScope, varDecl, whitespaceOffsetBeforeVarDecl, commentTokens);
+                    enclosingScope, varDecl, whitespaceOffsetBeforeVarDecl, commentTokens, file);
             } else {
                 previousNode = scopeContents.varDecl[scopeContents.varDecl.length - 1];
                 if (!scopeContents.hasNonVarDecl) {
--- a/lib/rules/require-var-decl-first.js
+++ b/lib/rules/require-var-decl-first.js
@@ -1,0 +1,240 @@
+/**
+ * Requires `var` declaration to be on the top of an enclosing scope
+ *
+ * Types: `Boolean`
+ *
+ * Values: `true`
+ *
+ * if `requireVarDeclFirst` defined as a `true` value, it will report errors of `var` declarations that
+ * does not appear on the top of a function scope.
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireVarDeclFirst": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var x = 1,
+ *     y = 2;
+ * ```
+ * ```js
+ * var x = 1;
+ * var y = 2;
+ * ```
+ * ```js
+ * var x = 1;
+ * // comments
+ * var y = 2;
+ * ```
+ * ```js
+ * var x = 1;
+ * // comments
+ * // comments 2
+ * var y = 2;
+ * ```
+ * ```js
+ * const a = 1;
+ * const b = 2;
+ * ```
+ * ```js
+ * var x = 1;
+ * function y() {var z;};
+ * ```
+ * ```js
+ * var x = 1;
+ * var y = function () {var z;};
+ * ```
+ * ```js
+ * var w = 1;
+ * function x() {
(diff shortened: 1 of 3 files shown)


lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+
+                        commentLength += comment.range[1] - comment.range[0] + comment.whitespaceBefore.length;
+                    }
+
+                    if (node.range[0] - whiteSpaceLength - commentLength <= lastNode.range[1]) {
+                        return;
+                    }
+                }
+
+                parentNodeStack.hasNonVarDecl = true;
```
markelog: You modifying the AST?

test/specs/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,94 @@
...
+                { statement: 'var a;var count;while(count<10){var b=1;}const c=1;', errorCount: 2 },
+                { statement: 'var a;var count;do{var b=1;}while(count<10);const c=1;', errorCount: 2 },
+                { statement: 'var a;for(var count=0; count<10;count++){var c;}var d;', errorCount: 3 },
+                { statement: 'var a; for (var count = 0; count < 10; count++) { var c; } var d;', errorCount: 3 },
+                { statement: 'var a;\nfor (var count = 0; count < 10; count++) {\nvar c;\n}\nvar d;', errorCount: 3 }
+            ];
+
+            scenarios.forEach(function(scenario) {
+                describe('statement' + JSON.stringify(scenario.statement), function() {
+                    it('should return ' + scenario.errorCount + ' error' + (scenario.errorCount === 1 ? '' : 's'),
```
markelog: Not really a fan of that kind of declaration, i understand it is much easier to define, but it is much harder to support or isolate.

Not saying you should redo it, but want to hear some opinions about it.

lib/rules/require-var-decl-first.js
markelog: Oh, i see, you don't

markelog:
Could you add some comments, like for `while` `for` loops, or why you need to `break`/`continue`, etc.

Need some else to give "LGTM" since this logic is not obvious like for the other rules.

markelog:
@zxqfox @mrjoelkemp @mikesherov @hzoo i would like to merge this, so it would go in to 2.0, but need a at least one "LGTM" from you

oredi:
@markelog sorry could not look into the comments yet. I will try to put them in before you start your monday. Although while working on this I have to say the rule was more complex then most of what I have work on so far.

markelog:
@oredi which is makes so cool :-)

lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+            var count;
+            var comment;
+            var commentLength;
+            var parentNode;
+            var parentNodeStack;
+            var lastNode;
+            var whiteSpaceLength;
+
+            parentNode = node.parentNode;
+            while (parentNode.type !== 'Program' &&
```
mrjoelkemp: Extract to a function so the name indicates what this does. Not clear why this is necessary.

mrjoelkemp: Ah, i see. Youre trying to find the wrapping function of the node,  if any. Yes, a function like getFunctionContaining(node) would help clarify.

lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+                parentNode.type !== 'FunctionDeclaration' &&
+                parentNode.type !== 'FunctionExpression') {
+                parentNode = parentNode.parentNode;
```
mrjoelkemp: parentNode represents the containing function. It should be called such. This should help make the code clearer.

lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+            }
+            parentNodeStack = parentNodes[parentNode.range[0]];
+
+            if (parentNodeStack.declarations.length === 0) {
+                parentNodeStack.declarations.push(node);
+            } else {
+                lastNode = parentNodeStack.declarations[parentNodeStack.declarations.length - 1];
+                parentNodeStack.declarations.push(node);
+                if (!parentNodeStack.hasNonVarDecl) {
+                    if (node.range[0] <= lastNode.range[1]) {
```
mrjoelkemp: Can you extract to a named variable to explain what's being checked here?

lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+                        return;
+                    }
+
+                    whiteSpaceLength = file.getFirstNodeToken(node).whitespaceBefore.length;
+                    commentLength = 0;
```
mrjoelkemp: For clarity and reducing the amount of code in this callback, please extract the computation of this to a function.

lib/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,179 @@
...
+                        }
+
+                        if (lastNode.range[0] >= comment.range[1]) {
+                            continue;
+                        }
+
+                        commentLength += comment.range[1] - comment.range[0] + comment.whitespaceBefore.length;
+                    }
+
+                    if (node.range[0] - whiteSpaceLength - commentLength <= lastNode.range[1]) {
```
mrjoelkemp: Not clear what this is checking. Extract to a named variable please

test/specs/rules/require-var-decl-first.js
```diff
@@ -0,0 +1,94 @@
...
+
+            it('should report an error for value "false"', function() {
+                assert.throws(function() {
+                    checker.configure({ requireVarDeclFirst: false });
+                });
+            });
+        });
+
+        describe('rule checks', function() {
+            var scenarios = [
```
mrjoelkemp: Would be easier to have a list of valid snippets. We can assume that they'd have an errorCount of zero if they were members.

test/specs/rules/require-var-decl-first.js
mrjoelkemp: Agreed. The test code isn't a description of the behavior being tested. It's that which tests the behavior. The number of errors isn't important when stating which behaviors or preconditions result in a failure.

mrjoelkemp:
Great work here @oredi! Excited for this rule. However, as my comments suggest, I worry about the maintenance of the current state of the code. Any improvements to stating the high level algorithm through function usage is a big win. 

Feel free to ping me for clarification or when the edits are made. Thanks for the great effort here.

oredi:
@markelog  and @mrjoelkemp, I will need some time to look into the changes. I have found some bugs today as well. I will update the PR again when I can get the fixes in and hopefully make the implementation clearer

oredi:
@markelog @mrjoelkemp

I restructured the tests and code, hopefully this version is clearer. Please let me know if further edits are required

markelog:
hm, it will fail for the case like this -

```js
function tst(node) {
    var array, // array of class names
        ncn = node.className;
}
```

markelog:
When it is ready, would like to add to presets - 

Crockford - http://javascript.crockford.com/code.html#variable declarations
"The var statement should be the first statement in the function body." 

Wikimedia - https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Declarations
"Variables must be declared before use and should be the first statement of the function body."

WordPess - https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#declaring-variables-with-var
"Each function should begin with a single comma-delimited var statement that declares any local variables necessary."

/cc @Krinkle, @paulschreiber

jacksonrayhamilton:
Happy to see this rule being implemented. I'll try it out soon.

I'd like to see a test case for (ignoring) the Directive Prologue (i.e., `"use strict"` directives); see [14.1.1](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-directive-prologues-and-the-use-strict-directive).

jacksonrayhamilton:
Also, I think this rule should be called `requireVariableDeclarationFirst`, as "declaration" is spelled correctly in 5 other rules, and `requireMultipleVarDecl` should be renamed to `requireMultipleVariableDeclarations` with the old name as an obsolete alias.

oredi:
@markelog Thanks I missed that scenario, I have put in the fix with the update to the tests.

@jacksonrayhamilton Support for this has been added to the latest commit, now the rule allows the placement of 'use strict' as the first statement of the Program, Function Declaration or Function Expression scope. I am keeping this as a separated commit as it is a special condition. If everyone agrees I combine them into a single commit.

I am keeping the name for now but if the others agree with the name change I will do that.

mrjoelkemp:
Great job here @oredi! The code reads much better now; hopefully you agree. Thanks so much for making the edits.

markelog:
I predict that we will receive a ticket that would ask to allow case like this - 
```js
// Example
var i;
```

markelog:
Thank you @oredi, nicely done

oredi:
@markelog  Thanks and I believe the following code will not cause any issue as the code handles that condition as well. Btw shall I put in another PR to add the rule to the presets for Crockford, Wikimedia and WordPess. And if yes do I need to create an issue for it or just the PR will do?
    // Example
    var i;

@mrjoelkemp Thanks. And the truth is I like this way of coding method. It literally allows me to KISS the code. :)

oredi:
@markelog  sorry looks like you added the presets. Thanks

PR #1526, opened 2015-07-13
New rule: add JSDoc rules

### LINKED ISSUE #986: Execution without options

### LINKED ISSUE #1359: Make jsDoc plugin part of the jscs dist

markelog:
Documentation synchronization is an issue, plus jscs-dev/jscs-dev.github.io/issues/26 jscs-dev/jscs-dev.github.io/issues/25 are also unpleasant.

Fixes #1359

@zxqfox @mdevils thoughts?

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,4 @@
 {
-    "plugins": ["jsdoc"],
     "preset": "google",
     "fileExtensions": [".js", "jscs"],
 
@@ -13,10 +12,6 @@
 
     "safeContextKeyword": "_this",
 
-    "excludeFiles": [
-        "test/data/**"
-    ],
-
     "jsDoc": {
         "checkAnnotations": "closurecompiler",
         "checkParamNames": true,
@@ -28,5 +23,9 @@
         "checkTypes": "capitalizedNativeCase",
         "checkRedundantAccess": true,
         "requireNewlineAfterDescription": true
-    }
+    },
+
+    "excludeFiles": [
+        "test/data/**"
+    ]
 }
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -774,6 +774,9 @@ Configuration.prototype.registerDefaultRules = function() {
         These rules are linked explicitly to keep browser-version supported.
     */
 
+    // Register jsdoc plugin
+    this.registerRule(require('../rules/jsdoc'));
+
     /* ES6 only */
     this.registerRule(require('../rules/require-parentheses-around-arrow-param'));
     this.registerRule(require('../rules/disallow-parentheses-around-arrow-param'));
--- a/lib/rules/jsdoc.js
+++ b/lib/rules/jsdoc.js
@@ -1,0 +1,799 @@
+/**
+ * Validate jsdoc comments
+ *
+ * ## Usage
+ *
+ * ```json
+ * {
+ *     "jsDoc": {
+ *         "checkAnnotations": "closurecompiler",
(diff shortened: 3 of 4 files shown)


zxqfox:
Oh, thanks for that. But, hmmm... ;-) it's dirty! Do we really want to load the only rule and copy-paste docs like that?

What I really want here:
- change file structure a little to have splitted docs and code (also tests maybe?);
- 

File structure could be then like:
```sh
rules/
  rule1/
    index.js
    index.md
  rule2.js # with current structure
```

And we could load them as we do it now but `*.md` files would be easier to load to build and also they will be highlighted, we could check them separately, etc.

And with that structure we have to build as many plugins as we need.

I'm not sure also that we should register `jsDoc` as a single rule. I would like to see it like in eslint plugins: `jsdoc/rule1`. But this issue is like so very complicated to solve that nobody want to open that pandora box. :smirk_cat:

Btw. Why you stick to 1.0.0?

Looks nice at all. If everybody fine with that method — ok.

markelog:
> Do we really want to load the only rule and copy-paste docs like that?

Well, i don't, i really don't like that, but this is currently a one way to do that, we could do hierarchy, but that would mean site would need to support that and we would need to dynamically move it from `node_modules` 

> I'm not sure also that we should register jsDoc as a single rule. I would like to see it like in eslint plugins: jsdoc/rule1. But this issue is like so very complicated to solve that nobody want to open that pandora box.

Would you like to open that? Would be really cool, if you did.

> Btw. Why you stick to 1.0.0?

Because if you release new version, you might change documentation, but this dup will not change :-(

zxqfox:
> Because if you release new version, you might change documentation, but this dup will not change :-(

But the last version is 1.1.0

> Would you like to open that? Would be really cool, if you did.

Would be cool if anyone will do it ;-) Let's listen to guys and @mdevils to decide what to do here.

p.s. I fear this react and webpack used in our site ;-(

markelog:
> p.s. I fear this react and webpack used in our site ;-(

Join the club, but common, this is nothing for bemtree priv.js veteran :-)

zxqfox:
@markelog Nice joke ;-D

PR #1537, opened 2015-07-17
disallowKeywordsOnNewLine: Allow comments before keywords

### LINKED ISSUE #1421: An option to allow keywords on new line if there are comments when disallowKeywordsOnNewLine is true
Valid
```js
if (x < 0) {
    x++;
} else {
    x--;
}

if (x < 0) {
    x++;
} 
// comments
else {
    x--;
}
```

Invalid
```js
if (x < 0) {
    x++;
}
else {
    x--;
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/19255710-an-option-to-allow-keywords-on-new-line-if-there-are-comments-when-disallowkeywordsonnewline-is-true?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oredi:
Allow a special case for the keywords 'else' in 'if else' and 'while' in 'do while' to go on a new line if it is preceded by a comment token

Fixes #1421

### DIFF
--- a/test/specs/rules/disallow-keywords-on-new-line.js
+++ b/test/specs/rules/disallow-keywords-on-new-line.js
@@ -130,6 +130,68 @@ describe('rules/disallow-keywords-on-new-line', function() {
         );
     });
 
+    describe('legal block comments before key word (#1421)', function() {
+        it('should not report legal keyword placement for a "if else"',
+            function() {
+            checker.configure({ disallowKeywordsOnNewLine: ['else'] });
+            assert(
+                checker.checkString(
+                    'if (x) {\n' +
+                        'x++;\n' +
+                    '}\n' +
+                    '/* comments */\n' +
+                    'else {\n' +
+                        'x--;\n' +
+                    '}'
+                ).isEmpty()
+            );
+        });
+
+        it('should not report legal keyword for a "do while"', function() {
+            checker.configure({ disallowKeywordsOnNewLine: ['while'] });
+            assert(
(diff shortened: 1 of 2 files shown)


test/specs/rules/disallow-keywords-on-new-line.js
```diff
@@ -130,6 +130,37 @@ describe('rules/disallow-keywords-on-new-line', function() {
         );
     });
 
+    describe('legal comments before key word (#1421)', function() {
+        it('should not report legal keyword placement for a "if else"',
+            function() {
```
mrjoelkemp: this was done for minimumLineLength? Would like this on the previous line

mrjoelkemp:
Nice work here @oredi! 

Can we have a test for a single line comment? Still not seeing why the regex was used – a test would help clarify the comment.

oredi:
@mrjoelkemp I added the tests for Single comment and removed the regex. I was having issues initially with single line comments as I did not place a line break after the comment. So that commented out the test code after the single line comment. After looking into the other rules for comments, I think I got it right now.

mrjoelkemp:
lgtm. nice work!

markelog:
Thank you!

PR #1540, opened 2015-07-19
Turn on es5 by default (and turn off es3)

zxqfox:
/cc @markelog @hzoo @mikesherov

### DIFF
--- a/lib/rules/require-curly-braces.js
+++ b/lib/rules/require-curly-braces.js
@@ -89,19 +89,19 @@ module.exports.prototype = {
 
         var typeIndex = this._typeIndex;
 
-        if (typeIndex['if'] || typeIndex['else']) {
+        if (typeIndex.if || typeIndex.else) {
             file.iterateNodesByType('IfStatement', function(node) {
-                if (typeIndex['if'] && isNotABlockStatement(node.consequent)) {
+                if (typeIndex.if && isNotABlockStatement(node.consequent)) {
                     addError('If', node);
                 }
-                if (typeIndex['else'] && isNotABlockStatement(node.alternate) &&
+                if (typeIndex.else && isNotABlockStatement(node.alternate) &&
                     node.alternate.type !== 'IfStatement') {
                     addError('Else', file.getPrevToken(file.getFirstNodeToken(node.alternate)));
                 }
             });
         }
 
-        if (typeIndex['case'] || typeIndex['default']) {
+        if (typeIndex.case || typeIndex.default) {
             file.iterateNodesByType('SwitchCase', function(node) {
                 // empty case statement
                 if (node.consequent.length === 0) {
@@ -112,30 +112,30 @@ module.exports.prototype = {
                     return;
                 }
 
-                if (node.test === null && typeIndex['default']) {
+                if (node.test === null && typeIndex.default) {
                     addError('Default', node);
                 }
 
-                if (node.test !== null && typeIndex['case']) {
+                if (node.test !== null && typeIndex.case) {
                     addError('Case', node);
                 }
             });
         }
 
-        if (typeIndex['while']) {
+        if (typeIndex.while) {
             checkBody('WhileStatement', 'While');
         }
 
-        if (typeIndex['for']) {
+        if (typeIndex.for) {
             checkBody('ForStatement', 'For');
             checkBody('ForInStatement', 'For in');
         }
 
-        if (typeIndex['do']) {
+        if (typeIndex.do) {
             checkBody('DoWhileStatement', 'Do while');
         }
 
-        if (typeIndex['with']) {
+        if (typeIndex.with) {
             checkBody('WithStatement', 'With');
         }
     }
--- a/lib/rules/require-dot-notation.js
+++ b/lib/rules/require-dot-notation.js
@@ -1,11 +1,17 @@
 /**
  * Requires member expressions to use dot notation when possible
  *
- * Types: `Boolean` or `String`
+ * Types: `Boolean` or `Object`
  *
  * Values:
  *  - `true`
- *  - `"except_snake_case"` allow quoted snake cased identifiers
+ *  - `"except_snake_case"` (*deprecated* use `"allExcept": ["snake_case"]`) allow quoted snake cased identifiers
+ *  - `Object`:
+ *    - `'allExcept'` array of exceptions:
+ *       - `'keywords'` allow quoted identifiers made of reserved words
+ *       - `'snake_case'` allow quoted snake cased identifiers
+ *
+ * N.B.: keywords are always allowed with es3 enabled (http://jscs.info/overview.html#es3)
  *
  * JSHint: [`sub`](http://www.jshint.com/docs/options/#sub)
  *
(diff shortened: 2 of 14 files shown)


lib/rules/require-dot-notation.js
```diff
@@ -62,30 +99,50 @@
...
     configure: function(options) {
+        if (typeof options !== 'object') {
+            assert(
+                options === true || options === 'except_snake_case',
+                this.getOptionName() + ' option requires either a true value or an object'
+            );
+
+            var _options = {};
+            if (options === 'except_snake_case') {
+                console.warn('Value "except_snake_case" was deprecated\n' +
```
mikesherov: This should be an error not a warm considering this is2.0

markelog: I think we should remove warn completely, but when we start warn about such things, we should do it in a general way, so we wouldn't need to add deprecation noticed in every rule.

Just have file with the old way and the new way. @mdevils already working on the mapping file.

Besides, never use `console` in rules, since such errors are not consumable through reporters.

presets/wordpress.json
```diff
@@ -1,5 +1,8 @@
 {
     "preset": "jquery",
+    "requireDotNotation": {
+        "allExcept": ["snake_case", "keywords"]
```
markelog: It is better to make such changes more visible i.e. have special commit and link on why this was done

lib/reserved-words.js
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Check word for being an reserved word.
+ *
+ * @param {String} word - word to check
+ * @param {String|Number} [dialect] - dialect or version
+ * @param {Boolean} [strict] - strict mode
+ * @returns {?Boolean}
+ */
+exports.check = function check(word, dialect, strict) {
```
markelog: This looks pretty great, i would even move this to standalone module, since it might be useful for other projects as well.
It seems though `esutils` module already does this job, although with different interface.

lib/rules/require-curly-braces.js
```diff
@@ -112,30 +112,30 @@ module.exports.prototype = {
                     return;
                 }
 
-                if (node.test === null && typeIndex['default']) {
+                if (node.test === null && typeIndex.default) {
                     addError('Default', node);
                 }
 
-                if (node.test !== null && typeIndex['case']) {
+                if (node.test !== null && typeIndex.case) {
```
markelog: So this code will break in more old browsers?

lib/rules/require-dot-notation.js
```diff
@@ -1,11 +1,16 @@
...
  *
  * Values:
  *  - `true`
- *  - `"except_snake_case"` allow quoted snake cased identifiers
+ *  - `Object`:
+ *    - `'allExcept'` array of exceptions:
+ *       - `'keywords'` allow quoted identifiers made of reserved words
+ *       - `'snake_case'` allow quoted snake cased identifiers
+ *
+ * N.B.: keywords are always allowed with es3 enabled (see option es3)
```
markelog: This could be a link

lib/rules/require-dot-notation.js
```diff
@@ -1,11 +1,16 @@
 /**
  * Requires member expressions to use dot notation when possible
  *
- * Types: `Boolean` or `String`
+ * Types: `Boolean` or `Object`
  *
  * Values:
  *  - `true`
- *  - `"except_snake_case"` allow quoted snake cased identifiers
```
markelog: I think it would preferable do changes like that in the 3.0 version

lib/rules/require-dot-notation.js
```diff
@@ -15,17 +20,32 @@
...
+ * var a = b['c'];
+ * var a = b['snake_cased'];
+ * var a = b['_camelCased'];
+ * var a = b['camelCased_'];
+ * ```
+ *
+ * #### Example for allExcept snake_case
+ *
+ * ```js
+ * "requireDotNotation": { "allExcept": "snake_case" }
```
markelog: This should be a `"allExcept": ["snake_case" ]`?

lib/rules/require-dot-notation.js
```diff
@@ -33,26 +53,43 @@
...
+ *
+ * ```
+ * var a = b['await']; // reserved word in ES6
+ * ```
+ *
+ * #### Example for allExcept keywords with esnext
  *
  * ```js
- * var a = b['c'];
+ * "requireDotNotation": { "allExcept": "keywords" }
```
markelog: Same as above

lib/rules/require-dot-notation.js
zxqfox: Let's drop it then. I doubted about this if. Now I see we shouldn't add it.

presets/wordpress.json
zxqfox: I'll move it to separate commit.

lib/rules/require-dot-notation.js
zxqfox: Yep. My bad.

lib/rules/require-dot-notation.js
zxqfox: Can you elaborate?

lib/rules/require-curly-braces.js
zxqfox: IE8 and older doesn't support it. Yep ;-\
Should we set `{es3: true}` in our `.jscsrc`?

lib/rules/require-dot-notation.js
markelog: Even though a lot of rule values are deprecated, we do not removing them, leaving this task for 3.0

markelog: This wasn't explicitly said anywhere, it just where the flow of development is

lib/rules/require-curly-braces.js
markelog: And in Android 2.3, i'd say we don't care about them

lib/reserved-words.js
zxqfox: Well, looks like would be nice to use it. But there are no ES3 support atm.

markelog: More reasons to move `reserved-words` to standalone module :-)

lib/rules/require-dot-notation.js
zxqfox: Ah, true. Thanks! I'll add here deprecation note.

lib/reserved-words.js
zxqfox: Moved https://github.com/zxqfox/reserved-words/blob/master/test/reserved-words.js

I suggest to move it to jscs-dev namespace ;-)

zxqfox:
Need one more review here.

lib/reserved-words.js
markelog: Will you use this package in this pull?

markelog: I guess we can do that later, after squashing couple commits - LGTM

zxqfox: I was sure I already used this package here ;-\

zxqfox: Shame on me ;-( I've forgot to push...

PR #1542, opened 2015-07-19
Integrate CST: Work in progress

mdevils:
Continuing CST **Work In Progress** pull request. Want to keep you informed about the current integration I'm working on.

In this pull request:

* Use CST instead of AST with token hacks (https://github.com/mdevils/cst)
* Simplify JsFile and rules.
* Improve assertions in tests using chai and chai extensions (it is very hard to make huge changes with the current error messages).

Feel free to discuss changes. Any help is very appreciated.

I try to keep everything rebased.

/cc @markelog @hzoo @mrjoelkemp @mikesherov @zxqfox

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -1,4 +1,7 @@
 var treeIterator = require('./tree-iterator');
+var cst = require('cst');
+var Parser = cst.Parser;
+var Program = cst.types.Program;
 
 /**
  * Operator list which are represented as keywords in token list.
@@ -25,7 +28,6 @@ var JsFile = function(params) {
     this._parseErrors = [];
     this._filename = params.filename;
     this._source = params.source;
-    this._tree = {tokens: [], comments: []};
 
     this._es3 = params.es3 || false;
     this._es6 = params.es6 || false;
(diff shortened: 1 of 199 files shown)


lib/js-file.js
```diff
@@ -110,24 +123,25 @@ JsFile.prototype = {
...
-                return true;
-            }
-
-            if (region.rule === ruleName || region.rule === '*') {
-                enabled = region.enabled;
-            }
-        }, this);
-
-        return enabled;
+    isEnabledRule: function(ruleName, element) {
```
markelog: Are you planning to rework these methods as well? Seems unrelated to proposed changes?

lib/js-file.js
```diff
@@ -415,11 +408,11 @@ JsFile.prototype = {
      */
     getNodesByType: function(type) {
         if (typeof type === 'string') {
-            return this._index[type] || [];
+            return this._program.selectNodesByType(type);
         } else {
             var result = [];
             for (var i = 0, l = type.length; i < l; i++) {
-                var nodes = this._index[type[i]];
+                var nodes = this._program.selectNodesByType(type[i]);
```
markelog: Looks much better, although it seems there is no need for the `for` anymore, some of the `Array#<methods>` or `lodash` methods should do the job?

lib/js-file.js
```diff
@@ -447,17 +440,18 @@ JsFile.prototype = {
...
-        var typeIndex = {};
-        types.forEach(function(type) {
-            typeIndex[type] = true;
-        });
-
-        this.getTokens().forEach(function(token, index, tokens) {
-            if (typeIndex[token.type]) {
-                cb(token, index, tokens);
+        var tokens;
+        if (Array.isArray(type)) {
```
markelog: It seems argument check should look like the same way as in `getNodesByType` method

lib/js-file.js
```diff
@@ -581,7 +575,15 @@ JsFile.prototype = {
      * Returns comment token list, built using esprima.
      */
     getComments: function() {
-        return this._tree.comments;
+        var comments = [];
+        var token = this._program.firstToken;
```
markelog: Maybe method like that should be included in cst?

lib/js-file.js
```diff
@@ -670,152 +649,6 @@ JsFile.prototype = {
...
-    },
-
-    /**
-     * Applies whitespace information to the token list.
-     *
-     * @param {Object[]} tokens
-     * @param {String} source
-     * @private
-     */
-    _applyWhitespaceData: function(tokens, source) {
```
markelog: Was this method being replaced?

test/chai-extensions.js
```diff
@@ -0,0 +1,145 @@
+var chai = require('chai');
+var utils = chai.utils;
+
+chai.use(function(chai, utils) {
+    /**
+     * Error assertion for `Errors` instances.
+     */
+    chai.Assertion.addChainableMethod('error', function(message) {
```
markelog: Would document those extensions at the wiki page, otherwise this could get confusing

markelog:
Since changes are pretty big and will get bigger, i think we need to make this green for the travis - add all those `skip`'s or disable testing for the yet unmodified rules. And land it to `3.0` otherwise diff would became pretty big so it would be impossible to review properly.

lib/js-file.js
mdevils: This is a very frequently called function so I decided to have a fast `for`.

lib/js-file.js
mdevils: Yep, probably yes.

lib/js-file.js
mdevils: Yes, whitespace are tokens now, so no `whitespaceBefore` anymore.

test/chai-extensions.js
mdevils: +1

mdevils:
@markelog OK, I will make it green soon.

lib/js-file.js
mdevils: I have to rewrite it somehow, since comments don't have fixed positions anymore.

hzoo:
Babel moved comment attachment to the parser instead of estraverse to allow for comments in empty blocks - check out the recent babel-eslint changes

PR #1543, opened 2015-07-19
disallowSpacesInAnonymousFunctionExpression: Remove excess assert

aivus:
I want to override preset settings, but can't, because of this assert.

Google pereset contain [disallow rule](https://github.com/jscs-dev/node-jscs/blob/master/presets/google.json#L54), and I can't disable `beforeOpeningRoundBrace` rule by removing it (because assert doesn't allow empty rules)

### DIFF
--- a/lib/rules/disallow-spaces-in-anonymous-function-expression.js
+++ b/lib/rules/disallow-spaces-in-anonymous-function-expression.js
@@ -63,12 +63,6 @@ module.exports.prototype = {
             );
         }
 
-        assert(
-            options.beforeOpeningCurlyBrace || options.beforeOpeningRoundBrace,
-            this.getOptionName() + ' must have beforeOpeningCurlyBrace ' +
-            ' or beforeOpeningRoundBrace property'
-        );
-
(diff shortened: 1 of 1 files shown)


aivus:
Also this asset [exists in other rules too](https://github.com/jscs-dev/node-jscs/search?utf8=%E2%9C%93&q=must+have+beforeOpeningCurlyBrace&type=Code).

mrjoelkemp:
Hi @aivus. That's a necessary part of the rule to validate its configuration values. To solve your problem, you can set the rule's value to null in your config. 

Specifically, set the preset in your config, then state the rule you want to override and then assign its value to null.

hzoo:
Yeah in the docs - http://jscs.info/overview.html#presets.

This seems to come up a lot - maybe we can change the error message if someone puts false to ask them to put null or change the assert messages?

mrjoelkemp:
Or we could just support both `null` and `false` for turning off a rule. @markelog

hzoo:
Would be good since it seems everyone wants to do it!

markelog:
Don't remember why we didn't do it before, worth a shot though!

PR #1545, opened 2015-07-20
New Rule: requireSortedObjectKeys

ruimarinho:
Requires sorted object keys.

Types: `Boolean` or `String`

Values:
  - `true` (alias to `asc`)
  - `"asc"`: requires sorting in ascending order
  - `"desc"`: requires sorting in descending order

#### Example

```js
"requireSortedObjectKeys": "asc"
```

##### Valid

```js
var x = {
  x: 'foo',
  y: 'bar'
}
```

##### Invalid

```js
var x = {
  y: 'foo',
  x: 'bar'
}
```

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -953,6 +953,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-space-before-comma'));
 
     this.registerRule(require('../rules/require-space-before-comma'));
+
+    this.registerRule(require('../rules/validate-order-in-object-keys'));
 };
 
 /**
--- a/lib/rules/validate-order-in-object-keys.js
+++ b/lib/rules/validate-order-in-object-keys.js
@@ -1,0 +1,153 @@
+
+/**
+ * Validates the order in object keys.
+ *
+ * Types: `Boolean` or `String`
+ *
+ * Values:
+ *  - `true` (alias to `asc`)
+ *  - `"asc"`: requires sorting in ascending order
+ *  - `"asc-insensitive"`: requires sorting in ascending order (case-insensitive)
+ *  - `"asc-natural"`: requires sorting in ascending natural order
+ *  - `"desc"`: requires sorting in descending order
+ *  - `"desc-insensitive"`: requires sorting in descending order (case-insensitive)
+ *  - `"desc-natural"`: requires sorting in descending natural order
+ *
+ * #### Example
+ *
+ * ```js
+ * "validateOrderInObjectKeys": "asc"
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var x = {
+ *  x: 'foo',
+ *  y: 'bar'
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var x = {
+ *  y: 'foo',
+ *  x: 'bar'
+ * }
+ * ```
+ */
+
+var assert = require('assert');
+var naturalSort = require('natural-compare');
+
+/**
+ * Sort in ascending order.
+ */
+function asc(a, b) {
+    return String(a) < String(b) ? -1 : 1;
+}
+
+/**
+ * Sort in ascending order (case-insensitive).
+ */
+function ascInsensitive(a, b) {
+    var lowercaseA = String(a).toLowerCase();
+    var lowercaseB = String(b).toLowerCase();
+
+    if (lowercaseA < lowercaseB) {
+        return -1;
+    }
+
+    if (lowercaseA > lowercaseB) {
+        return 1;
+    }
+
+    return asc(a, b);
+}
+
+/**
+ * Natural sort in ascending order.
+ */
+function ascNatural(a, b) {
+    return naturalSort(a, b);
+}
+
+/**
+ * Native sort in descending order.
+ */
+function desc(a, b) {
+    return asc(a, b) * -1;
+}
+
+/**
+ * Sort in descending order (case-insensitive).
+ */
+function descInsensitive(a, b) {
+    return ascInsensitive(a, b) * -1;
+}
+
+/**
+ * Native sort in descending order.
+ */
+function descNatural(a, b) {
+    return naturalSort(a, b) * -1;
+}
+
+/**
+ * Available sort methods.
+ */
+var methods = {
+    asc: asc,
+    'asc-insensitive': ascInsensitive,
+    'asc-natural': ascNatural,
+    desc: desc,
+    'desc-insensitive': descInsensitive,
+    'desc-natural': descNatural
+};
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true || Object.keys(methods).indexOf(options) !== -1,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+
+        this._sort = methods[options] || methods.asc;
+    },
+
+    getOptionName: function() {
+        return 'validateOrderInObjectKeys';
+    },
+
+    check: function(file, errors) {
+        var sort = this._sort;
+
+        file.iterateNodesByType('ObjectExpression', function(node) {
+            var keys = node.properties.map(function(property) {
+                return (property.key.name || property.key.value);
+            });
+
+            var sorted = keys.slice(0).sort(sort);
+            var unsorted;
+
+            for (var i = 0; i < keys.length; i++) {
+                if (keys[i] !== sorted[i]) {
+                    unsorted = i;
+
+                    break;
+                }
+            }
+
+            if (undefined !== unsorted) {
+                errors.add(
+                    'Object keys must be in ' + (/asc/.test(sort.name) ? 'ascending' : 'descending') + ' order',
+                    node.properties[unsorted].loc.start
+                );
+            }
+        });
+    }
+};
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "jscs-jsdoc": "1.1.0",
     "lodash.assign": "~3.2.0",
     "minimatch": "~2.0.1",
+    "natural-compare": "~1.2.2",
     "pathval": "~0.1.1",
     "prompt": "~0.2.14",
     "reserved-words": "^0.1.1",
--- a/test/specs/rules/validate-order-in-object-keys.js
+++ b/test/specs/rules/validate-order-in-object-keys.js
@@ -1,0 +1,193 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/validate-order-in-object-keys', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    describe('should alias option value true to asc', function() {
+        beforeEach(function() {
+            checker.configure({ validateOrderInObjectKeys: 'asc' });
+        });
+
+        it('should report unsorted object keys', function() {
+            assert(checker.checkString('var obj = {\na:1,\n_:2,\nb:3\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\na:1,\nc:2,\nb:3\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\nb_:1,\na:2,\nb:3\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\nb_:1,\nc:2,\nC:3\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\n$:1,\n_:2,\nA:3,\na:4\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\n1:1,\n2:4,\nA:3,\n\'11\':2\n}').getErrorCount() === 1);
+            assert(checker.checkString('var obj = {\n\'#\':1,\nÀ:3,\n\'Z\':2,\nè:4\n}').getErrorCount() === 1);
+        });
+
+        it('should not report sorted object keys', function() {
+            assert(checker.checkString('var obj = {\n_:2,\na:1,\nb:3\n}').isEmpty());
+            assert(checker.checkString('var obj = {\na:1,\nb:3,\nc:2\n}').isEmpty());
+            assert(checker.checkString('var obj = {\na:2,\nb:3,\nb_:1\n}').isEmpty());
+            assert(checker.checkString('var obj = {\nC:3,\nb_:1,\nc:2\n}').isEmpty());
+            assert(checker.checkString('var obj = {\n$:1,\nA:3,\n_:2,\na:4\n}').isEmpty());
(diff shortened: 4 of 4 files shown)


zxqfox:
Since there is no sense to have disallowSortedObjectKeys. Maybe we can rename this rule to validateOrderInObjectKeys or something?

ruimarinho:
Sounds good to me.

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true || options === 'asc' || options === 'desc',
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+
+        this.sort = options === true ? 'asc' : options;
```
zxqfox: Better to prefix this with `_` .

test/specs/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,49 @@
...
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+    });
+
+    it('should alias option value true to asc', function() {
+        checker.configure({ requireSortedObjectKeys: true });
+
+        assert(checker.getConfiguration().getConfiguredRule('requireSortedObjectKeys').sort === 'asc');
```
mrjoelkemp: This reaches into a rule's implementation. Would prefer to assert the implied sort order with a test like the other values (i.e., testing that it reports on descending keys but not on ascending)

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+
+    if (a < b) {
+        return -1;
+    }
+
+    if (a > b) {
+        return 1;
+    }
+
+    return 0;
```
mrjoelkemp: You could remove this since objects don't have duplicate keys.

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+
+    if (a > b) {
+        return -1;
+    }
+
+    if (a < b) {
+        return 1;
+    }
+
+    return 0;
```
mrjoelkemp: You could remove this since objects don't have duplicate keys.

mrjoelkemp:
Nice work here @ruimarinho! Minor issues in the PR, but a solid contribution. Excited for this rule.

lib/rules/require-sorted-object-keys.js
zxqfox: I'd say `return a < b ? -1 : 1` would be fine.

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+ * ```js
+ * var x = {
+ *  y: 'foo',
+ *  x: 'bar'
+ * }
+ * ```
+ */
+
+function asc(a, b) {
+    if (/[_]/.test(a)) {
```
zxqfox: This can return false positive for snake cased keys like `snake_case`.
E.g. this object will catch that bug:
```
{
  b_: 'asd',
  a: 'qwe'
}
```

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    if (a > b) {
+        return 1;
+    }
+
+    return 0;
+}
+
+function desc(a, b) {
```
zxqfox: Since there are just two options (asc and desc) why not use `asc` in `desc`? Like `return -asc(a);`

lib/rules/require-sorted-object-keys.js
```diff
@@ -0,0 +1,118 @@
...
+
+    check: function(file, errors) {
+        var sort = this.sort;
+
+        file.iterateNodesByType('ObjectExpression', function(node) {
+            var keys = node.properties.map(function(property) {
+                return property.key.name;
+            });
+
+            var sorted = keys.slice(0).sort(sort === 'asc' ? asc : desc);
```
zxqfox: Just thought: If there will be many large objects in project this can affect performance.

lib/rules/require-sorted-object-keys.js
ruimarinho: Done, thanks.

test/specs/rules/require-sorted-object-keys.js
ruimarinho: Right, I understand that this layer (parsed configuration rules) may be too *far away* from this test. I wanted to avoid code repetition, but if you feel it's more inline with the general test model, it's fine by me.

lib/rules/require-sorted-object-keys.js
ruimarinho: Nice catch! Thank you.

lib/rules/require-sorted-object-keys.js
ruimarinho: Yep, removed, thanks.

lib/rules/require-sorted-object-keys.js
ruimarinho: The reason why a custom sort is needed (instead of `[].sort()` + `[].sort().reverse()`) is that underscores are considered words. This means that this object would be valid:

```js
var obj = {
  A: 'foo',
  _: 'qux'
  a: 'bar'
}
```

...since 65 (`A`) < 95 (`_`) < 97 (`a`). Stylistically, I don't think this is interesting, as symbols are typically sorted at the top or bottom. I believe this opinionated change is acceptable since in future anyone can create other sort functions if they don't find the current one fitting (e.g. `asc-native`). What are your thoughts?

lib/rules/require-sorted-object-keys.js
ruimarinho: The best optimisation I see for now (but in my opinion in a different phase), would be to compare each property key name with the next one and immediately break out of the loop if a non-expected key is found instead of sorting all keys. I've thought about it before submitting the PR but I wasn't particularly enjoying the implementation I had.

ruimarinho:
@mrjoelkemp thank you :) happy to contribute!

lib/config/configuration.js
```diff
@@ -949,6 +949,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-matching-function-name'));
 
     this.registerRule(require('../rules/disallow-space-before-semicolon'));
+
+    this.registerRule(require('../rules/validate-order-in-object-keys'));
```
ruimarinho: @mrjoelkemp what about sorting these `registerRule` calls? I had some difficulty figuring out if I should just add the new rule to the bottom.

lib/rules/validate-order-in-object-keys.js
```diff
@@ -0,0 +1,102 @@
...
+
+        this._sort = options === true ? 'asc' : options;
+    },
+
+    getOptionName: function() {
+        return 'validateOrderInObjectKeys';
+    },
+
+    check: function(file, errors) {
+        var sort = this._sort;
```
ruimarinho: When will ES6 come to core so we can remove this 😇?

lib/config/configuration.js
hzoo: Yeah right now adding it to the end is fine - maybe someone can do a separate PR to sort it at some point.

lib/rules/validate-order-in-object-keys.js
```diff
@@ -0,0 +1,102 @@
...
+ * ```js
+ * var x = {
+ *  y: 'foo',
+ *  x: 'bar'
+ * }
+ * ```
+ */
+
+function asc(a, b) {
+    if (/^[_]/.test(a)) {
```
jacksonrayhamilton: There shouldn't be special cases for underscores. Some people use this as a convention for signifying "private" properties, but some people don't do that. Some people use e.g. "$" and "$$" (angular), others suffix with underscores (node, `util.inherits`). Others don't believe that publicly-accessible properties should even be treated as "private" (because it's not enforced at the language level).

You could instead add a special option for something like this, such as a way for users to specify their own regexp.

jacksonrayhamilton:
I would be nice if there was a "case-insensitive" option for this.

ruimarinho:
@jacksonrayhamilton regarding your inline comment about underscores, the only thing that the custom sort does is to group what we usually consider symbols (`$`, `_`) together. Underscores are the exception because they are considered words. 

Given the following object:

```js
var obj = {
  A:1,
  a:2,
  _:3,
  $:4
}
```

Is it expected to be considered sorted when it looks like this:

```js
var obj = {
  A:1,
  _:3,
  $:4,
  a:2
}
```

Or this:

```js
var obj = {
  _:3,
  $:4,
  A:1,
  a:2
}
```

Personally, I think the last version is the natural one, but I'm ok with keeping the default sort algorithm if we agree it's the best option :)

jacksonrayhamilton:
I see, char codes 91-96 are sandwiched between the upper and lowercase letters. The 58-64 symbols are also kinda "just there". Maybe both of those ranges should be shifted down or up or something.

jacksonrayhamilton:
Of course, reorganizing ASCII could make this rule more confusing to users. In fact, enforcing this for anything other than the latin alphabet would confuse people. Who wants to get an error for putting a ">" key before his "<" key in his AST parser?

You may want to just exclude any symbols from the sort order checking, or trim them off the front before comparing latin letters.

ruimarinho:
The problem is that it enters the field of natural sorting in js. I'd rather use something like https://github.com/litejs/natural-compare-lite then and add `natural-asc` and `natural-desc`. Would it be ok to add this package or equivalent?

ruimarinho:
Added several new sort methods - it's now possible to choose between native, case-insensitive, locale and natural sorting.

mrjoelkemp:
That new library that you imported might be affecting other tests according travis? Does that library mutate a native object?

mrjoelkemp:
Actually, I'm wrong. A completely separate PR is also failing. There might be a dependency bump that is causing builds to break.

mrjoelkemp:
Seems like babel v5.8.3 is causing duplicated comments. 

@hzoo any chance you could see if the latest babel master passes the test suite? I had a hard time getting a git url download and/or an npm link of a local copy of babel to work well.

hzoo:
Will check it out tomorrow - babel changed how comments are attached (through parser instead of estraverse) so will have to change babel-jscs like in babel-eslint

hzoo:
Ok try again - babel-jscs@2.0.1 - if not will to try fix in the morning

ruimarinho:
@mrjoelkemp @hzoo thanks for the input - `babel-jscs` was indeed the culprit, but just found that `localeCompare` is broken on some versions of node. For that reason, unless we write specific tests for those versions, then it's best to leave `(asc|desc)-locale` sorting out for now.

mrjoelkemp:
:+1:

lib/rules/validate-order-in-object-keys.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    return asc(a, b);
+}
+
+/**
+ * Natural sort in ascending order.
+ */
+function ascNatural(a, b) {
+    return String.naturalCompare(a, b);
```
markelog: So this module implicitly extends `String` prototype, extending and doing so implicitly could be considered as harmful,

Is there any alternative to this package?

ruimarinho: Indeed, I don't understand why the author chose to extend the native String object like this. The alternatives aren't that great and even this solution has its own set of "bugs". See https://github.com/javve/natural-sort/issues/7#issuecomment-57134825 for context. Any help here would be greatly appreciated!

markelog: @hzoo, @mrjoelkemp, @zxqfox any ideas?

mrjoelkemp: If logically equivalent, I'd rather use https://github.com/javve/natural-sort to avoid the String.prototype modification. 

We should let the community indicate the edge cases to consider. Trying to handle all of the cases would lead to a bloated and overly complicated solution likely with a blend of tools used.

markelog: @ruimarinho could you try `natural-sort` package instead?

ruimarinho: @markelog unless I'm missing something, it looks like the `natural-sort` npm package is frontend only (https://github.com/studio-b12/natural-sort/blob/master/dist/natural-sort.js). The sort function is not exported.

If you meant https://github.com/javve/natural-sort, unfortunately it's not available on npm (only component). I've given https://www.npmjs.com/package/natural-compare a go and it seems to be exactly like the original module but without the native prototype intrusion. It doesn't look like it's actively maintained, but time will tell. Rebased with master.

mrjoelkemp:
Great find with the new plugin. LGTM

markelog:
> it's not available on npm (only component). I've given https://www.npmjs.com/package/natural-compare a go and it seems to be exactly like the original module but without the native prototype intrusion. It doesn't look like it's actively maintained, but time will tell. 

Let's ask, @megawac could help us out? Can we relay on this module?

megawac:
Hi @markelog et al,

Last time I needed [`naturalCompare`](https://github.com/epeli/underscore.string/blob/master/naturalCmp.js) or [`levenshtein`](https://github.com/epeli/underscore.string/blob/master/levenshtein.js) I reached for [`underscore.string`](http://epeli.github.io/underscore.string/)'s implementation. I would recommend those implementations over `natural-compare`.

megawac:
And yeah, the package on npm is fine I think I just updated it to be cjs and not install itself on `String`, but I'd have to check. I'll throw up the fork on a github repo tomorrow

ruimarinho:
@megawac just took `underscore.string` for a spin and found some issues with its [implementation](https://github.com/epeli/underscore.string/blob/master/naturalCmp.js) - for instance, an object with keys `[ '$', 'A', '_', 'a' ]` would be sorted to `[ '$', 'A', '_', 'a' ]` instead of `[ '$', '_', 'A', 'a' ]`. Obviously this is a subjective one, but the latter seems more natural to me.

Shall we stay with `natural-compare`?

markelog:
Thank you @megawac.

@ruimarinho i'd say you could submit this issue to the `underscore.string` repo, i guess `underscore.string` is much less risky then `natural-compare`.

megawac:
Yeah, I don't think underscore.string will object to that change. We've
made some similar edge case changes to naturalCmp recently

/cc @stoeffel <https://github.com/stoeffel>

stoeffel:
@magawac no objections to that. Sounds reasonable, that "symbols" come first.

markelog:
@ruimarinho we probably gonna release soon, would you have time to finish this up?

ruimarinho:
@markelog the current implementation of `underscore.string` is based on [Google Closure's Library](http://google.github.io/closure-library/api/source/closure/goog/string/string.js.src.html#l409) and doesn't seem to be a good fit for symbols. Would it be ok to release with the current package and update it when/if we can get underscore.string's implementation improved later on?

markelog:
Okay, created an issue for it - https://github.com/jscs-dev/node-jscs/issues/1647. Let's do this as is

markelog:
Landed, would be cool to autofix it though :-)

PR #1546, opened 2015-07-20
requirePaddingNewlinesAfterBlocks: arrays act like objects for the last item

christophehurpeau:
And added the airbnb check for the section [18.6](https://github.com/airbnb/javascript#18.6)

```js
const arr = [
  function foo() {
  },

  function bar() {
  },
];

return arr;
```

### DIFF
--- a/lib/rules/require-padding-newlines-after-blocks.js
+++ b/lib/rules/require-padding-newlines-after-blocks.js
@@ -46,13 +46,14 @@
  *     func(
  *          function() {
  *          }
- *
  *     );
  *
  *     var a = [
  *         function() {
- *         }
+ *         },
  *
+ *         function() {
+ *         }
  *     ]
  *
  * }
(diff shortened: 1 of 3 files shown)


markelog:
Why? Could you use `inArrayExpressions` option?

christophehurpeau:
`inArrayExpressions` means anywhere in the array.
This require newlines between functions inside the array except the last item, the same behavior than for objects

cyrilf:
:+1: seems consistent with the `objects` rules indeed

christophehurpeau:
up ?

markelog:
Sounds reasonable, could you add the same logic for function call parentheses?

christophehurpeau:
added for call parentheses :)

christophehurpeau:
up ?

markelog:
Thank you

PR #1547, opened 2015-07-20
CLI: set "maxErrors" to Infinity with enabled "fix" option

markelog:


mrjoelkemp:
lgtm minus the test fail

markelog:
Weird, some prs failed, but now, without changing single thing they passing, hm

mrjoelkemp:
babel-jscs worked around some breaking changes in babel that were causing
the failures. The fixes were released in patch versions which travis picked
up fine.

sindresorhus:
Shouldn't this be in the API and not just the CLI? So every consumer don't have to reimplement it. Re: https://github.com/jscs-dev/gulp-jscs/pull/93#issuecomment-144929578

markelog:
Probably, yeah, at this weekend i was planning to refactor `grunt-jscs` module, to have hands on experience on how we should do it

PR #1548, opened 2015-07-20
Preset:CLI: set "airbnb" as default preset

### LINKED ISSUE #986: Execution without options

markelog:
Fixes #986 

According to our traffic stat and stars of that repo this is a most popular javascript code style out there.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -49,13 +49,12 @@ module.exports = function(program) {
     if (!config && !program.preset && !program.autoConfigure) {
         if (program.config) {
             console.error('Configuration source', program.config, 'was not found.');
-        } else {
-            console.error('No configuration found. Add a .jscsrc file to your project root or use the -c option.');
+            defer.reject(4);
+
+            return returnArgs;
         }
 
-        defer.reject(4);
-
-        return returnArgs;
+        program.preset = 'airbnb';
     }
 
     if (!args.length && process.stdin.isTTY && typeof program.autoConfigure !== 'string') {
--- a/test/specs/cli.js
+++ b/test/specs/cli.js
@@ -83,24 +83,19 @@ describe('cli', function() {
         });
     });
 
-    it('should return error if default config was not found', function() {
+    it('should set airbnb preset if there is no config (#986)', function() {
         sinon.spy(console, 'error');
         sinon.stub(configFile, 'load', function() {
             return undefined;
         });
 
-        var result = cli({
-            args: []
-        });
-        var text = 'No configuration found. Add a .jscsrc file to your' +
-            ' project root or use the -c option.';
+        var checker = cli({
+            args: ['test/data/cli/error.js']
+        }).checker;
 
-        return result.promise.always(function() {
-            assert.equal(console.error.getCall(0).args[0], text);
-
-            configFile.load.restore();
-            console.error.restore();
-        });
+        assert.equal(checker._configuration.getPresetName(), 'airbnb');
+        configFile.load.restore();
+        console.error.restore();
     });
 
     it('should call checker if config does not exist but "preset" option is defined', function() {


mrjoelkemp:
Lgtm

hzoo:
If we are doing this somehow we should let the user know it's the airbnb preset right `Checking against the airbnb preset` and/or docs change?

markelog:
@hzoo other linters don't do anything like that, so i'm not sure in what way we can show so it wouldn't too invasive, docs change are indeed absolutely necessary.

jacksonrayhamilton:
There shouldn't be a default preset. JSCS is an unopinionated style checker; it has to be, because the only style guide anyone likes is his own. Choosing a favorite style guide contradicts that quality.

mrjoelkemp:
> because the only style guide anyone likes is his/her own

That's not entirely true as the airbnb preset is the most highly visited part of the JSCS repo. 

> Choosing a favorite style guide contradicts that quality.

If you supply your own jscsrc, then you are overriding the default. This change does not stop anyone from doing that. Supplying a default preset, in addition to `--auto-configure` allow folks to download the tool and avoid any setup.

jacksonrayhamilton:
This application shouldn't decide that one particular preset is more worthy of a default slot. There are numerous qualities which make for a "good" style guide, but the weight of those qualities is a matter of opinion. Even popularity has to be weighted. It happens to be quantifiable, but that doesn't mean "1000 stars + no semi-colons > 500 stars + strict semi-colons;" basing a decision like this on popularity, or any data point for that matter, requires some bias, which an unopinionated tool should be free of.

Besides that fundamental issue, there will be others:

1. A particular style guide receives favoritism from the premiere JavaScript style checker. This boosts the popularity of the style guide, whether or not the style guide "deserves" it.
1. The favorite style guide of the year falls out of fashion the next. Unfortunately, it's a breaking change to choose a new favorite, so unless you're okay with breaking userland, you're stuck with it.
1. Choosing default settings lets you run the tool without configuration, but as soon as you add configuration, those default settings go away. "Where'd they go?" Should they still be there even if the configuration is an empty object? If so, "how do I get rid of that?" Or the inverse, you use a checker plugin that checks files automatically, but now it wants code without a `.jscsrc` to look a certain way. "Why'd it choose *that* way?"

Concerning Airbnb specifically (and hopefully not validating that it is worth discussing the merits of *any* style guide as a default), I would attribute its high number of hits due to its name starting with "A," thus being sorted to the top of [this list](http://jscs.info/overview.html#presets), which includes a link to the json file I think you are referencing. You could validate my prediction by checking if `crockford.json` and `google.json` have the 2nd and 3rd most hits respectively, if so I'd suspect a fluke.

As for making the tool easier to use "out-of-the-box," you could automatically walk the user through the auto-configuration flow when he tries to use the tool without configuration, and then do a normal check afterwards. That would be nearly as easy for the user, and much more respectful, and less confusing considering the third point above.

mrjoelkemp:
Fair points @jacksonrayhamilton. Thanks for your thoughts. We wanted to try it out in 2.0 and remove it in 3.0 if there was substantial backlash. 

You bring up a good catch though. @markelog would a user have to set preset to null to provide their own custom config. If they don't nullify the preset field, do we view it as overriding the airbnb preset?

mikesherov:
This is still a :-1: from me, and :-1: I believe from @mdevils . @mdevils care to weigh in?

markelog:
Created the issue

hzoo:
:-1: from me I guess (said the same in the issue)

markelog:
Okay ladies and gentlemen, it seems we reached the tie.
By vote of JSCS core team (sorry @hzoo) - 

+1 from me and @mrjoelkemp,
-1 from @mikesherov and @mdevils

It seems final decision is on @zxqfox, after that 2.0 is finally out of the door, otherwise me and @mdevils thinking we should start using `insight` in the next release (introduce it in minor or patch release) to get more definitive data and resurrect that issue for 3.0 if necessary.

@zxqfox i encourage you to read full discussion on https://github.com/jscs-dev/node-jscs/issues/986. So what it would be?

lahmatiy:
> According to our traffic stat and stars of that repo this is a most popular javascript code style out there.

Looks like there is some hype around airbnb code style. It's great for airbnb, but `jscs` shouldn't care about it. If any preset will be by default it isn't fair for any others.

zxqfox:
:-1: for now. Details here: https://github.com/jscs-dev/node-jscs/pull/1548#issuecomment-124875962

mrjoelkemp:
Safe to close?

markelog:
Closing for now

PR #1536, opened 2015-07-17
disallow-space-before-commas

ShashankaNataraj:
Adds a new rule to disallow spaces before commas.

Issue is raised here: https://github.com/jscs-dev/node-jscs/issues/1447

PR #1550, opened 2015-07-21
Misc: check-my-coverage script for locally checking coverage

mrjoelkemp:
Ref https://github.com/jscs-dev/node-jscs/pull/1536#issuecomment-122843241

cc @markelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "autofix-tests": "node test/scripts/integration.js",
     "coverage": "unit-coverage run -p common",
     "coverage-html": "unit-coverage run -p common -r html -o coverage.html",
+    "check-my-coverage": "npm run coverage-html && open coverage.html",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -121,6 +121,7 @@
     "autofix-tests": "node test/scripts/integration.js",
     "coverage": "unit-coverage run -p common",
     "coverage-html": "unit-coverage run -p common -r html -o coverage.html",
+    "check-my-coverage": "npm run coverage-html && open coverage.html",
```
mikesherov: perhaps `coverage-report`?

markelog: That could be useful! Although i'm not sure who cross os this thing.

mrjoelkemp: True. Wouldn't work on windows. We don't get a lot of tickets where we tell users about coverage-html though. So it might not be worthwhile.

PR #1557, opened 2015-07-23
support option  `false` to disable this checking `disallowDanglingUnderscores`

fishbar:


mrjoelkemp:
Hi @fishbar. Thanks for contributing. The proper way to skip the checking of a rule is to supply a value of `null` for that rule in your config: http://jscs.info/overview.html.

PR #1559, opened 2015-07-23
Added JSON reporter

romanblanco:
New reporter that returns errors in JSON format

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -149,7 +149,7 @@ jscs path[ path[...]] --preset=jquery
 In order to add/remove preset rules you will need to create a `.jscsrc` config file.
 
 ### `--reporter` (`-r`)
-`jscs` itself provides six reporters: `checkstyle`, `console`, `inline`, `inlinesingle`, `junit`, `text` and `unix`.
+`jscs` itself provides eight reporters: `checkstyle`, `console`, `inline`, `inlinesingle`, `junit`, `text`, `unix` and `json`.
 ```
 jscs path[ path[...]] --reporter=console
 ```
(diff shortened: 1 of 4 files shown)


mrjoelkemp:
Nice work here @romanblanco. Thanks for contributing! LGTM

markelog:
It seems amount of reporters grows pretty fast, we need to be more conservative here.  

LGTM

markelog:
btw, what is use-case for this?

romanblanco:
@markelog we need JSON output to be able to integrate jscs to GitHub bot for pull request monitoring.

markelog:
Isn't that easier to use API for that sort of thing? What do you do for other runners?

markelog:
?

romanblanco:
@markelog What API do you mean?

markelog:
Lets start from the other end, how do get that info from the other linters? Like jshint, eslint?

markelog:
fyi, we probably gonna release this with 2.1 version, i would like to just better understand your use-case

romanblanco:
@markelog till this time, we only used [haml-lint](https://github.com/brigade/haml-lint) and [rubocop](https://github.com/bbatsov/rubocop) linters in our GitHub bot (which has similar use to https://www.houndci.com/), as you can see [here](https://github.com/ManageIQ/miq_bot/blob/master/app/workers/commit_monitor_handlers/commit_range/rubocop_checker.rb#L30). They both provide JSON output. We would like to use other linters, which means, we need to either make a wrapper to output linter results in JSON, or add another reporter right into linter.

markelog:
And you don't want to provide global/relative reporter for the `--reporter` flag?

romanblanco:
Which global/relative reporter flag you got on mind? I don't quite understand that question.

markelog:
See http://jscs.info/overview.html#-reporter-r, like https://github.com/wurmr/jscs-teamcity-reporter

romanblanco:
Well, yes. I probably could specify my own reporter, if it would be problem to add another into node-jscs. I didn't noticed this option before.

markelog:
Thank you!

PR #1565, opened 2015-07-25
validateIndetation: reduce RegExp create count (optimization)

lahmatiy:
Originaly `validateIndetation` rule creates new RegExp for every single line to check.
PR fix it, rule creates one RegExp per file (it's 100 times less on our code base).
Not sure, but probably it will be better add new private property `_rNotIndentChar` and creates just one RegExp in `configure`.

### DIFF
--- a/lib/rules/validate-indentation.js
+++ b/lib/rules/validate-indentation.js
@@ -246,7 +246,6 @@ module.exports.prototype = {
         }
 
         function getIndentationFromLine(line) {
-            var rNotIndentChar = new RegExp('[^' + indentChar + ']');
             var firstContent = line.search(rNotIndentChar);
             if (firstContent === -1) {
                 firstContent = line.length;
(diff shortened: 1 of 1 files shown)


lahmatiy:
Some numbers. Before changes it creates 21189 regexps for our code base, and now ~200. Just calling `new RegExp('[^' + ' ' + ']')` for 20k times takes about 1-2% of overall checking time :)

markelog:
With isolated `validate-indentation` tests for `time mocha` command.

With changes - `0m1.276s`, without - `0m1.281s`. 

Is that theoretical speed increase? Where did you get that numbers? How big should be the code-base to notice this?

mikesherov:
@markelog lets try 900 files.

markelog:
Tried with jscs code-base with only one `validate-indentation` enabled - 
with `0m1.867s`, without - `0m1.917s`

lahmatiy:
File count is not important, but lines count matter. In my case there are 20k+ lines marked for indentation check.
Truly speaking it's not a huge performance bust. But still creating RegExp is not so great, even it's saves just ~30ms for me.
Another thing is memory. 20k `new RegExp('[^' + ' ' + ']')` allocate 3680000 bytes (3,5MB) RAM for V8. And as they are immediately destroying it's trashing GC as well.
Anyway, moving just one line of code to proper place will make `jscs` a little bit better ;)

markelog:
Thanks :-)

PR #1566, opened 2015-07-26
Add jscs-jsdoc plugin

leotm:
Fixes Sublime 3 SublimeLinter-jscs  *Error: Unsupported rule: jsDoc*
error_unsupported-rule_jsdoc

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,6 +1,7 @@
 {
     "preset": "google",
     "fileExtensions": [".js", "jscs"],
+    "plugins": ["jscs-jsdoc"],
 
     "requireParenthesesAroundIIFE": true,
     "maximumLineLength": 120,


markelog:
This issues was existed because that plugin was added to dev deps. Should work for current master

leotm:
The problem still persists for me with current master, unless you add `"plugins": ["jscs-jsdoc"]`.

Or is the plan to completely remove all the `jsDoc` rules, so that it's no longer compulsory, and purely optional?

markelog:
Okay, lets talk, could you describe the issue? Since here you editing internal .jscsrc config.

With 2.0, which was published recently - https://github.com/jscs-dev/node-jscs/releases/tag/v2.0.0
you should have the ability use JSDoc rules without specifying it in the jscs `plugins` or in package.json dependencies.

leotm:
I download latest release `node-jscs-2.0.0.zip`, extract to `node-jscs-2.0.0`, inside create `script.js`, enter `var a = 2;` and get the same error in the screenshot.

Adding `"plugins": ["jscs-jsdoc"]` to the `.jscsrc` fixes the dependency to the `JsDoc` rules, removing the error.

markelog:
If you use this plugin - https://packagecontrol.io/packages/SublimeLinter-jscs

Then you probably using global version of `jscs`, which is older then 2.0, which causes the error

leotm:
`npm jscs -v` for me returns `2.11.3`...

markelog:
That is you npm version, you need to do `jscs --verison`

leotm:
Agh that was the issue, cheers.

PR #1572, opened 2015-07-28
Link to built-in JSCS plugin for JetBrains IDEs

SimenB:
Not sure if you'd like to word it in some fancy way, so I went for short and sweet :smile:


mrjoelkemp:
Thanks for contributing @SimenB!

PR #1574, opened 2015-07-28
Docs: use correct links to new rules

irnc:
* jscs.info is not served over https://, links changed to use http://
* `s/Tralling/Trailing`
* `ObjectKeysOnNewLine` changed to `requireObjectKeysOnNewLine`


markelog:
Thank you!

PR #1597, opened 2015-07-31
Allow .jscsrc extensions for "preset" value and improve documentation for it

markelog:


OVERVIEW.md
```diff
@@ -221,12 +221,19 @@ Extends defined rules with preset rules.
 
 Type: `String`
 
-Values: `"airbnb"`, `"crockford"`, `"google"`, `"jquery"`, `"mdcs"`, `"node-style-guide"`, `"wikimedia"`, `"wordpress"`, `"yandex"`
+Values: You can choose one of the default presets: `"airbnb"`, `"crockford"`, `"google"`, `"jquery"`, `"mdcs"`, `"node-style-guide"`, `"wikimedia"`, `"wordpress"`, `"yandex"`.
+
+Or you can load from it local path or as node module
```
mrjoelkemp: `Or you can load it from a local path or node module:`

OVERVIEW.md
```diff
@@ -221,12 +221,19 @@ Extends defined rules with preset rules.
...
 
 #### Example
 
 ```js
 "preset": "jquery"
+"preset": "./path-to-your-preset"
+
+// If your preset called "jscs-your-preset-node_modules-path" 
+// You can either define full name or omit "jscs-" prefix -
```
mrjoelkemp: I don't really see the point to the sugar of `jscs-`. It only helps users who want to fork the built-in presets and retain the name. But if they're forking it, they can name it whatever they'd like and just supply that as the preset value.

Any big wins with this prefix?

lib/cli-config.js
```diff
@@ -72,7 +76,15 @@ exports.getContent = function(config, directory) {
                 )
             );
         }
+    } else {
+
+        // Try to load it as a node module
+        try {
+            content = require(config);
+        } catch (e) {}
```
mrjoelkemp: What will the user see if the resolution fails? Said differently, what does the user see if this function returns `undefined`? I assume it'll bubble to cli.js and then show an error about a missing config?

test/specs/config/node-configuration.js
```diff
@@ -127,6 +127,26 @@ describe('modules/config/node-configuration', function() {
...
 
+        it('should load external preset with .jscsrc extension', function() {
+            configuration.registerDefaultRules();
+
+            configuration.load({
+                preset: path.resolve(__dirname + '/../../data/configs/jscsrc/external.jscsrc')
+            });
+
+            var exist = false;
+            configuration.getRegisteredRules().forEach(function(rule) {
```
mrjoelkemp: Use `.some` instead

mrjoelkemp:
Test fail, but LGTM otherwise

OVERVIEW.md
markelog: Prefixes usually added so they wouldn't be confused with other packages since it name without prefix might be misleading and have possibility of the name conflict - https://twitter.com/ironshay/status/370525864523743232

ESlint recent change of the same nature and popularity of https://github.com/sindresorhus/load-grunt-tasks packages somewhat proves it, plus it seems popular for the jscs too - https://www.npmjs.com/search?q=jscs

So we know ppl use those prefixes a lot, so we might take advantage of it...

And you know, it is nicer to write `preset: my-company`, then `preset: jscs-my-company`.

lib/cli-config.js
markelog: Not exactly, but yeah

test/specs/config/node-configuration.js
markelog: Sure, although it seems logical to do in the different commit, since we use `forEach` in the test above

markelog:
Hm, it successes with `mocha` command but fails with `npm run coverage`.

@mdevils could you help us out?

mdevils:
Looks like a bug in unit-coverage, will fix it

OVERVIEW.md
mrjoelkemp: Fair points. +1

PR #1602, opened 2015-08-02
requireMatchingFunctionName: fix critical bug and add tests

### LINKED ISSUE #1601: requireMatchingFunctionName error with predefined variables
If you have defined a variable name before, then assign it to an (anonymous) function, requireMatchingFunctionName will crash. Example code:

```javascript
var abc = function() {
  var myFunction;

  myFunction = function(name) {
    console.log('hello', name);
  };
};
```
The error message:
>Error running rule requireMatchingFunctionName: This is an issue with JSCS and not your codebase.
TypeError: Cannot read property 'name' of undefined
    at checkForMember (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:10499:38)
    at file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:10473:21
    at Array.forEach (native)
    at Object.3.JsFile.iterateNodesByType (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:1755:42)
    at Object.83.module.exports.check (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:10468:14)
    at Object.<anonymous> (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:17118:22)
    at Array.forEach (native)
    at Object.139.StringChecker._checkJsFile (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:17114:31)
    at Object.139.StringChecker.checkString (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/jscs.js:17059:14)
    at Object.<anonymous> (file:///C:/Users/Famlam/Desktop/RedundancyCheck/test%20suite/index.html?module=Other:5008:32) at input :

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25260334-requirematchingfunctionname-error-with-predefined-variables?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zxqfox:
Fixes #1601

### DIFF
--- a/lib/rules/require-matching-function-name.js
+++ b/lib/rules/require-matching-function-name.js
@@ -67,6 +67,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType(['FunctionExpression'], function(node) {
             switch (node.parentNode.type) {
+                // var foo = function bar() {}
                 // object.foo = function bar() {}
                 // object['foo'] = function bar() {}
                 case 'AssignmentExpression':
@@ -95,12 +96,36 @@ module.exports.prototype = {
     }
 };
 
+/**
+ * Fetching name from a Pattern
+ *
+ * @param {Pattern} pattern - E.g. left side of AssignmentExpression
+ * @returns {String|Boolean} - Resolved name or false (if there is an Expression)
+ */
+function _resolvePatternName(pattern) {
+    switch (pattern.type) {
+        case 'Identifier':
+            // prop = ...;
+            return pattern.name;
+        case 'Literal':
+            // obj['prop'] = ...;
+            return pattern.value;
+        case 'MemberExpression':
+            // obj.prop = ...;
+            return _resolvePatternName(pattern.property);
+        default:
+            // Something unhandy like obj['x' + 2] = ...;
+            return false;
+    }
+}
+
 function checkForMember(assignment, skip, errors) {
-    if (skip(assignment.left.property.name, assignment.right)) {
+    var _name = _resolvePatternName(assignment.left);
+    if (_name === false || skip(_name, assignment.right)) {
         return;
     }
 
-    if (assignment.left.property.name !== assignment.right.id.name) {
+    if (_name !== assignment.right.id.name) {
         errors.add(
             'Function name does not match member name',
             assignment.loc.start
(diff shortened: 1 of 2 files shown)


lib/rules/require-matching-function-name.js
```diff
@@ -96,11 +96,13 @@ module.exports.prototype = {
 };
 
 function checkForMember(assignment, skip, errors) {
-    if (skip(assignment.left.property.name, assignment.right)) {
+    var _left = assignment.left;
+    var _name = _left.property ? (_left.property.name || _left.property.value) : _left.name;
```
mrjoelkemp: Non-blocker: For readability, I find it helpful for maintenance when an example code is supplied that clarifies an assignment with a left property that doesn't have a name but has a value. Similarly, what's an assignment that doesn't have a property?

A contributor would have to guess or hope that coverage is 100% and figure out which tests correspond to which part of the condition.

mrjoelkemp:
LGTM aside from the JSHint complaints.

lib/rules/require-matching-function-name.js
zxqfox: There are three options of assignments:
```
var simpleVar = function () {};
obj.propertyName = function () {};
obj['propertyValue'] = function () {};
```
And _left property only exists for the second and the third line.

I'll add here comments.

mrjoelkemp: Comments would be great. Thanks @zxqfox!

zxqfox:
Would be nice to have another review.

mikesherov:
:+1:

test/specs/rules/require-matching-function-name.js
```diff
@@ -45,10 +65,30 @@ describe('rules/require-matching-function-name', function() {
...
+
+        it('should NOT report function assignation to an expression', function() {
+            assertNoErrors('var test = {}; test["x" + 2] = function bar(){};');
+        });
+
+        it('should NOT report function name match on ES6 destructuring', function() {
+            assertNoErrors('let [ bar ] = [ function bar(){} ];');
+        });
+
+        it.skip('should report function name mismatch on ES6 destructuring', function() {
```
mikesherov: On purpose?

zxqfox: For a second I though we need to check names for ES6-destructuring. Feels like need to drop this test.

PR #1604, opened 2015-08-02
Disallow arrow functions

### LINKED ISSUE #1580: Add "pessimistic" arrow function rules
There should be a `disallowArrowFunctions` rule, and a `disallowShorthandArrowFunctions` rule.

It turns out that you don't need `this` to write JavaScript. Object-orientation is perfectly viable with functions that return objects. In fact, it might be better, because if you create objects with functions, you can hide information with closures; whereas everything on `this` has to be publicly accessible.

Since `this` isn't needed, arrow functions' lexical `this` isn't useful. Arrow functions are also less versatile than `function`, because you can't give them names. Arrow functions' tricky syntax is also harmful to maintenance. Adding additional lines to an arrow function which implicitly returns a value will result in that function returning `undefined`, unless you remember to add an explicit `return`. The shorthand syntax is also ambiguous in terms of returning objects, as `(name) => {id: name}` is interpreted as a longhand arrow function with the label `id:`.

Some people might decide to just eliminate them altogether, since `function` seems a lot better if you aren't using `this`. But for people who don't mind `this`, they might want to reap the perceived benefits of arrow functions while avoiding their syntactic problems, by requiring they be written in the most-generalized form, the longhand form. Therefore, it is important these rules be added.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25101917-add-pessimistic-arrow-function-rules?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jacksonrayhamilton:
Fixes #1580.

### DIFF
--- a/lib/rules/disallow-arrow-functions.js
+++ b/lib/rules/disallow-arrow-functions.js
@@ -1,0 +1,56 @@
+/**
+ * Disallows arrow functions.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowArrowFunctions": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * // function expression in a callback
+ * [1, 2, 3].map(function (x) {
+ *     return x * x;
+ * });
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * // arrow function
+ * [1, 2, 3].map((x) => {
+ *     return x * x;
+ * });
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'disallowArrowFunctions';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType(['ArrowFunctionExpression'], function(node) {
+            errors.add('Do not use arrow functions', node.loc.start);
+        });
+    }
+};
--- a/lib/rules/disallow-arrow-functions.js
+++ b/lib/rules/disallow-arrow-functions.js
@@ -1,6 +1,17 @@
 /**
  * Disallows arrow functions.
  *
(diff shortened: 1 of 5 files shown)


mrjoelkemp:
Nice work here @jacksonrayhamilton! Thanks for contributing. 

LGTM

jacksonrayhamilton:
Added justifications per [this suggestion](https://github.com/jscs-dev/node-jscs/issues/1580#issuecomment-127043287).

hzoo:
I'd just like to note we have the `disallowNodeTypes` (http://jscs.info/rule/disallowNodeTypes.html) rule which could be used instead of the `disallowArrowFunctions` rule but doesn't mean we can't have both.

lib/rules/disallow-arrow-functions.js
```diff
@@ -0,0 +1,67 @@
...
+ * Disallows arrow functions.
+ *
+ * Why enable this rule? Arrow functions might cause more problems than they
+ * solve:
+ *
+ * - Object-orientation may be better without ECMAScript's `this`.
+ * - You can't name an arrow function.
+ * - Arrow functions' syntax can cause maintenance issues; see
+ *   `disallowShorthandArrowFunctions`.
+ * - Arrow functions shouldn't be used on prototypes, as objects' methods,
```
markelog: `objects' `?

lib/rules/disallow-shorthand-arrow-functions.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * Require arrow functions to use a block statement (explicit return).
+ *
+ * Why enable this rule? Arrow functions' syntax can cause maintenance issues:
```
markelog: Same as above

markelog:
Couple notes, otherwise good to go

lib/rules/disallow-arrow-functions.js
ronkorving: Possessive form of a plural word ending in 's', seems correct to me.

mrjoelkemp: Yeah, it's correct; though it always looks strange even to native English speakers :)

PR #1605, opened 2015-08-02
Fix reporter name in error message (fixes #1603)

### LINKED ISSUE #1603: Bug: wrong error message when reporter doesn't found
Command:

```
jscs . -r foo
```

Output:

```
Reporter "null" doesn't exist.
```

Expected:

```
Reporter "foo" doesn't exist.
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25282921-bug-wrong-error-message-when-reporter-doesn-t-found?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lahmatiy:
Output correct reporter name in error message when reporter doesn't found

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -70,7 +70,7 @@ module.exports = function(program) {
     returnArgs.reporter = reporter.path;
 
     if (!reporter.writer) {
-        console.error('Reporter "%s" doesn\'t exist.', reporter.writer);
+        console.error('Reporter "%s" doesn\'t exist.', program.reporter);
         returnArgs.reporter = reporter.path;
         defer.reject(6);
 


mrjoelkemp:
Seems like our tests just check for an error code. If we really wanted to test this change, we could enhance https://github.com/jscs-dev/node-jscs/blob/master/test/specs/cli.js#L433-L444 to check the error message. 

Mind giving that a try?

lahmatiy:
I'm not sure we need to test it. Otherwise all messages should be tested.
And actually I don't know how to get printed message. Believe console patching is not good aproach.  Any ideas?

markelog:
> I'm not sure we need to test it. Otherwise all messages should be tested.

Yeah, they should be all tested and believe we do that

> And actually I don't know how to get printed message. Believe console patching is not good aproach. Any ideas?

Stubbing is the only way, we do that all over the place

mrjoelkemp:
ping @lahmatiy. Would love to include this for the 2.1.0 release happening on Monday. Any chance the tests could be added by then?

markelog:
Okay, i added tests, thank you @

PR #1608, opened 2015-08-03
SpacesInAnonymousFunctionExpression: consider ES6 "constructor" method

### LINKED ISSUE #1607: requireSpacesInAnonymousFunctionExpression breaks ES6 classes
I use the following configuration:

```json
"requireSpacesInAnonymousFunctionExpression": {
	"beforeOpeningRoundBrace": true,
	"beforeOpeningCurlyBrace": true
}
```

And it breaks in a simple ES6 class on each declared method. Saying for example (in verbose mode):

```
requireSpacesInAnonymousFunctionExpression: Missing space before opening round brace at lib/MyWriter.js :
     7 |class MyWriter extends Transform {
     8 |
     9 | constructor() {
---------------------^
    10 |  super({ writableObjectMode: true });
    11 |
```

The ES6 code:

```js
class MyWriter extends Transform {
  constructor() {
    super({ writableObjectMode: true });
  }
}
```

These are ES6 class methods, I have `"esnext": true` set, so I don't expect my methods to be considered anonymous functions.

PS: If this is all because of Babel turning these methods into anonymously assigned prototype functions, sure it makes sense why this happens. But it's not obvious and while perhaps by spec these *are* anonymous functions assigned to the prototype, I think they deserve their own rule or perhaps some exception value.

PS2: Wow, even adding a space before the opening brace (ie: `constructor () {`) gives the **same** error. Mind blown.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25306619-requirespacesinanonymousfunctionexpression-breaks-es6-classes?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Fixes #1607

### DIFF
--- a/lib/rules/disallow-spaces-in-anonymous-function-expression.js
+++ b/lib/rules/disallow-spaces-in-anonymous-function-expression.js
@@ -90,8 +90,8 @@ module.exports.prototype = {
                 return;
             }
 
-            // if shorthand method
-            if (parent.method) {
+            // shorthand or constructor methods
+            if (parent.method || parent.type === 'MethodDefinition') {
                 functionNode = parent.key;
             }
 
--- a/lib/rules/require-spaces-in-anonymous-function-expression.js
+++ b/lib/rules/require-spaces-in-anonymous-function-expression.js
@@ -90,8 +90,8 @@ module.exports.prototype = {
                 return;
             }
 
-            // if shorthand method
-            if (parent.method) {
+            // shorthand or constructor methods
+            if (parent.method || parent.type === 'MethodDefinition') {
                 functionNode = parent.key;
             }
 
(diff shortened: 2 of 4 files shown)


zxqfox:
:+1:

hzoo:
:+1:

hzoo:
Can this be merged? I think we need the same fix for `SpacesInFunctionExpression` - can do it here or I'l do another PR. Need to fix https://github.com/jscs-dev/babel-jscs/issues/21 which is similar

hzoo:
Never mind looks like I already it did before in https://github.com/jscs-dev/node-jscs/pull/1388 and forgot to do it for `inAnonymousFunctionExpression`?

PR #1422, opened 2015-06-01
requireTemplateStrings: check if left/right is a string

hzoo:
Fixes #1416

Did a little refactor for if we need to add more types?

PR #1610, opened 2015-08-03
Allow string + string concatenation.

michelle:
This also makes the logic match the comment.

Use case: single-line messages that are too long, like warnings or assertions or test descriptions.

```javascript
// Desired code:
it('should allow really long but peripheral strings to be concatenated, ' +
   'instead of forcing interpolation, which draws attention.', function() {
  if (someFailureCase) {
    console.error('This is a really really long error message that conveys a' +
                  'lot of information. A lot.');
  }
});

// The alternative, if we want to get past `requireTemplateStrings`:
it(`should allow really long but peripheral strings to be concatenated, \
instead of forcing interpolation, which draws attention.`, function() {
  if (someFailureCase) {
    console.error(`This is a really really long error message that conveys a \
lot of information. A lot.`);
  }
});
```

### DIFF
--- a/test/specs/rules/require-template-strings.js
+++ b/test/specs/rules/require-template-strings.js
@@ -7,86 +7,155 @@ describe('rules/require-template-strings', function() {
     beforeEach(function() {
         checker = new Checker();
         checker.registerDefaultRules();
-        checker.configure({ esnext: true, requireTemplateStrings: true });
     });
 
-    it('should report the use of string concatenation with a identifier on the left', function() {
-        assert(checker.checkString('a + "a"').getErrorCount() === 1);
+    function errorChecks(string) {
+        it('should report the use of string concatenation with a identifier on the left', function() {
+            assert(checker.checkString('a + ' + string).getErrorCount() === 1);
+        });
+
+        it('should report the use of string concatenation with a identifier on the right', function() {
+            assert(checker.checkString(string + ' + a').getErrorCount() === 1);
+        });
+
+        it('should report the use of string concatenation with right handed binary expression',
+            function() {
+                assert(checker.checkString('"test" + (a + b)').getErrorCount() === 1);
+            }
+        );
+
+        it('should report the use of string concatenation with left handed binary expression',
+            function() {
+                assert(checker.checkString('(a + b) + "test"').getErrorCount() === 1);
+            }
+        );
+
+        it('should report for the use of string concatenation with a CallExpression',
+            function() {
+                assert(checker.checkString('a() + ' + string).getErrorCount() === 1);
+                assert(checker.checkString(string + ' + a()').getErrorCount() === 1);
+            }
+        );
+
+        it('should report for the use of string concatenation with a MemberExpression',
+            function() {
+                assert(checker.checkString('a.b + ' + string).getErrorCount() === 1);
+                assert(checker.checkString(string + ' + a.b').getErrorCount() === 1);
+            }
+        );
+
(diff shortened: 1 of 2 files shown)


mrjoelkemp:
Thanks for contributing. That seems like a breaking change for the rule. 

I don't see a good way of adding an exception to allExcept. 

There's always disabling of the rule per file with jscs: disable. 

If we had support for extending jscsrc files, then you could have a test-only jscsrc that disabled that rule. 

@hzoo, any opinion on this?

hzoo:
In the issue I mentioned string + string but left it in https://github.com/jscs-dev/node-jscs/pull/1422#issuecomment-107425306 so I had originally intended for allowing it.

> I would say no, we always can wait for the community response on that. - @markelog

mrjoelkemp:
Then I guess we're good here. I foresee folks from the opposite camp wanting a strict mode that disallows string to string. Perhaps we should just do allExcept: ['stringConcatenation']. Though wordy, it doesn't weaken the rule for everyone. 

Thoughts?

zxqfox:
`requireTemplateStrings: except string concat` is good for me

:+1:

hzoo:
Since it's not obvious everyone wants it, I think the `allExcept` option would be ok.

michelle:
Cool! Thanks for the feedback--I'll go ahead and add `allExcept: ['stringConcatenation']`

michelle:
Updated.

How do we feel about template string concatenation? I've also added this to the PR, but can remove if we feel this is breaking.

zxqfox:
> How do we feel about template string concatenation?

Seems legit

michelle:
Cool, will add to comment at the top.

Also, this iteration cargo cults from `requirePaddingNewLinesBeforeLineComments`, which has `allExcepts` as a string rather than an array. I can switch it to an array if preferred.

hzoo:
Yeah, I guess all the `allExcept` options should be an array to allow for multiple of them so the older rules should be changed as well but that's another issue. Otherwise looks good

michelle:
Switched over to using an Array for `allExcept` and updated the documentation/tests.

zxqfox:
Looks good ;-) Nicely done, @michelle 

:+1:

test/specs/rules/require-template-strings.js
```diff
@@ -7,86 +7,155 @@ describe('rules/require-template-strings', function() {
...
+
+        it('should report the use of string concatenation with right handed binary expression',
+            function() {
+                assert(checker.checkString('"test" + (a + b)').getErrorCount() === 1);
+            }
+        );
+
+        it('should report the use of string concatenation with left handed binary expression',
+            function() {
+                assert(checker.checkString('(a + b) + "test"').getErrorCount() === 1);
```
mrjoelkemp: Non-blocker: prefer `assert.equal` instead of a `===` usage for better error messages when this test fails.

test/specs/rules/require-template-strings.js
```diff
@@ -7,86 +7,155 @@ describe('rules/require-template-strings', function() {
...
+            }
+        );
+
+        it('should report for the use of string concatenation with a CallExpression',
+            function() {
```
mrjoelkemp: Non-blocker: a bit confusing on a separate line. I immediately thought this was a nameless function declaration and was confused.

mrjoelkemp:
Nice work here @michelle. Thanks for contributing!

hzoo:
:tada: yeah and thanks for all the updates!

PR #1615, opened 2015-08-04
Misc: update esprima to 2.5.0

hzoo:


mrjoelkemp:
LGTM

zxqfox:
:+1:

mrjoelkemp:
Yeah, we haven't added any functionality directly with this bump, so it's a patch. If we had esprima aa a caret dep, users would get it immediately. Not seeing why it shouldn't qualify for a patch.

mikesherov:
Esprima 2.5.0 brings with it all remaining ES6 funtionality. Technically, this is for a minor version considering it has new features, and some may consider allowing es6 grammar when they haven't asked for it to be a bug.

mrjoelkemp:
> some may consider allowing es6 grammar when they haven't asked for it to
be a bug.

Based on our policy, yes, that's a minor bump. Good point.

hzoo:
Ok I guess this can move to a 2.1.0 milestone

PR #1617, opened 2015-08-04
New Rule: requireDefaultParameters (ES6)

### LINKED ISSUE #1387: New Rule: requireDefaultParameters (ES6)
[airbnb/javascript#7.7](https://github.com/airbnb/javascript#7.7)

> Use default parameter syntax rather than mutating function arguments.

```js
// really bad
function handleThings(opts) {
  // No! We shouldn't mutate function arguments.
  // Double bad: if opts is falsy it'll be set to an object which may
  // be what you want but it can introduce subtle bugs.
  opts = opts || {};
  // ...
}

// still bad
function handleThings(opts) {
  if (opts === void 0) {
    opts = {};
  }
  // ...
}

// good
function handleThings(opts = {}) {
  // ...
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15587668-new-rule-requiredefaultparameters-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1387 

WIP - Kind of crazy to check for the other cases (not the `||` ones) but maybe it's required?

```js
// didn't do yet
function foo(a, b) {
   a = typeof a === 'undefined' ? 42 : a; // swap !== for ===
}

function pick(arg, def) {
   return (typeof arg == 'undefined' ? def : arg); // return instead of assignment
}
```

### DIFF
--- a/lib/rules/require-default-parameters.js
+++ b/lib/rules/require-default-parameters.js
@@ -1,0 +1,75 @@
+/**
+ * Requires use of default parameters instead of checking for undefined in the function body
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireDefaultParameters": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * function multiply(a, b = 1) {}
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * function multiply(a, b) { b = b || 1; }
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireDefaultParameters';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('AssignmentExpression', function(node) {
+            var left = node.left;
+            var right = node.right;
+            if (
+                // b = b || 1
+                right.type === 'LogicalExpression' &&
+                left.type === 'Identifier' && right.left.type === 'Identifier' &&
+                left.name === right.left.name ||
+
+                // b = typeof b !== 'undefined' ? b : 1
+                right.type === 'ConditionalExpression' &&
+                right.test.type === 'BinaryExpression' &&
+                right.test.left.type === 'UnaryExpression' && right.test.left.operator === 'typeof' &&
+                right.test.right.type === 'Literal' && right.test.right.value === 'undefined' &&
+                left.type === 'Identifier' && right.consequent.type === 'Identifier' &&
+                right.test.left.argument.type === 'Identifier' &&
+                left.name === right.consequent.name &&
+                left.name === right.test.left.argument.name
+            ) {
+                errors.add(
+                    'Use default parameters instead of checking for undefined',
+                    node.loc.start.line,
+                    node.loc.start.column
+                );
+            }
+        });
+    }
+
+};
--- a/test/specs/rules/require-default-parameters.js
+++ b/test/specs/rules/require-default-parameters.js
@@ -1,0 +1,38 @@
+var Checker = require('../../../lib/checker');
+var assert = require('assert');
+
+describe('rules/require-default-parameters', function() {
+    var checker;
+
+    beforeEach(function() {
+        checker = new Checker();
+        checker.registerDefaultRules();
+        checker.configure({ requireDefaultParameters: true, esnext: true });
+    });
+
+    it('should not report use of default parameters', function() {
+        assert(checker.checkString([
+            'function multiply(a, b = 1) {',
+                'return a * b;',
+            '}'
+        ].join('\n')).isEmpty());
+    });
+
+    it('should report use of checking for undefined parameters with the || operator', function() {
+        assert(checker.checkString([
+            'function multiply(a, b) {',
+                'b = b || 1;',
+                'return a * b;',
+            '}'
+        ].join('\n')).getErrorCount() === 1);
+    });
+
+    it('should report use of checking for undefined parameters with ternary and typeof', function() {
+        assert(checker.checkString([
+            'function multiply(a, b) {',
+                'b = typeof b !== \'undefined\' ?  b : 1;',
+                'return a * b;',
+            '}'
+        ].join('\n')).getErrorCount() === 1);
+    });
+});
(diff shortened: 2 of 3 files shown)


lib/rules/require-default-parameters.js
```diff
@@ -0,0 +1,75 @@
...
+                // b = typeof b !== 'undefined' ? b : 1
+                right.type === 'ConditionalExpression' &&
+                right.test.type === 'BinaryExpression' &&
+                right.test.left.type === 'UnaryExpression' && right.test.left.operator === 'typeof' &&
+                right.test.right.type === 'Literal' && right.test.right.value === 'undefined' &&
+                left.type === 'Identifier' && right.consequent.type === 'Identifier' &&
+                right.test.left.argument.type === 'Identifier' &&
+                left.name === right.consequent.name &&
+                left.name === right.test.left.argument.name
+            ) {
```
gibson042: What a hairy `if` test! :dizzy_face:

gibson042:
I think this will erroneously report on updating a non-parameter variable (e.g., `var special = false; … for ( p in o ) { special = special || test(o[p]); … }`, cf. [jquery:src/effects.js](https://github.com/jquery/jquery/blob/b3b2d6c3dd51fbdc69e1942e9af75cc99a1834c2/src/effects.js#L118)), and fail to report when using `== null` or `=== undefined` to check an argument (e.g., `a = a == null ? false : a`).

hzoo:
Yeah we need to check if the identifier is a param somehow. And yeah probably needs to error for `===` and `!==` and the other cases I mentioned before

lib/rules/require-default-parameters.js
```diff
@@ -0,0 +1,75 @@
...
+            var left = node.left;
+            var right = node.right;
+            if (
+                // b = b || 1
+                right.type === 'LogicalExpression' &&
+                left.type === 'Identifier' && right.left.type === 'Identifier' &&
+                left.name === right.left.name ||
+
+                // b = typeof b !== 'undefined' ? b : 1
+                right.type === 'ConditionalExpression' &&
```
mrjoelkemp: This doesn't capture the `if` statement form though. 

It also doesn't capture `!(typeof b === 'undefined')` or `b === undefined`

lib/rules/require-default-parameters.js
```diff
@@ -0,0 +1,75 @@
...
+        return 'requireDefaultParameters';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('AssignmentExpression', function(node) {
+            var left = node.left;
+            var right = node.right;
+            if (
+                // b = b || 1
+                right.type === 'LogicalExpression' &&
```
mrjoelkemp: Instead of a massive `if`, I'd opt for a strategy based on the `right.type` value.

```js
var resolvers = {
  LogicalExpression: function() {
    // Some example code that clarifies the condition
    return left.type === 'Identifier' && right.left.type === 'Identifier' && left.name === right.left.name;
  },
  ConditionalExpression: function() {
    return right.test.type === 'BinaryExpression' && right.test.left.type === 'UnaryExpression' && right.test.left.operator === 'typeof' && test.right.type === 'Literal' && right.test.right.value === 'undefined' && left.type === 'Identifier' && right.consequent.type === 'Identifier' && right.test.left.argument.type === 'Identifier' && left.name === right.consequent.name && left.name === right.test.left.argument.name;
  }
};
```

then the check code becomes: 

```js
var resolver = resolvers[right.type];
if (resolver(right, left)) {
  errors.add(...)
}
```

Something like that would break up the condition into logical parts.

lib/rules/require-default-parameters.js
```diff
@@ -0,0 +1,75 @@
...
+                left.type === 'Identifier' && right.left.type === 'Identifier' &&
+                left.name === right.left.name ||
+
+                // b = typeof b !== 'undefined' ? b : 1
+                right.type === 'ConditionalExpression' &&
+                right.test.type === 'BinaryExpression' &&
```
mrjoelkemp: Some explaining variables would maybe help tame this condition:

```js
var isTypeOfUndefined = right.test.type === 'BinaryExpression' && right.test.left.type === 'UnaryExpression' && right.test.left.operator === 'typeof' && right.test.right.type === 'Literal' && right.test.right.value === 'undefined';
```

Then the if reads pretty nicely and is easier to maintain.

test/specs/rules/require-default-parameters.js
```diff
@@ -0,0 +1,38 @@
+var Checker = require('../../../lib/checker');
```
mrjoelkemp: Add test for `if` form:

```js
if (typeof b !== 'undefined')
```

and the other cases not covered by your `if` condition

lib/rules/require-default-parameters.js
```diff
@@ -0,0 +1,75 @@
...
+
+    check: function(file, errors) {
+        file.iterateNodesByType('AssignmentExpression', function(node) {
+            var left = node.left;
+            var right = node.right;
+            if (
+                // b = b || 1
+                right.type === 'LogicalExpression' &&
+                left.type === 'Identifier' && right.left.type === 'Identifier' &&
+                left.name === right.left.name ||
```
mrjoelkemp: This yields false positives though. If a user defaults any other variable, your rule deems it an error. Seems like you'd want to isolate your checks to the function parameters.

mrjoelkemp:
Nice work so far @hzoo. This is a toughie but a goodie.

lib/rules/require-default-parameters.js
hzoo: Yeah I don't know if we have a helper to check for function params.

mrjoelkemp: We don't have a helper. We have node iteration by function types. However, if you need to have utility functions on a function node (like getting the list of params), my vote would be for creating `lib/Function` that is a class that takes in a node and computes the params list from it. It could then have utility methods on it for getting the params list.

It might feel like a yagni to create this abstraction, but `lib/utils` really shouldn't be a grab bag of random methods.

markelog: > It might feel like a yagni to create this abstraction, but lib/utils really shouldn't be a grab bag of random methods.

True that!

markelog:
> Yeah we need to check if the identifier is a param somehow

It could be done with `escope`, sounds like other issues could be handled as well, would you like to finish this up?

hzoo:
Going to revisit later

PR #1618, opened 2015-08-04
Preset: Set jsDoc.checkTypes to "strictNativeCase" for Wikimedia

Krinkle:
Follows-up 5f8c939. Per www.mediawiki.org documentation, Wikimedia
notates primitives in lower case. Only uppercase when referring
to the identifier of a constructor.

Ref <https://mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Types>.
Ref jscs-dev/jscs-jsdoc#137.

/cc @jdforrester

### DIFF
--- a/presets/wikimedia.json
+++ b/presets/wikimedia.json
@@ -79,7 +79,7 @@
         "checkReturnTypes": true,
         "checkRedundantReturns": true,
         "requireReturnTypes": true,
-        "checkTypes": "capitalizedNativeCase",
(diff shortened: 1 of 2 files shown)


zxqfox:
I think we need to update tests for the preset also: https://github.com/jscs-dev/node-jscs/blob/master/test/data/options/preset/wikimedia.js#L19

Krinkle:
Done!

PR #1619, opened 2015-08-04
validateIndentation: Add support for 'offset' switch cases

mtp401:
'offset' switch cases are cases indented by a half indent. Example
(indentSize = 4):
```js
switch (val) {
  case "1":
    a();
    break;
  default:
    b();
    break;
}
```

### DIFF
--- a/lib/rules/validate-indentation.js
+++ b/lib/rules/validate-indentation.js
@@ -32,7 +32,7 @@
  * }
  * ```
  *
- * ##### Invalid example for mode `2`
+ * ##### invalid example for mode `2`
  *
  * ```js
  * if (a) {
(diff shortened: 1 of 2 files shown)


mrjoelkemp:
cc @mikesherov

markelog:
Hm, could you show an example of such code? Is there any code-style that recommends to do that? Is there any editors/IDE that do that?

mtp401:
Google C++ style (and by derivation Google JavaScript Style) recommends this convention. See https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Loops_and_Switch_Statements

markelog:
Not exactly, they recommend to use 2 spaces indentation, so 2 not only for switches but for every block constuction

markelog:
Okay, if you want to continue this discussion, ping us up

PR #1620, opened 2015-08-05
requireBlocksOnNewline: Add object option to handle comments

### LINKED ISSUE #1417: autofix allows edge case to be on one line even if requireBlocksOnNewline is set to true
Hi.

I know the issue I'm reporting is a bit of an edge case and maybe it'll be eventually labelled as "wontfix". However, I was playing with a few weird conditions to test the autofix feature. I set the `requireBlocksOnNewline` to `true` and tested the following code:

```js
if (prop) { /* Do something here with prop... */ } else { return; }
```

I'd expect the code to be formatted as follows:

```js
if (prop) {
   /* Do something here with prop... */
} else {
   return;
}
```

However, this is how it's actually formatted:

```js
if (prop) { /* Do something here with prop... */ } else {
   return;
}
```

The reason is, of course, that a comment is not considered as a block, but it still feels weird.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18951574-autofix-allows-edge-case-to-be-on-one-line-even-if-requireblocksonnewline-is-set-to-true?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oredi:
Add object option { includeComments: true } and { includeComments: true,
minStatements: 1} that allows the rule to turn on or off the inclusion of
comments as part of the block being checked using exisitng rules. A new
config option is introduced to maintain compatability with existing code like
the crockford preset which ignores comments.

Fixes #1417

### DIFF
--- a/lib/rules/require-blocks-on-newline.js
+++ b/lib/rules/require-blocks-on-newline.js
@@ -1,15 +1,35 @@
 /**
  * Requires blocks to begin and end with a newline
  *
- * Types: `Boolean` or `Integer`
+ * Types: `Boolean`, `Integer`, `Object`
  *
- * Values: `true` validates all non-empty blocks,
- * `Integer` specifies a minimum number of statements in the block before validating.
+ * Values:
+ *  - `true` validates all non-empty blocks
+ *  - `Integer` specifies a minimum number of lines containing elements in the block before validating
+ *  - `Object`:
+ *      - `'includeComments'`
+ *          - `true` includes comments as part of the validation
+ *      - `'minLines'`
+ *          - `Integer` specifies a minimum number of lines containing elements in the block before validating
  *
  * #### Example
  *
  * ```js
  * "requireBlocksOnNewline": true
+ * ```
+ * ```js
+ * "requireBlocksOnNewline": 1
+ * ```
+ * ```js
+ * "requireBlocksOnNewline": {
+ *      includeComments: true
+ * }
+ * ```
+ * ```js
+ * "requireBlocksOnNewline": {
+ *      includeComments: true,
+ *      minLines: 1
+ * }
  * ```
  *
  * ##### Valid for mode `true`
@@ -18,6 +38,18 @@
  * if (true) {
  *     doSomething();
  * }
+ * var abc = function() {};
+ * ```
+ * ```
+ * if (true) { //comments
+ *     doSomething();
+ * }
+ * var abc = function() {};
+ * ```
+ * ```
+ * if (true) {
+ *     doSomething();
+ * /** comments *\/}
  * var abc = function() {};
  * ```
  *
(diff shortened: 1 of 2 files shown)


lib/rules/require-blocks-on-newline.js
```diff
@@ -37,27 +65,161 @@
...
+ *
+ * ##### Valid for mode {
+ *      includeComments: true,
+ *      minStatements: `1`
+ * }
+ *
+ * ```js
+ * if (true) {
+ *     //comments
+ *     doSomething();
```
markelog: Well, those are not statements, they are function calls, which makes them an expressions, maybe we need to choose more general term?

oredi:
@markelog how about minLinesOfCode?

oredi:
@markelog Had some free time today so I updated the PR to use minLinesOfCode instead of minStatements.

lib/rules/require-blocks-on-newline.js
```diff
@@ -1,16 +1,32 @@
 /**
  * Requires blocks to begin and end with a newline
  *
- * Types: `Boolean` or `Integer`
+ * Types: `Boolean`, `Integer`, `Object`
  *
  * Values: `true` validates all non-empty blocks,
- * `Integer` specifies a minimum number of statements in the block before validating.
+ * `Integer` specifies a minimum number of statements in the block before validating,
+ * `Object` validates all non-empty blocks with comments included in the validation
```
markelog: Could structure those docs in similar manner - https://github.com/jscs-dev/node-jscs/commit/c536b21ca16b74e44081f4c7d89a38e08439e23c#diff-98bfebb24c1ec81d05c332bf6483e9beR6
?

lib/rules/require-blocks-on-newline.js
```diff
@@ -1,16 +1,32 @@
 /**
  * Requires blocks to begin and end with a newline
  *
- * Types: `Boolean` or `Integer`
+ * Types: `Boolean`, `Integer`, `Object`
  *
  * Values: `true` validates all non-empty blocks,
- * `Integer` specifies a minimum number of statements in the block before validating.
+ * `Integer` specifies a minimum number of statements in the block before validating,
```
markelog: `statements`?

lib/rules/require-blocks-on-newline.js
```diff
@@ -1,16 +1,32 @@
...
+ * `Object` validates all non-empty blocks with comments included in the validation
+ * and it accepts and optional property to specify minimum number of statements in the block before validating
```
markelog: `statements`?

markelog:
Would like to hear another opinion for the new name properties, @hzoo, @mrjoelkemp, @zxqfox?

lib/rules/require-blocks-on-newline.js
```diff
@@ -20,6 +36,18 @@
...
+ * ```
+ * if (true) { //comments
+ *     doSomething();
+ * }
+ * var abc = function() {};
+ * ```
+  * ```
+ * if (true) {
+ *     doSomething();
+ * //comments}
```
zxqfox: Closing bracket under line comment.

lib/rules/require-blocks-on-newline.js
```diff
@@ -37,27 +65,161 @@
...
+ * }
+ * if (true) { doSomething(); }
+ * var abc = function() {};
+ * ```
+ *
+ * ```js
+ * if (true) {
+ *     doSomething();
+ *     doSomethingElse();
+ *     //comments }
```
zxqfox: And here

lib/rules/require-blocks-on-newline.js
```diff
@@ -37,27 +65,161 @@
...
```
zxqfox: And here.

zxqfox:
I think statements are better here than lines of code, but guess entities (or something similar and abstract) are better.

markelog:
@zxqfox we can't use "statement", since as i said, a lot of constructions that would be counted by this rule are not "statements" in terms of ECMAScript specification.

oredi:
I will fix the line comments to become block comments before the closing braces and also restructure the options to follow the referenced documentation. As for the name I will wait for more feedback

zxqfox:
If we also counter multilines inside multiline comment blocks then it could be just `minLines` (without `OfCode`). Good naming is always a hard part ;-)

markelog:
Bikeshedding always a "hard part" as said in that popular proverb ;).

`minLines` sounds good to me

oredi:
I am good with minLines as well but any suggestions the word statements in the documentation as well

   specifies a minimum number of statements in the block before validating

Should I change the

markelog:
constructions? entities?

zxqfox:
elements?

markelog:
nodes?

anyway, @oredi, choose you like best :-)

oredi:
I will try elements, I am away from my PC. Will get to updating the PR in my Day. :)

oredi:
Updated with changes from the comments

markelog:
Thank you!

PR #1621, opened 2015-08-05
requireParenthesesAroundArrowParam: account for a single rest parameter

### LINKED ISSUE #1616: requireParenthesesAroundArrowParam incorrectly flags rest parameters
`.jscsrc`:

```json
{
  "esnext": true,
  "requireParenthesesAroundArrowParam": true
}
```

Code that breaks `requireParenthesesAroundArrowParam`:

```js
const foo = (...args) => {};
```

This however, is okay:

```js
const foo = (args) => {};
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25368094-requireparenthesesaroundarrowparam-incorrectly-flags-rest-parameters?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1616

### DIFF
--- a/lib/rules/require-parentheses-around-arrow-param.js
+++ b/lib/rules/require-parentheses-around-arrow-param.js
@@ -48,6 +48,9 @@ module.exports.prototype = {
     check: function(file, errors) {
         function isWrapped(node) {
             var openParensToken = file.getPrevToken(file.getFirstNodeToken(node));
+            if (openParensToken.value === '...') {
+                openParensToken = file.getPrevToken(openParensToken);
+            }
             var closingParensToken = file.getNextToken(file.getLastNodeToken(node));
             var closingTokenValue = closingParensToken ? closingParensToken.value : '';
 
--- a/test/specs/rules/require-parentheses-around-arrow-param.js
+++ b/test/specs/rules/require-parentheses-around-arrow-param.js
@@ -21,4 +21,8 @@ describe('rules/require-parentheses-around-arrow-param', function() {
     it('should not report an arrow function expression with parens around multiple parameters', function() {
         assert(checker.checkString('[1, 2].map((x, y) => x * x);').isEmpty());
     });
+
+    it('should not report an arrow function expression with a single rest param #1616', function() {
+        assert(checker.checkString('[1, 2].map((...x) => x);').isEmpty());
+    });
 });


hzoo:
This just checks if the type is of `RestElement` and gets the prevToken which should be `(`

lib/rules/require-parentheses-around-arrow-param.js
```diff
@@ -48,6 +48,9 @@ module.exports.prototype = {
     check: function(file, errors) {
         function isWrapped(node) {
             var openParensToken = file.getPrevToken(file.getFirstNodeToken(node));
+            if (node.type !== 'RestElement') {
+                openParensToken = file.getPrevToken(openParensToken);
```
mrjoelkemp: Having a really hard time visualizing/understanding how this works. Why do you need to do this for non-rest elements? If rest elements were failing, performing an action for non-rest elements doesn't seem right (though I see the tests passing). Any insight is appreciated.

mrjoelkemp:
:+1:

lib/rules/require-parentheses-around-arrow-param.js
hzoo: Yeah me too - had to console.log some stuff - Kind of wish there was a utility that showed the ast and then allowed logging/programming around it very easily (so it's easy to debug jscs) - like prevToken, nextToken etc.

But yeah it doesn't make sense. I mean't to check for `...`

Doing a PR now - also adding disallow rule since I forgot.

hzoo:
Ok there's no need to check for disallow since it's a different ast without the parens.

mrjoelkemp:
Regarding your comment about debugging. I've been using [node-debug](https://github.com/visionmedia/debug) a bunch in my projects. We could have a debug channel as the rule name for more control. 

I think it would be cool to do `it.only` on a test, and then run `DEBUG=* mocha` to see the entire debug logging for that test (printing out asts and other info).

Happy to have this incorporated into core (though there's a valid argument for it seemingly cluttering the codebase, so I'll leave it to @markelog to give the go ahead for using `debug`)

lib/rules/require-parentheses-around-arrow-param.js
```diff
@@ -48,6 +48,9 @@ module.exports.prototype = {
     check: function(file, errors) {
         function isWrapped(node) {
             var openParensToken = file.getPrevToken(file.getFirstNodeToken(node));
+            if (openParensToken.value === '...') {
```
mrjoelkemp: This is a drastically different condition and the test seemingly didn't have to change. Something is fishy about that.

hzoo: I think it's because there is 2 parens `((` so it would have the same result if you prevToken.

mrjoelkemp:
Thanks @hzoo

PR #1622, opened 2015-08-05
New Rule: requireComputedPropertyInObject (ES6)

### LINKED ISSUE #1384: requireComputedPropertyInObject (ES6)
requireComputedPropertyInObject

[airbnb/javascript#3.4](https://github.com/airbnb/javascript/tree/master#3.4)

> Use computed property names when creating objects with dynamic property names.

```js
  function getKey(k) {
    return `a key named ${k}`;
  }

  // bad
  const obj = {
    id: 5,
    name: 'San Francisco',
  };
  obj[getKey('enabled')] = true;

  // good
  const obj = {
    id: 5,
    name: 'San Francisco',
    [getKey('enabled')]: true,
  };
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15299111-requirecomputedpropertyinobject-es6?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1384 

Although there's probably false positives for setting a dynamic property from somewhere else like a parameter? Then we'd have to check if the object was created previous or is a parameter?

```js
function a(b) {
  b[func()] = true;
}
```

### DIFF
--- a/lib/rules/require-computed-property-in-object.js
+++ b/lib/rules/require-computed-property-in-object.js
@@ -1,0 +1,63 @@
+/**
+ * Use computed property names when creating objects with dynamic property names
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * Version: `ES6`
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireComputedPropertyInObject": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * const obj = {
+ *   id: 5,
+ *   [getKey('enabled')]: true, // computed property
+ * };
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * const obj = {
+ *   id: 5
+ * };
+ * obj[getKey('enabled')] = true; // property created outside of object creation
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireComputedPropertyInObject';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('AssignmentExpression', function(node) {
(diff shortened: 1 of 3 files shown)


lib/rules/require-computed-property-in-object.js
```diff
@@ -0,0 +1,63 @@
...
+ * };
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * const obj = {
+ *   id: 5
+ * };
+ * [getKey('enabled')] = true; // property created outside of object creation
```
gibson042: Missing `obj` before the first left bracket.

lib/rules/require-computed-property-in-object.js
```diff
@@ -0,0 +1,63 @@
...
+        );
+    },
+
+    getOptionName: function() {
+        return 'requireComputedPropertyInObject';
+    },
+
+    check: function(file, errors) {
+        file.iterateNodesByType('AssignmentExpression', function(node) {
+            if (node.left.type === 'MemberExpression' && node.left.computed) {
```
gibson042: Yeah, this will report in tons of cases where it shouldn't. I think a more reasonable restriction would be to property assignments immediately following an object literal assignment.

hzoo: Yeah that sounds good to me.

hzoo: What's the best way to go about doing that? - I see we have `findPrevToken` helper which can have a type. Like check if the previous node is a `VariableDeclarator` with `init.type === ObjectExpression`

gibson042: I'd probably start with the object literal assignment and iterate _forward_, breaking on the first non-`AssignmentExpression` (but allowing assignments to be wrapped in `SequenceExpression` and/or `ExpressionStatement`) or the first assignment where "left" is not a `MemberExpression` corresponding to the "left" of the object literal assignment.

hzoo:
Going to revisit later

PR #1625, opened 2015-08-05
requireSpaceBeforeObjectValues: fails with spread in object literals

### LINKED ISSUE #1624: Problems using object spread operator with requireSpaceBeforeObjectValues
When trying to use object spreading while requireSpaceBeforeObjectValues is set to true, JSCS will fail to parse the entire file, dumping a large stack trace into the console. This is a minimal example that demonstrates the problem:

```js
var x = {};
var response = {y: {message: 'hello'}, ...x};
```

And the error that is generated:

```sh
$ node_modules/.bin/jscs jscsTest.js 
Error running rule requireSpaceBeforeObjectValues: This is an issue with JSCS and not your codebase.
TypeError: Cannot read property '_tokenIndex' of undefined
    at Object.JsFile.getNextToken (/home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/js-file.js:318:26)
    at /home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/rules/require-space-before-object-values.js:53:37
    at Array.forEach (native)
    at /home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/rules/require-space-before-object-values.js:43:29
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/js-file.js:478:42)
    at Object.module.exports.check (/home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/rules/require-space-before-object-values.js:42:14)
    at null.<anonymous> (/home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/string-checker.js:165:22)
    at Array.forEach (native)
    at StringChecker._checkJsFile (/home/dcpdev/projects/clicklist_web/client/node_modules/jscs/lib/string-checker.js:161:31) at jscsTest.js :
     1 |var x = {};
--------^
     2 |var response = {y: {message: 'hello'}, ...x};
     3 |


1 code style error found.
```

If requireSpaceBeforeObjectValues is turned off in the `.jscsrc`, then the file is parsed with no problems. I am using version 2.0.0 of JSCS.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25409717-problems-using-object-spread-operator-with-requirespacebeforeobjectvalues?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

natemara:
Fixes #1624

### DIFF
--- a/lib/rules/disallow-space-after-object-keys.js
+++ b/lib/rules/disallow-space-after-object-keys.js
@@ -134,7 +134,7 @@ module.exports.prototype = {
             var maxKeyEndPos = 0;
             var tokens = [];
             node.properties.forEach(function(property) {
-                if (property.shorthand || property.kind !== 'init') {
+                if (property.shorthand || property.kind !== 'init' || property._babelType === 'SpreadProperty') {
                     return;
                 }
 
--- a/lib/rules/require-space-after-object-keys.js
+++ b/lib/rules/require-space-after-object-keys.js
@@ -41,7 +41,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand || property.kind !== 'init') {
+                if (property.shorthand || property.kind !== 'init' || property._babelType === 'SpreadProperty') {
                     return;
                 }
 
--- a/lib/rules/require-space-before-object-values.js
+++ b/lib/rules/require-space-before-object-values.js
@@ -41,11 +41,13 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand || property.method || property.kind !== 'init') {
+                if (property.shorthand || property.method || property.kind !== 'init' ||
+                    property._babelType === 'SpreadProperty') {
                     return;
                 }
 
                 var keyToken = file.getFirstNodeToken(property.key);
+
                 var colon = file.findNextToken(keyToken, 'Punctuator', ':');
 
                 errors.assert.whitespaceBetween({
--- a/test/specs/rules/disallow-space-after-object-keys.js
+++ b/test/specs/rules/disallow-space-after-object-keys.js
@@ -185,5 +185,18 @@ describe('rules/disallow-space-after-object-keys', function() {
             assert(checker.checkString('var x = { a () { } };').getErrorCount() === 1);
         });
 
+        it('should allow object literal spreading with spread at end', function() {
+            assert(checker.checkString(
+                'var b = {};\n' +
+                'var x = {a: 1, ...b};'
+            ).isEmpty());
+        });
+
+        it('should allow object literal spreading with spread at beginning', function() {
+            assert(checker.checkString(
+                'var b = {};\n' +
+                'var x = {...b, a: 1};'
+            ).isEmpty());
+        });
     });
 });
--- a/test/specs/rules/require-space-after-object-keys.js
+++ b/test/specs/rules/require-space-after-object-keys.js
@@ -40,6 +40,20 @@ describe('rules/require-space-after-object-keys', function() {
             checker.configure({ esnext: true });
         });
 
+        it('should allow object literal spreading with spread at end', function() {
+            assert(checker.checkString(
+                'var b = {};\n' +
+                'var x = {a : 1, ...b};'
+            ).isEmpty());
+        });
+
+        it('should allow object literal spreading with spread at beginning', function() {
+            assert(checker.checkString(
+                'var b = {};\n' +
+                'var x = {...b, a : 1};'
+            ).isEmpty());
+        });
+
         it('should report es6-methods without a space. #1013', function() {
             assert(checker.checkString('var x = { a() { } };').getErrorCount() === 1);
         });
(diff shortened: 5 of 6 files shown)


hzoo:
For anyone who is confused about `._babelType`, I added that as a private prop to check type since we replaced it in other cases https://github.com/jscs-dev/babel-jscs/blob/1d15f1afa7dc7af24726187fcfb58cf9cdc4ff38/acorn-to-esprima.js#L185

This still brings up the issue about whether we will support babel in jscs itself or if we need to make a custom repo/rules for es7+ features. I'm ok with either.

mrjoelkemp:
> This still brings up the issue about whether we will support babel in jscs itself or if we need to make a custom repo/rules for es7+ features. I'm ok with either

Was this brought up in another issue?

ES7+ rules have the potential to make backward incompatible changes frequently. We try do that only in major versions. 

* To get around this, we could document that certain es7-labeled rules are prone to breakages in patch versions.
* It's an easier user-experience if we have the rules in core. 

I vote to add es7+ rules into core.

lib/rules/disallow-space-after-object-keys.js
```diff
@@ -134,7 +134,7 @@ module.exports.prototype = {
             var maxKeyEndPos = 0;
             var tokens = [];
             node.properties.forEach(function(property) {
-                if (property.shorthand || property.kind !== 'init') {
+                if (property.shorthand || property.kind !== 'init' || property._babelType === 'SpreadProperty') {
```
mrjoelkemp: Not a fan of checking a private member of the node.

mrjoelkemp: 1. Under which circumstances was the type being changed within acorn to esprima?
2. Why was the original type not preserved as `type` and a babel-specific type kept private?

hzoo: This is done because in babel-jscs we change the SpreadProperty node type https://github.com/jscs-dev/babel-jscs/blob/1d15f1afa7dc7af24726187fcfb58cf9cdc4ff38/acorn-to-esprima.js#L181-L189

hzoo: There was this issue https://github.com/babel/babel-eslint/issues/142

https://github.com/babel/babel-eslint/pull/143/files

But it caused another issue in https://github.com/babel/babel-eslint/issues/142#issuecomment-118707124 so had to keep `node.type === 'Property'` and make a new property to keep info about `SpreadProperty`.

But maybe it's only tied to eslint and not jscs?

mrjoelkemp: Ah, it's es7! I guess the complications with supporting es7 syntax is already rearing its head. I understand now, you can't/shouldn't modify the tree with an unsupported node type, so `_babelType` was created to hold artificial types. 

What concerns me is the maintenance of rules that depend on a private property (`_babelType` ) that was made up by an ast transformation lib. IF this feature gets standardized by EStree, then we'll have to wait for Esprima or Acorn to implement it, then update our rule to stop using `_babelType`.

* Making `_babelType` a public property doesn't really solve the issue (though it makes a statement that the artificial node type is part of babel-jscs' public api)

Would love an opinion from @mdevils, @markelog, or @mikesherov on suggestions for moving forward with es7 features.

hzoo: So what happened with eslint was they weren't supporting anything not in spec so we have babel-eslint and also eslint-plugin-babel which rewrites eslint rules to incorporate es7 stuff like decorators etc. However in ESLint 1.0 (http://eslint.org/blog/2015/07/eslint-1.0.0-released/#new-language-features) `ExperimentalRestProperty` and `ExperimentalSpreadProperty` which are both es7 are supported in espree because of all the react users so they wouldn't need to use babel-eslint.

lib/rules/require-space-after-object-keys.js
```diff
@@ -41,7 +41,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(property) {
-                if (property.shorthand || property.kind !== 'init') {
+                if (property.shorthand || property.kind !== 'init' || property._babelType === 'SpreadProperty') {
```
mrjoelkemp: Is this going to affect the CST integration at all?

lib/rules/disallow-space-after-object-keys.js
mrjoelkemp: Thanks for clarifying, Henry. 

What breaks for JSCS if you just make the type `'SpreadProperty`?

hzoo: We just pushed it for babel-eslint and someone reported a bug, so we'll have to test it for jscs. Would be cool to find some projects (es5 and esnext) to test on and add to tests.

mrjoelkemp: As far is this PR is concerned, I'm okay with letting this in as an interim fix. Will wait for one of the other maintainers to give a thumb before merging.

markelog: >  I'm okay with letting this in as an interim fix

Would be cool to have an issue for that, so we wouldn't forget

mrjoelkemp: Opened https://github.com/jscs-dev/node-jscs/issues/1633. Wording of the issue title might not be the best. Feel free to edit.

mrjoelkemp:
Thanks for contributing @natemara!

natemara:
Glad to help!

Issue #1373, opened 2015-05-13
Internal error in requireDollarBeforejQueryAssignment


aknuds1:
```
[info] jscs results in errors
[info]   requireDollarBeforejQueryAssignment @ line 1 char 0
[info]     Error running rule requireDollarBeforejQueryAssignment: This is an issue with JSCS and not your codebase.
[info] Please file an issue (with the stack trace below) at: https://github.com/jscs-dev/node-jscs/issues/new
[info] TypeError: Cannot read property 'indexOf' of undefined
[info]     at /Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/rules/require-dollar-before-jquery-assignment.js:112:24
[info]     at Array.forEach (native)
[info]     at Object.JsFile.iterateNodesByType (/Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/js-file.js:426:42)
[info]     at Object.module.exports.check (/Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/rules/require-dollar-before-jquery-assignment.js:72:14)
[info]     at null.<anonymous> (/Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/string-checker.js:141:22)
[info]     at Array.forEach (native)
[info]     at StringChecker._checkJsFile (/Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/string-checker.js:137:31)
[info]     at StringChecker.checkString (/Users/arve/Projects/gradlehub/portal/node_modules/jscs/lib/string-checker.js:117:14)
[info]     at Object.checkSource (/Users/arve/Projects/gradlehub/portal/node_modules/jscs-loader/index.js:88:24)
[info]     at Object.<anonymous> (/Users/arve/Projects/gradlehub/portal/node_modules/jscs-loader/index.js:137:19)```

**Comment by hzoo** [comment]
Could you also post a code snippet and jscs version? From the stack trace the rule is `requireDollarBeforejQueryAssignment` and `varName` is `undefined`.

**Comment by aknuds1** [comment]
Hm, I've tried running jscs by itself on the file, then the bug isn't triggered, maybe there's a bug in jscs-loader. Will try to find out...

The jscs version is 1.13.1 btw.

**Comment by aknuds1** [comment]
@hzoo I've traced it to line 80 of rules/require-dollar-before-jquery-assignment.js. Problem is that `left.name` is undefined, any idea why this might be? My log of the value of `left`:
```
[...]

PR #1636, opened 2015-08-07
Fix error if object destructuring in require-dollar-before-jquery

SimenB:
#1373

I have no idea how to do this. I left a comment in the file. Could I get some help? :smile:

### DIFF
--- a/lib/rules/require-dollar-before-jquery-assignment.js
+++ b/lib/rules/require-dollar-before-jquery-assignment.js
@@ -76,6 +76,10 @@ module.exports.prototype = {
             var right;
 
             if (type === 'VariableDeclarator') {
+                if (token.id.type === 'ObjectPattern') {
+                    return;
+                }
+
                 left = token.id;
                 varName = left.name;
                 right = token.init;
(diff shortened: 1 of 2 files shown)


lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -76,9 +76,15 @@ module.exports.prototype = {
...
-                varName = left.name;
-                right = token.init;
+                if (token.id.type === 'Identifier') {
+                    left = token.id;
+                    varName = left.name;
+                    right = token.init;
+                } else { // token.id.type === 'ObjectPattern'
+                    // What do I check here?
+                    // token.id.properties is an array both times during the test.
+                    // How to I check first on, then the other?
```
SimenB: I could return maybe, you don't use destructuring for jQuery assignments, do you?

mrjoelkemp:
I just did something like: 

image

In regards to the use within destructuring, in the case of

```js
var {foo} = {foo: $('.foo')};
```

The validation would occur on the right have side.

If it were:

```js
var {foo} = {$foo: $('.foo')};
```

Then the destructuring is invalid since the destructured var would need to be `$foo`.

As such, simply returning is viable.

Would you like to make those adjustments to the PR?

SimenB:
That I can do! I'll do it after work

SimenB:
@mrjoelkemp How about now? :smile:

mrjoelkemp:
LGTM

mrjoelkemp:
We'll fix the commit message before merge. Thanks for contributing!

SimenB:
Oh, that one I forgot, I just squashed away. I'll give the msg a whack

markelog:
Thank you

PR #1644, opened 2015-08-07
SpacesIn*: fix for shorthand methods/class methods, update tests

hzoo:
Fixes jscs-dev/babel-jscs#21

Just copied over the changes
```js
if (parent.method || parent.type === 'MethodDefinition') {
  var functionNode = node.id || node;	
  functionNode = parent.key;
}
```

and copied all the tests over.

~~Still need to do the disallow rules.~~ (Combing functionExpression, anonymousFunctionExpression, namedFunctionExpression as a single rule would be nice)

### DIFF
--- a/lib/rules/disallow-spaces-in-function-expression.js
+++ b/lib/rules/disallow-spaces-in-function-expression.js
@@ -80,6 +80,8 @@ module.exports.prototype = {
         var beforeOpeningCurlyBrace = this._beforeOpeningCurlyBrace;
 
         file.iterateNodesByType('FunctionExpression', function(node) {
+            // for a named function, use node.id
+            var functionNode = node.id || node;
             var parent = node.parentNode;
 
             // Ignore syntactic sugar for getters and setters.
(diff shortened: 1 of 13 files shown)


markelog:
Sounds like some parts of these changes belongs in `utils` module

hzoo:
You mean like `parent.type === 'Property' && (parent.kind === 'get' || parent.kind === 'set')` could be a `util.isGetterMethod` (naming is hard).

Only other one is `parent.method || parent.type === 'MethodDefinition'` which checks for shorthand or class method.

Not sure if it needs to be an util

markelog:
We use those constructions a lot, i guess it would be better to not duplicate the same logic, since if there would be a error, you would need to check all those places and write tests for them too.

We could create another helper module for those cases

mrjoelkemp:
Those are all static methods of a Function class.

markelog:
Okay, created an issue for the `utils` hazard - https://github.com/jscs-dev/node-jscs/issues/1662

PR #1645, opened 2015-08-07
requireAnonymousFunctions: Add exception for function declarations

### LINKED ISSUE #1569: requireAnonymousFunctions support functionDeclaration exception
As I understand https://github.com/jscs-dev/node-jscs/pull/510 this rule should only apply to function expressions, not declarations, but it does apply to both (effectively disallowing function declarations)

    jscs minimal.js 
    Functions must not be named at minimal.js :
         1 |function foo() {
    --------^
         2 |    return 'foo';
         3 |}

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25045893-requireanonymousfunctions-support-functiondeclaration-exception?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kaicataldo:
Fixes #1569

### DIFF
--- a/lib/rules/require-anonymous-functions.js
+++ b/lib/rules/require-anonymous-functions.js
@@ -3,7 +3,11 @@
  *
  * Type: `Boolean`
  *
- * Value: `true`
+ * Values:
+ *  - `true`
+ *  - `Object`:
+ *    - `'allExcept'` array of exceptions:
+ *       - `'functionDeclarations'` ignores function declarations
  *
  * #### Example
  *
@@ -34,6 +38,19 @@
  *
  * });
  * ```
+ *
+ * ##### Valid for `{"allExcept": ["functionDeclarations"]}`
+ *
+ * ```js
+ * function foo() {
+ *
+ * }
+ *
+ * $('#foo').click(function(){
+ *
+ * })
+ * ```
+ *
  */
 
 var assert = require('assert');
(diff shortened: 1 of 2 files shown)


lib/rules/require-anonymous-functions.js
```diff
@@ -3,7 +3,11 @@
  *
  * Type: `Boolean`
  *
- * Value: `true`
+ * Values:
+ *  - `true`
+ *  - `Object`:
+ *    - `'allExcept'` array of exceptions:
+ *       - `'functionDeclarations'` ignores function declarations
```
markelog: Since rule already has 'functions' in their name, maybe we could just use `declarations` instead of `functionDeclarations`?

test/specs/rules/require-anonymous-functions.js
```diff
@@ -30,4 +30,18 @@ describe('rules/require-anonymous-function', function() {
         assert(checker.checkString('$("hi").click(function named(){});').getErrorCount() === 1);
         assert(checker.checkString('var x = function named(){};').getErrorCount() === 1);
     });
+
+    describe('exceptfunctionDeclarations', function() {
```
markelog: Could you call it `allExcept: ["declarations"]`, we do that in couple places, i guess that would be good trend to follow

lib/rules/require-anonymous-functions.js
```diff
@@ -34,6 +38,19 @@
...
  * ```
+ *
+ * ##### Valid for `{"allExcept": ["functionDeclarations"]}`
+ *
+ * ```js
+ * function foo() {
+ *
+ * }
+ *
+ * $('#foo').click(function(){
```
markelog: Perhaps additional spacing is required - `$('#foo').click(function()/*here*/{`?

markelog:
Looks good

lib/rules/require-anonymous-functions.js
kaicataldo: Sure thing - pre-existing examples changed to match as well

markelog:
Awesome, thank you

PR #1648, opened 2015-08-09
Changelog: use conventional-changelog and conventional-github-releaser

stevemao:


markelog:
Interesting, although i run on github requests limitation couple times, even with authentication

stevemao:
Can you use a different token for each project?

markelog:
> Can you use a different token for each project?

Yeah, try that, but github sometimes gives me 4998 times from 5000 tried for the last hour, or something like that, even if i just created the new user and use it crendentials.

markelog:
Anyway, will try these changes at new release

stevemao:
Thanks. It might take a while to make it fit in your workflow and meet your requirement but in a long run it must be beneficial.

markelog:
Closed via c9ee4f70d99f535d1d88425b3088382983164569

PR #1649, opened 2015-08-09
Object property rules: support spread in object literals

hzoo:
Rules modified:

- disallowSpaceBeforeObjectValues
- requireAlignedObjectValues
- requireQuotedKeysInObjects

Ref #1624

Support the rest of the object property rules that would instead object spread (es7)

### DIFF
--- a/lib/rules/require-quoted-keys-in-objects.js
+++ b/lib/rules/require-quoted-keys-in-objects.js
@@ -44,7 +44,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(prop) {
-                if (prop.shorthand || prop.method || prop.kind !== 'init') {
+                if (prop.shorthand || prop.method || prop.kind !== 'init' || prop._babelType === 'SpreadProperty') {
                     return;
                 }
 
--- a/lib/rules/require-quoted-keys-in-objects.js
+++ b/lib/rules/require-quoted-keys-in-objects.js
@@ -25,6 +25,7 @@
  */
 
 var assert = require('assert');
+var utils = require('../utils');
 
 module.exports = function() { };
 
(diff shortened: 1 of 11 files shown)


lib/rules/require-quoted-keys-in-objects.js
```diff
@@ -44,7 +44,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateNodesByType('ObjectExpression', function(node) {
             node.properties.forEach(function(prop) {
-                if (prop.shorthand || prop.method || prop.kind !== 'init') {
+                if (prop.shorthand || prop.method || prop.kind !== 'init' || prop._babelType === 'SpreadProperty') {
```
markelog: Hmm, we can provide function in `utils` module which could get that value, so this private property getter thing would be encapsulated and when push come to shove we could easily change it logic.

hzoo: Ok what do you think it should be called? `utils.getBabelType`, `utils.getType`?

markelog: `utils.getBabelType` might best since, we take just that, not another `node.type` or anything else

hzoo: Ok pushed

lib/utils.js
```diff
@@ -160,6 +160,15 @@ exports.promisify = function(fn) {
 };
 
 /**
+ * Wrapper to encapsulate getting private property for babel type
+ *
+ * @returns {String}
+ */
+exports.getBabelType = function(property) {
+    return property._babelType;
```
markelog: And test?

hzoo: Ok wasn't sure the best way to test but this should work.

lib/rules/require-quoted-keys-in-objects.js
mrjoelkemp: Why not an AST class that contains tree-iterator logic and getBabelType?

hzoo:
good to merge?

mrjoelkemp:
Thought about alternate solutions to adding a new function to utils, but all other abstractions are too specific. 

For example, making a `Property` class with the `getBabelType` is misleading as more than just properties can have a `_babelType`. 

Making it part of an `AST` class could work, but `_babelType` isn't really a property of an AST (though subtrees are also syntax trees). Though, if we were to make an AST class, then we could possible move tree-iterator logic to it. 

Thoughts?

markelog:
I guess we should figure out this after in different issue, so we wouldn't hold this on.

PR #1650, opened 2015-08-09
disallowEmptyBlocks: allow blocks with comments

### LINKED ISSUE #1611: Add option to disallowEmptyBlocks to allow blocks with comments
https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-empty-blocks.js

I'd like this to be allowed:

```
if (foo) {
  ...
} else if (bar) {
  // We shouldn't do anything here because of ...
} else {
  ...
}
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25336551-add-option-to-disallowemptyblocks-to-allow-blocks-with-comments?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

faceleg:
Requires rule to be configured with `allExcept: ['comments']`

Fixes #1611

### DIFF
--- a/lib/rules/disallow-empty-blocks.js
+++ b/lib/rules/disallow-empty-blocks.js
@@ -34,10 +34,21 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(options) {
-        assert(
-            options === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        if (typeof options !== 'object') {
+            assert(
+                options === true,
+                this.getOptionName() + ' option requires a true value or should be removed'
+            );
+
+            var _options = {
+                allExcept: []
+            };
+            return this.configure(_options);
+        }
+
+        if (Array.isArray(options.allExcept)) {
+            this._exceptComments = options.allExcept.indexOf('comments') > -1;
+        }
     },
 
     getOptionName: function() {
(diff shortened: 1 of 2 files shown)


test/specs/rules/disallow-empty-blocks.js
```diff
@@ -125,4 +125,37 @@ describe('rules/disallow-empty-blocks', function() {
     it('should not report empty block for function expressions', function() {
         assert(checker.checkString('var a = function(){};').isEmpty());
     });
+
+    describe('options as object', function() {
+        describe('allExcept as option', function() {
```
markelog: It seems you could unite those two describes, also, could you call it `allExcept: ["comments"]`, we do that in couple places, i guess that would be good trend to follow.

lib/rules/disallow-empty-blocks.js
```diff
@@ -34,10 +34,21 @@ module.exports = function() {};
...
 
     configure: function(options) {
-        assert(
-            options === true,
-            this.getOptionName() + ' option requires a true value or should be removed'
-        );
+        if (typeof options !== 'object') {
+            assert(
+                options === true,
+                this.getOptionName() + ' option requires a true value or should be removed'
```
markelog: Hm, i guess we should change that warning now?

markelog:
Could you also update documentation on top of the rule file?

test/specs/rules/disallow-empty-blocks.js
faceleg: Good idea, I copied most of this from disallow-operator-before-line-break.

faceleg:
@markelog done, let me know if anything else needs to change

markelog:
Thank you!

PR #1661, opened 2015-08-11
Docs: Fix typos in requireCapitalizedComments

yous:
See http://jscs.info/rule/requireCapitalizedComments.html (snapshot: https://github.com/jscs-dev/jscs-dev.github.io/blob/6d3018fd5940236c7ec306ee9e492415078563b2/rule/requireCapitalizedComments.html).


markelog:
Thank you!

PR #1667, opened 2015-08-12
Docs: Update documentation of disallowKeywordsOnNewLine, requirePaddingNewlinesBeforeKeywords, requirePaddingNewlinesInBlocks, requireVarDeclFirst

oredi:
Updating the type, value and example sections of the 4 rules stated in the subject to be consistent with other rule. I left it as 4 commits in the PR since I am modifying different rules. I can squash it if required.

It would be great to have some feedback on requirePaddingNewlinesBeforeKeywords as it has a default array list that is too big to fit into the value section. As I could not find an existing rule with a long list in the documentation. I left it in the example section and reference it in the description.

    * Values:
    *
    * - `true` specifies that the default keywords shown in the example require an empty line above it
    * - `Array` specifies quoted keywords which require an empty line above it
    *
    * #### Example
    *
    * ```js
    * "requirePaddingNewlinesBeforeKeywords": [
    *     "do",
    *     "for",
    *     "if",
    *     "else",
    *     "switch",
    *     "case",
    *     "try",
    *     "catch",
    *     "void",
    *     "while",
    *     "with",
    *     "return",
    *     "typeof",
    *     "function"
    * ]
    * ```

### DIFF
--- a/lib/rules/require-padding-newlines-in-blocks.js
+++ b/lib/rules/require-padding-newlines-in-blocks.js
@@ -1,18 +1,22 @@
 /**
  * Requires blocks to begin and end with 2 newlines
  *
- * Types: `Boolean` or `Integer`
+ * Types: `Boolean`, `Integer`, `Object`
  *
- * Values: `true` validates all non-empty blocks,
- * `Integer` specifies a minimum number of statements in the block before validating.
+ * Values:
+ *  - `true` validates all non-empty blocks
+ *  - `Integer` specifies a minimum number of lines containing elements in the block before validating
+ *  - `Object` (at least one of properties must be true):
+ *      - `'open'`
+ *          - `true` validates that there is a newline after the opening brace in a block
+ *          - `false` ignores the newline validation after the opening brace in a block
+ *      - `'close'`
+ *          - `true` validates that there is a newline before the closing brace in a block
+ *          - `false` ignores the newline validation before the closing brace in a block
  *
  * #### Example
  *
  * ```js
- * "requirePaddingNewlinesInBlocks": true
- * "requirePaddingNewlinesInBlocks": 1
- * "requirePaddingNewlinesInBlocks": { "open": true, "close": true }
- * "requirePaddingNewlinesInBlocks": { "open": false, "close": true }
  * "requirePaddingNewlinesInBlocks": { "open": true, "close": false }
  * ```
  *
--- a/lib/rules/disallow-keywords-on-new-line.js
+++ b/lib/rules/disallow-keywords-on-new-line.js
@@ -1,9 +1,11 @@
 /**
  * Disallows placing keywords on a new line.
  *
- * Type: `Array`
+ * Types: `Array`
  *
- * Values: Array of quoted keywords
+ * Values:
+ *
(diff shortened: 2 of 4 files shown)


lib/rules/require-padding-newlines-in-blocks.js
```diff
@@ -1,18 +1,22 @@
...
+ *          - `true` validates that there is a newline after the opening brace in a block
+ *          - `false` ignores the newline validation after the opening brace in a block
+ *      - `'close'`
+ *          - `true` validates that there is a newline before the closing brace in a block
+ *          - `false` ignores the newline validation before the closing brace in a block
  *
  * #### Example
  *
  * ```js
- * "requirePaddingNewlinesInBlocks": true
```
markelog: Why did you remove those?

markelog:
Looks good!

lib/rules/require-padding-newlines-in-blocks.js
oredi: These were removed to keep the comments DRY. They were there because in the old documentation we do not explain the options parameter. With the new form of documentation, the values section actually provides a more descriptive version of this information. I think it looks better this way and is more consistent with all other rules that had only 1 example listed

oredi:
@markelog Once this PR is done. I am thinking of going through the other rules to do the same. 

Should I follow what we did there and create an issue/PR for every 4-5 rules I am going to update?

Or shall I do 1 issue and multiple PRs  for all the rules tobe updated?

lib/rules/require-padding-newlines-in-blocks.js
markelog: Fair enough

markelog:
> It would be great to have some feedback on requirePaddingNewlinesBeforeKeywords as it has a default array list that is too big to fit into the value section. 

You could do the same thing we did for http://jscs.info/rule/disallowSpaceAfterBinaryOperators.html 
In other words no need to mention full list of default values. Since that `spacedKeywords` array could change, but docs might not.

> I can squash it if required

It would be cool, to make this commit like this - 

```
Docs: improve documentation for various rules

    * disallowKeywordsOnNewLine
        Modify the type and value sections to be consistent with other rules

    * requirePaddingNewlinesInBlocks
        Modify the type, value and example sections to be consistent with other rules

    * requireVarDeclFirst
        Modify the type, value and example sections to be consistent with other rule

    * requirePaddingNewlinesBeforeKeywords
        Modify the type, value and example sections to be consistent with other rule.
        Added valid and invalid cases for array example

    Fixes #1658
```    

See http://jscs.info/contributing.html#commit-message-format

oredi:
@markelog Updated with latest comments. Thanks for the commit message, I was trying to look for something like that as the generic format in the documentation did not cover cases with multiple rules in a single commit like this.

markelog:
> I was trying to look for something like that as the generic format in the documentation did not cover cases with multiple rules in a single commit like this.

I only ask about this style, so we wouldn't clutter commit history.

> Should I follow what we did there and create an issue/PR for every 4-5 rules I am going to update?

That would be cool!

markelog:
Thank you!

Issue #1678, opened 2015-08-14
How can you disable maxErrors?


danechitoaie:
Looking through the code if the value is null then this check should be disabled.

Is this correct?

How do you pass the null value from command line?

hzoo:
You can pass it in the `.jscsrc` config file - also not sure you can pass null but you can pass `Infinity` (searching in the codebase)

```js
{
  "maxErrors": Infinity
}

**Comment by zxqfox** [comment]
Is Infinity supported by JSON?

**Comment by markelog** [comment]
> Is Infinity supported by JSON?

totally

@danechitoaie, not sure of what you asking, if the question is `How can you disable maxErrors?` you can also provide a cli argument too

**Comment by gibson042** [comment]
Um, what? `Infinity` is most certainly _not_ [valid JSON](https://tools.ietf.org/html/rfc7159#section-6): 
> Numeric values that cannot be represented in the grammar below (such as Infinity and NaN) are not permitted.

**Comment by markelog** [comment]
> Um, what? Infinity is most certainly not valid JSON:

okay, okay, you don't need to act so suprised, like this is a common knowledge :-).   
```js
{
  "maxErrors": null
}
```

should be used instead

danechitoaie:
@markelog So I had two questions:

1. How do you disable maxErrors in general? (I got an answer to this above. Looks like you need to set null in .jscsrc)
2. How do you disable maxErrors when using the cli? (it's still not clear here to me how to do this. What do I pass as argument in the cli? `null` doesn't work there as it's interpreted as string.

markelog:
> How do you disable maxErrors when using the cli? (it's still not clear here to me how to do this. What do I pass as argument in the cli? null doesn't work there as it's interpreted as string.

jscs --max-errors Infinity <path>.

Okay, reopening, since we have two types of issues here:
1. Need to document disabling maxErrors option through config and cli
2. Need to improve error output, since now we have this -
```bash
$ jscs --max-errors null lib/cli.js
`maxErrors` option requires positive number or null value`
```

markelog:
btw, @danechitoaie would you like to help us out, and try to tackle one of these issues?

danechitoaie:
Yes, sure. How can I help?

markelog:
[...]

PR #1682, opened 2015-08-15
Added the ability to pass null for maxErrors when running from cli

danechitoaie:
This pull request is related to https://github.com/jscs-dev/node-jscs/issues/1678

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -178,7 +178,7 @@ jscs path[ path[...]] --error-filter=path/to/my/module.js
 Clean output without colors.
 
 ### `--max-errors` (`-m`)
-Set the maximum number of errors to report
+Set the maximum number of errors to report (pass "null" to disable this).
 
 ### `--help` (`-h`)
 Outputs usage information.
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -178,7 +178,7 @@ jscs path[ path[...]] --error-filter=path/to/my/module.js
 Clean output without colors.
 
 ### `--max-errors` (`-m`)
-Set the maximum number of errors to report (pass "null" to disable this).
+Set the maximum number of errors to report (pass -1 to report all errors).
 
 ### `--help` (`-h`)
 Outputs usage information.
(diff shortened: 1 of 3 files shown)


lib/config/configuration.js
```diff
@@ -453,7 +453,7 @@ Configuration.prototype._processConfig = function(config) {
     }
 
     if (options.hasOwnProperty('maxErrors')) {
-        this._loadMaxError(options.maxErrors);
+        this._loadMaxError(options.maxErrors === 'null' ? null : options.maxErrors);
```
markelog: Hm, i wonder who this usually done for such flags, like passing "0", can't remember any popular commands with similar logic.

Anyway, could you add tests for it?

/cc @zxqfox

OVERVIEW.md
```diff
@@ -178,7 +178,7 @@ jscs path[ path[...]] --error-filter=path/to/my/module.js
 Clean output without colors.
 
 ### `--max-errors` (`-m`)
-Set the maximum number of errors to report
+Set the maximum number of errors to report (pass "null" to disable this).
```
markelog: How about http://jscs.info/overview.html#maxerrors?

lib/config/configuration.js
danechitoaie: Usually I've seen -1 used to disable this kind of things. But if we replace null with -1 it will not be backwards compatible. 

If we keep both null and also add -1 I don't think it really makes sense to have these much options and it might be confusing for the users.

How do you think is best to proceed with this?

markelog: Well, it logic to have `null` value in the `.jscsrc` config, since this is how we disable rules, consistency and all that, but passing such value to the cli seems pretty uncommon, just like passing `Infinity` if you think about it.

So i think we can keep both, but document only `null` value for `maxErrors` of `.jscsrc` (when in reality it would accept both) and document `-1` for `max-errors` of cli option  (when in reality it would again accept both) ?

Does it sounds like a good compromise?

zxqfox: :+1: for both

danechitoaie: Ok, will update the PR as soon as I find some time.

markelog:
fyi: we probably gonna release soon, you might still make it :-)

danechitoaie:
Sorry, these days were crazy at work... you don't even want to know... I'll try over the weekend to review the changes and run some tests to make sure it works as expected.

hzoo:
Or we can just finish up the changes?

markelog:
@danechitoaie
> Sorry, these days were crazy at work... you don't even want to know...

I do know, i'm living it :-)

@hzoo 
Only if @danechitoaie does it mind, are you want to take that in?

hzoo:
We can still wait since it's only for the cli

danechitoaie:
Updated the PR. Are we ok now?

markelog:
Would be cool to have a bit more tests for new values

danechitoaie:
Maybe someone can help with the current build error. I'm not familiar with the unit testing framework used.

markelog:
I have finished this up, thank you!

danechitoaie:
Cool, thank you. :+1:

PR #1683, opened 2015-08-16
disallowQuotedKeysInObjects: fix allowing quoted non-reserved keys

### LINKED ISSUE #1669: disallowQuotedKeysInObjects: "allButReserved" allows quoting all keys
When using "allButReserved" as a rule, JSCS does not flag *any* quoted key anymore. For example, running JSCS 2.1.0 on the following snippet produces no error, although it should flag the second key:

```
var obj = {
  'default': true,
  'no-reserved-keyword': false
}
```

I think the error got introduced in [this commit](https://github.com/jscs-dev/node-jscs/commit/91e9a38c45bb41e54feccb6d004be73d54c024c4#diff-850f77639a8b2662d1628faca25a1c98R68) - the final `true` should probably be the third argument to `reservedWords.check`?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25675943-disallowquotedkeysinobjects-allbutreserved-allows-quoting-all-keys?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zxqfox:
Fixes #1669 

/cc @fubu @markelog @hzoo

### DIFF
--- a/lib/rules/disallow-quoted-keys-in-objects.js
+++ b/lib/rules/disallow-quoted-keys-in-objects.js
@@ -6,8 +6,11 @@
  * Values:
  *
  *  - `true` for strict mode
- *  - `"allButReserved"` allows ES3+ reserved words to remain quoted which is helpful
- *    when using this option with JSHint's `es3` flag.
+ *  - `"allButReserved"` (*deprecated* use `"allExcept": ["reserved"]`)
+ * - `Object`:
+ *    - `"allExcept"` array of exceptions:
+ *       - `"reserved"` allows ES3+ reserved words to remain quoted
+ *         which is helpful when using this option with JSHint's `es3` flag.
  *
  * #### Example
  *
@@ -21,7 +24,7 @@
  * var x = { a: { default: 1 } };
  * ```
  *
- * ##### Valid for mode `"allButReserved"`
+ * ##### Valid for mode `{"allExcept": ["reserved"]}`
  *
  * ```js
  * var x = {a: 1, 'default': 2};
@@ -43,11 +46,14 @@ module.exports.prototype = {
 
     configure: function(options) {
         assert(
-            options === true || options === 'allButReserved',
-            this.getOptionName() + ' option requires a true value or "allButReserved"'
+            options === true || options === 'allButReserved' || typeof options === 'object',
+            this.getOptionName() + ' option requires a true value or an object'
         );
 
-        this._mode = options;
+        this._exceptReserved = options === 'allButReserved';
+        if (Array.isArray(options.allExcept)) {
+            this._exceptReserved = options.allExcept.indexOf('reserved') !== -1;
+        }
     },
 
     getOptionName: function() {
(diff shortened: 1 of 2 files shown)


lib/rules/disallow-quoted-keys-in-objects.js
```diff
@@ -65,7 +65,7 @@ module.exports.prototype = {
                     typeof key.value === 'string' &&
                     KEY_NAME_RE.test(key.value)
                 ) {
-                    if (mode === 'allButReserved' && (reservedWords.check(key.value, file.getDialect()), true)) {
+                    if (mode === 'allButReserved' && reservedWords.check(key.value, file.getDialect(), true)) {
```
markelog: funny )

test/specs/rules/disallow-quoted-keys-in-objects.js
```diff
@@ -67,4 +67,12 @@ describe('rules/disallow-quoted-keys-in-objects', function() {
 
         assert(checker.checkString('var x = { "null": 1, undefined: "foo" }').isEmpty());
     });
+
+    it('should report non-reserved quoted keys when "allButReserved" mode is used. #1669', function() {
```
markelog: We have some tests for this option now, i think it would be logical to move it into its own `describe`  and move
```js
checker = new Checker();
checker.registerDefaultRules();
checker.configure({ disallowQuotedKeysInObjects: 'allButReserved' });
```
bit there too

zxqfox: Yeah, we also need here to go for `{ allExcept: ["reserved"] }`. Should we do it now or... in other PR?

markelog: Deprecating `allButReserved `? Yeah, that would be cool.

> Should we do it now or... in other PR?

If that is not too much trouble

zxqfox: @markelog Something like that: https://github.com/zxqfox/node-jscs/commit/fbd1e86a6d00e0dc44dc58fcf161bba924d80613

zxqfox:
Need one more review ;-) Rework and allButReserved deprecation included.

/cc @markelog

lib/rules/disallow-quoted-keys-in-objects.js
```diff
@@ -43,11 +46,14 @@ module.exports.prototype = {
 
     configure: function(options) {
         assert(
-            options === true || options === 'allButReserved',
-            this.getOptionName() + ' option requires a true value or "allButReserved"'
+            options === true || options === 'allButReserved' || typeof options === 'object',
+            this.getOptionName() + ' option requires a true value or object with "allExcept" key'
```
markelog: aaah, for some reason we use "`" for quoting such entities

markelog:
good to me

PR #1697, opened 2015-08-18
Docs: make it clearer node_modules is excluded, and ! can be used to …

hzoo:
…include

Maybe we don't need to do this but it may be clearer like this?

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -255,14 +255,13 @@ Values: Array of file matching patterns
 #### Example
 
 ```js
-"excludeFiles": ["node_modules/**", "src/!(bar|foo)"]
+// Use `"!foo"` to specifically include a file/folder
+"excludeFiles": ["folder_to_exclude/**", "src/!(bar|foo)"]
 ```
 
 #### Default
 
-```js
-"excludeFiles": ["node_modules/**"]
-```
+`node_modules` folder is excluded by default
 
 ### fileExtensions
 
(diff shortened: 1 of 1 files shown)


zxqfox:
:+1: from me.

@markelog `#Default` block looks like inconsistency ;-( What you think?

hzoo:
Yeah I think so. Showing it like `"excludeFiles": ["node_modules/**"]` was kind of weird since it makes it sound like you still have to do it yourself?

OVERVIEW.md
```diff
@@ -252,16 +252,14 @@ Type: `Array`
 
 Values: Array of file matching patterns
 
-#### Example
+> `node_modules` is excluded by default
```
markelog: The next string would also be included in the quote block

markelog:
@zxqfox 
You mean in `fileExtensions`? Yeah, i think we should be consistent

OVERVIEW.md
hzoo: Yeah I was just going to include both in the quote

hzoo:
DId you want me to change something with this?

markelog:
It seems now we have inconsistency of how we document `fileExtension` and `excludeFiles`?

OVERVIEW.md
markelog: Why this is a quote?

hzoo:
Ok updated

markelog:
LGTM

PR #1704, opened 2015-08-19
CLI:Configuration: load local jscs modules if present

### LINKED ISSUE #894: Global vs local installations and api changes
Discussion was started in #863 
It's more like a question for 2.0 than a real case.

Some points:
Using a globally installed `jscs` for 2 different projects that uses different rules can broke the check if one project uses new rules and other uses some deprecated rules. And locally installed jscs doesn't care about rules it uses. (I'm not sure but the latest version doesn't fully compatible with 1.5.x. The last time I've faced that it forced me to update .jscsrc file)
A big bunch of plugins for IDE requires global installation of `jscs`. It's easier just execute a `jscs` than search for it in `node_modules/.bin`, etc. But with global installation it should be updated somehow or notify user about updates. There is solution with `*-cli` package.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7623802-global-vs-local-installations-and-api-changes?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Fixes #894


zxqfox:
@markelog `npm i resolve --save` ?

markelog:
yep

mrjoelkemp:
You're a beast.

mrjoelkemp:
:+1:

PR #1717, opened 2015-08-22
Configuration: allow load of external entities from external preset

markelog:
@hzoo could you try this on your preset?


hzoo:
Yep looks like this PR works! `"preset": "preset-x"` and `"preset": "jscs-preset-x"` works with global npm install of jscs.

mrjoelkemp:
:+1:

Issue #1685, opened 2015-08-16
add common exceptions to requireCapitalizedComments by default


markelog:
Like for jshint, jscs, etc

alawatthe:
Hmm, why do we want to do this? Writing
```js
"requireCapitalizedComments": { "allExcept": ["jshint"] }
```
does not seem to be to complicated...

How would one then disable these default values?

Furthermore, eshint sometimes uses the keyword ```global``` to start comments (see [Specifying Globals](http://eslint.org/docs/user-guide/configuring#specifying-globals)). So should ```global``` then also be a default exception to requireCapitalizedComments? I can easily see comments starting with ```global``` which are not eslint comments and therefore should be capitalized.

markelog:
> Hmm, why do we want to do this?

To make rule more friendly, since such comments, are not exactly comments but controlling instructions, similar to the `"use strict"` literal

> How would one then disable these default values?

"requireCapitalizedComments": { "allExcept": [] }

> I can easily see comments starting with global which are not eslint comments and therefore should be capitalized.

Yeah, so there is two ways of dealing with this, argue that amount of false positive would be greatly lower then legitimated cases or even further, you could implement heuristic for each case, which would make false negatives to nil.

Or don't include it in the default list.

alawatthe:
For the record, I am not against these default values, I am just playing the devil's advocate here :-)

>> How would one then disable these default values?

>"requireCapitalizedComments": { "allExcept": [] }

This means, that setting the ```"allExcept"``` option overrides the default values, right? So the default values are only there for the
```js
"requireCapitalizedComments": true
```
case. This feels a little inconsistent :-/

markelog:
> This means, that setting the "allExcept" option overrides the default values, right?

Well, yeah, that is why they are "default".

> This feels a little inconsistent :-/

With what?

alawatthe:
> With what?

With setting the ```"allExcept"```option. Let me explain a little further:
Let's say I am using the ```"requireCapitalizedComments": true``` way and also use these default values. Next I want to add one further value, like so ```"requireCapitalizedComments": { "allExcept": ['foobar'] }``` and now the default values are gone. This means, I now get errors for using the default values, as I (implicitely) overwrote them.

I think our different opinions came from the fact that you saw the set of default values as one which can be overwritten and I saw it as a set, which can be extended by the user.

But ok, ok, you convinced me :-D
Shall I implement these changes?

markelog:
[...]

PR #1729, opened 2015-08-25
New Rule: validateCommentPosition

### LINKED ISSUE #651: Rule to validate the positioning of comments about a statement
### validateCommentPosition

Type: `'String'`

Values:

* 'above'
* 'beside'

### Example

`validateCommentPosition: 'above'`

Valid:

```js
// This is a valid comment
1 + 1;
```

Invalid:
```js
1 + 1; // An invalid comment
```

`validateCommentPosition: 'beside'`

Valid:

```js
1 + 1; // A valid comment
```

Invalid:
```js
// This is an invalid comment
1 + 1;
```

If you'd like to support both styles, then don't supply a value to this rule.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4696455-rule-to-validate-the-positioning-of-comments-about-a-statement?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bjdixon:
Fixes #651

### DIFF
--- a/lib/rules/validate-comment-position.js
+++ b/lib/rules/validate-comment-position.js
@@ -12,13 +12,13 @@
  * Value:
  *
  * - `Object`:
- *    - `mode`: `above` or `beside`
- *    - `allExcept`: array of quoted exceptions
+ *    - `position`: `above` or `beside`
+ *    - `allExcept`: array of quoted exceptions (comments that start with these values will be excepted)
  *
  * #### Example
  *
  * ```js
- * "validateCommentPosition": { mode: `above`, allExcept: [`pragma`] }
+ * "validateCommentPosition": { position: `above`, allExcept: [`pragma`] }
  * ```
  *
  * ##### Valid
@@ -36,7 +36,7 @@
  * ```
  *
  * ```js
- * "validateCommentPosition": { mode: `beside` }
+ * "validateCommentPosition": { position: `beside` }
  * ```
  *
  * ##### Valid
@@ -62,15 +62,15 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(options) {
-        var modes = {
+        var validPositions = {
             'above': 'above',
             'beside': 'beside'
         };
         var allExcept = options.allExcept;
         this._exceptions = inlineConfigurationKeywords;
         assert(
-            typeof options === 'object' && modes[options.mode],
-            this.getOptionName() + ' requires one of the following values: ' + Object.keys(modes).join(', ')
+            typeof options === 'object' && validPositions[options.position],
+            this.getOptionName() + ' requires one of the following values: ' + Object.keys(validPositions).join(', ')
         );
         if (Array.isArray(allExcept)) {
             assert(
@@ -79,7 +79,7 @@ module.exports.prototype = {
             );
             this._exceptions = this._exceptions.concat(allExcept);
         }
-        this._mode = options.mode;
+        this._position = options.position;
     },
 
     getOptionName: function() {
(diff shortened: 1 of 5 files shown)


markelog:
How about `/**/`?

bjdixon:
I guess I assumed this was for line comments. I'll add block comments if that's desired. Let me know.

Also, I had a question. For the 'beside' style this is fine:
``` 
// jscs: disable 
1 + 1;
``` 
but should enable trigger an error or pass?
```
// jscs: enable
1 + 1;
```

markelog:
> I guess I assumed this was for line comments.

Yeah, i was asking if you ignore them or not

> but should 'enable' trigger an error or pass?

i'd say just ignore them. I guess we would need to expand on other constructions as well? See discussion in https://github.com/jscs-dev/node-jscs/issues/1685

bjdixon:
I've added some tests showing that block comments are ignored. I've also added an array to ultils.js to filter out inline configurations and an allExcept parameter to this rule.

markelog:
LGTM

lib/rules/validate-comment-position.js
```diff
@@ -0,0 +1,117 @@
...
+ * Comments that start with the following keywords are also ignored:
+ * `eslint`, `jshint`, `jslint`, `istanbul`, `global`, `exported`, `jscs`, `falls through`
+ * eg. // jshint strict: true
+ *
+ * Type: `Object`
+ *
+ * Value:
+ *
+ * - `Object`:
+ *    - `mode`: `above` or `beside`
```
mrjoelkemp: Why not just call it `position`?

lib/rules/validate-comment-position.js
```diff
@@ -0,0 +1,117 @@
...
+
+    getOptionName: function() {
+        return 'validateCommentPosition';
+    },
+
+    check: function(file, errors) {
+
+        function isExcepted(comment) {
+            var trimmedComment = comment.trim();
+            for (var i = 0; i < exceptions.length; i++) {
```
mrjoelkemp: I was confused as to where `exceptions` was defined. Seems like that var declaration `var exceptions = this._exceptions;` could be moved up above this helper.

mrjoelkemp: Also consider just using `return exceptions.some(...`

lib/rules/validate-comment-position.js
```diff
@@ -0,0 +1,117 @@
...
+    },
+
+    getOptionName: function() {
+        return 'validateCommentPosition';
+    },
+
+    check: function(file, errors) {
+
+        function isExcepted(comment) {
+            var trimmedComment = comment.trim();
```
mrjoelkemp: Why do you need to trim?

test/specs/rules/validate-comment-position.js
```diff
@@ -0,0 +1,186 @@
...
+            }, assert.AssertionError);
+        });
+    });
+
+    describe('mode value "above"', function() {
+        beforeEach(function() {
+            checker.configure({ validateCommentPosition: { mode: 'above' }});
+        });
+
+        it('should report unexpected position for 1 + 1; // invalid comment', function() {
```
mrjoelkemp: Non-blocker: we typically don't put code in the descriptions. A general description of the code would suffice. Like `it('should report on a comment beside a statement', ...`

mrjoelkemp:
How about a test for:

```
/* my sweet comment */
/* my super sweet comment */
```

or any combination of stacked comments

mrjoelkemp:
Nice work here @bjdixon!

bjdixon:
@mrjoelkemp 

position is a much better description than mode. I've made that change throughout.

I've manually hoisted the variable declarations in the check function. Hopefully that's more readable now.

I used trim and didn't use ```exceptions.some(...)``` because I wanted to skip comments that begin with the keywords (ignoring whitespace). ```trimmedComment.indexOf(exceptions[i]) >= 0``` was a bug that I've corrected to ```trimmedComment.indexOf(exceptions[i]) === 0```. All the eslint, jshint etc pragmas begin with the keywords listed. I think that skipping comments if they just contain keywords may be too permissive. eg. ```// I really like global warming``` isn't a valid pragma but ```// global var1, var2 ``` is. Let me know if you disagree with this approach.

I've made changes to the test descriptions to (hopefully) be in keeping with the style you suggested. I also added a test for stacked comments.

mrjoelkemp:
Woot! Looks great @bjdixon. Thanks for the effort.

@markelog mind giving this another glance?

zxqfox:
Looks good. Is this the first rule without boolean value? ;-)

alawatthe:
I had a look at this pr while looking at https://github.com/jscs-dev/node-jscs/issues/1685. One thing I noted was, that you are checking, if the comment starts with one of the ```inlineConfigurationKeywords``` meaning, that a comment like

```js
// globalize this function
```
will be recognized as pragma (since "global" is one of the ```inlineConfigurationKeywords```), although it is not a pragma.

I would suggest something like

```js
 if (trimmedComment.split(/( |:|-)/)[0] === exceptions[i]) {
```

Yes, I know, this does not work with "falls through", but for this we could introduce a special case.

markelog:
Sounds like we should move pragma checker to `utils` not only the list alone

bjdixon:
@markelog Yeah, I think having the pragma checker function in utils will remove some duplication.

@alawatthe good point on words that begin comments that could be made up of keywords + other characters. I implemented using a small concatenation instead of the regex + extra special case check as it's faster and should be easier to maintain. If the consensus is to use regex though I'm happy to do that.

markelog:
Okay, added one more test so coveralls wouldn't swear

PR #1731, opened 2015-08-26
added exceptions for requireCapitalizedComments - fixes #1685

### LINKED ISSUE #1685: add common exceptions to requireCapitalizedComments by default

alawatthe:


markelog:
We just landed `utils.isPragma` - https://github.com/jscs-dev/node-jscs/commit/8db0ee95cab00a450f5bda2f8fc4e7d134d0e70e, could you applied it here?

alawatthe:
I thought a little more on the extend vs. overwrite the defaults topic. I think it is way more common, that a user wants to extend the default set, than overwrite it. Therefore I decided to extend the default set. This also adds consistency with ```validateCommentPosition```.

It looks like AppVeyor has some hiccups...

&lt;rant>
Yes, coveralls I know, that there is now one line less which is covered in the file.
But the file is still 100% covered. So why exactly are you failing my build? :-|
&lt;/rant>

markelog:
> But the file is still 100% covered. So why exactly are you failing my build? :-|

Added couple tests in https://github.com/jscs-dev/node-jscs/commit/6153d1fcdff98098de635b1a26e948e2d49214b5, now you are 100% 

Landed! Thank you

PR #1733, opened 2015-08-28
YAML configuration support

### LINKED ISSUE #1732: Yaml support for .jscsrc
I know this is a recurring theme about once a year. I would like to raise it once more, as my reasons are different from the ones before (where it was about the ability to comment out a rule).

To me it's very simple...
YAML is just so much nicer to the human eye than JSON. I love it, and I love how ESLint for example supports it out of the box. I think writing the configuration would look so much more readable in YAML format than JSON. Of course I'm not saying "drop JSON", but allow for both. It would be a real blessing for readability.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26231220-yaml-support-for-jscsrc?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ronkorving:
This adds YAML configuration file support. It also introduces [jsonlint](https://www.npmjs.com/package/jsonlint) to make configuration parse errors more verbose.

Closes #1732


hzoo:
Looks good (doesn't seem too bad) other than maybe adding some tests to see if the config works like a regular file- so a yaml file with some rules/options to see it working correctly.

ronkorving:
What would be the right place for that YAML file, and what would be right place for the test?

hzoo:
Not sure either right now :stuck_out_tongue: - https://github.com/jscs-dev/node-jscs/tree/master/test/data/configs looks ok - maybe a folder for yaml.

Or maybe the test could be that it should produce the same json/result after parsing both with the same rule/config?

ronkorving:
@markelog any input on this?

ronkorving:
This is how ESLint actually opens JSON and YAML: https://github.com/eslint/eslint/blob/02c6c848d7d70c285269c54718dbc9c12d3f0720/lib/config.js#L82-L88

It makes sense, as after stripping comments, [JSON is a subset of YAML 1.2](https://en.wikipedia.org/wiki/YAML#JSON). Perhaps I should update this PR to apply the same approach.

test/data/configs/yaml/.jscsrc
```diff
@@ -0,0 +1,3 @@
+# should allow comments in this type of config
+
+type: yaml
```
markelog: mm, interesting, so jscs would identify type of the format

lib/cli-config.js
```diff
@@ -70,12 +72,16 @@ exports.getContent = function(config, directory) {
...
-                stripJSONComments(
-                    fs.readFileSync(configPath, 'utf8')
-                )
-            );
+            data = fs.readFileSync(configPath, 'utf8');
+
+            try {
+                content = JSON.parse(stripJSONComments(data));
+            } catch (e) {
+                content = yaml.safeLoad(data);
```
markelog: Here we probably should throw our own error, to get meanigful output, so it would be less confusing if user uses json format

test/data/configs/yaml/.jscsrc
ronkorving: No, this is just for a unit test to match against.

markelog: I meant a `try..catch` thing, anyhow

lib/cli-config.js
ronkorving: How would something like this be (JSON is a subset of YAML, so this would work and is how ESLint loads .eslintrc):

```js
try {
  content = yaml.safeLoad(stripJSONComments(data));
} catch (e) {
  throw new Error('Parse error in ' + configPath);
}
```

markelog: How about we check (after trim) that first symbol is `{` if it is - throw json error, if it is not - try the yaml parser and throw their error?

ronkorving:
```yaml
{
  "foo": "bar"
}
```

^ that is valid YAML.

markelog: Since yaml is subset of json that is expected, although whole point of supporting yaml config is to not write it in json, isn't?

Anyway, this would be a downside of format recognition, if user would want to use yaml explicitly their would need to just rename the file. I guess it is still better then weird error.

ronkorving: So what's wrong with my suggested approach? (or isn't there?)
Again, it works wonders for ESLint.

markelog: If you would have error in .jscsrc file while using json format, you would throw yaml error, which could be confusing, not to mention two nested `try..catch`

ronkorving: Hence my custom exception in the code I wrote in a comment above. I'm not throwing a YAML error, nor a JSON error.

markelog: The one you proposing?
```js
throw new Error('Parse error in ' + configPath);
```

It is better to provide more information, it might be pretty big issue to find config in big yaml/json file

ronkorving: Yeah, I agree with you there. The error generated by JSON.parse is generally pretty useless though. It doesn't tell you what or where. Just that there's a parse error, right?

markelog: Something is better nothing, good point though, how about we try https://github.com/zaach/jsonlint, so we could output nice info?

ronkorving: Oh, I'm a big fan of jsonlint. All for it. But that seems to be outside of the scope of this PR. This PR with what I'm proposing does not make the error more cryptic. Fair enough, "parse error" is cryptic, but that's always been the case here. Is this something we can iterate on?

markelog: I wouldn't say it is out of scope, different commit maybe but not a pr, since this pr would change error output, from json error to custom one, with jsonlint we could preserve and improve that message rather then regress it

markelog: fyi, it is not a blocker, if you wouldn't want to do this, that is okay! You did a great thing here, we could finish this up for you

ronkorving: I don't even mind doing it myself :) I just think it would be good in a separate PR, that's all. Appreciate the feedback though, thanks :)

So for now, are you okay with me switching to the code I pasted above? I can update the commit to do that and have the thrown error be a generic `new Error('Parse error in ' + configPath);`.

ronkorving: Understood. The biggest challenge is figuring out what the user's intention was with regards to format. We could write it:

```js
data = stripJSONComments(fs.readFileSync(configPath, 'utf8'));

try {
  content = yaml.safeLoad(data);
} catch (yamlError) {
  // not valid YAML, so by definition also not valid JSON
  if (data[0] === '{') {
    // assume the intention was JSON
    content = jsonlint.parse(data);
  } else {
    throw yamlError;
  }
}
```

Would that work for you? I could update it to be like that.

markelog: Well, i have only one concern, that incorrect configs like
```js
{
preset: 'preset' // property without quotes
}
```

That could be confusing

ronkorving: That would be valid YAML and parsed correctly.

markelog: Yes, but user might meant to use json, i'm not sure that anyone would except file like
```js
{
preset: 'preset' 
}
```
to have a yaml format

ronkorving: I don't really see why not to be honest... It's valid YAML.
Are you sure you wanna go down the road of what is pretty yaml vs. ugly yaml and do detection based on those assumptions?

In any case, if the YAML parsing fails (because they really don't know what they're doing), they *will* get a pretty JSON parse error with the code I wrote above.

Would you rather have it reversed? Try JSON first, if that fails, try YAML, if that fails throw an error? I can switch it around...

markelog: Yaml first would be more logical, but i think it is less clear, since other devs might not expect it and that might be considered as magic or as our own format

markelog: > I can switch it around...

Yeah, i think it would be best for now, maybe we could switch it around in the future

lib/cli-config.js
```diff
@@ -68,14 +71,34 @@ exports.getContent = function(config, directory) {
...
+                        if (stripJSONComments(data).trim()[0] === '{') {
+                            // the intention was probably JSON
+                            throw jsonError;
+                        } else {
+                            // assume the intention was YAML
+                            throw yamlError;
+                        }
+                    }
+                }
+            }
```
ronkorving: @markelog let me know what you think of this.
Is the `{` check overkill?

lib/cli-config.js
```diff
@@ -68,14 +71,34 @@ exports.getContent = function(config, directory) {
         config = path.basename(config);
         ext = path.extname(configPath);
 
-        if (ext === '.js' || ext === '.json') {
```
ronkorving: FYI: I moved the JSON down the the final else statement, so that we can benefit from jsonlint's parser.

lib/cli-config.js
zxqfox: Better to let user know why his/her config wasn't loaded. :+1: for this check.

zxqfox: What you think about determining format with this check?
`if (=== '{') then try to parseJson else try to parseYaml.`

package.json
```diff
@@ -69,7 +69,9 @@
     "estraverse": "^4.1.0",
     "exit": "~0.1.2",
     "glob": "^5.0.1",
+    "js-yaml": "~3.4.0",
```
zxqfox: Perhaps it can be optional. Thoughts?

lib/cli-config.js
ronkorving: Because if it starts with { it may still be valid YAML.

zxqfox: Then this check is pointless. ;-(

package.json
ronkorving: Optional deps always install anyway (it's just that installation failure is allowed. read: compilation failure, and js-yaml has no compiled parts). It wouldn't change the installation behavior or requirements.

ronkorving:
@markelog let me know what you think of the current approach. I think you may like it.

ronkorving:
@markelog any input on the current implementation?

markelog:
Don't worry, one way or another these changes will be in the next release

ronkorving:
:+1:

PR #1737, opened 2015-08-31
remove capitalizedNativeCase from Google presets

samthor:
This isn't the case, and in many cases can encourage incorrect type annotations. The canonical style guide (plus I'm writing this in my full capacity as a Google engineer), [allows both cases equally](https://google.github.io/styleguide/javascriptguide.xml?showone=JavaScript_Types#JavaScript_Types), see the types on "String object" versus "String value".

For a clearer example- `capitalizedNativeCase` enforces `String` over `string`. However, `String` is a nullable, object form of `string`. The Closure compiler will then allow `null`, which might not be what you want.

### DIFF
--- a/presets/google.json
+++ b/presets/google.json
@@ -83,7 +83,7 @@
         "checkReturnTypes": true,
         "checkRedundantReturns": true,
         "requireReturnTypes": true,
-        "checkTypes": "capitalizedNativeCase",
+        "checkTypes": true,
         "checkRedundantAccess": true,
         "requireNewlineAfterDescription": true
     }


markelog:
LGTM

samthor:
Thanks! I can't merge this - can you do it?

hzoo:
The next release should be soon?

Somewhat relevant to this - do we want to have a discussion/issue about allowing owners to manage the preset config in separate repos? We would need to add them to our `package.json` and other users would need to manage the version themselves. (Saying this since there's the ability to create your own preset in 2.0) - could be done for jquery, airbnb, google? @jscs-dev

samthor:
@hzoo You presumably still have to merge my change, even if I'm waiting for the next release.

hzoo:
Ok done. Thanks @samthor

PR #1758, opened 2015-09-05
New Rule: disallowTabs

ottumm:
Disallows tabs anywhere in the file/input.

### DIFF
--- a/lib/rules/disallow-tabs.js
+++ b/lib/rules/disallow-tabs.js
@@ -46,8 +46,9 @@ module.exports.prototype = {
 
     check: function(file, errors) {
         file.getLines().forEach(function(line, i) {
-            if (line.match(/\t/)) {
-                errors.add('Tab found', i + 1, 0);
+            var match = line.match(/\t/);
+            if (match) {
+                errors.add('Tab found', i + 1, match.index);
             }
         });
     }
--- a/lib/rules/disallow-tabs.js
+++ b/lib/rules/disallow-tabs.js
@@ -1,0 +1,54 @@
+/**
+ * Disallows tabs everywhere.
+ *
+ * Type: `Boolean`
+ *
+ * Value: `true`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowTabs": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * if (true) {
+ * \s\sfoo();
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * if (true){
+ * \tfoo();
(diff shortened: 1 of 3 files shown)


lib/rules/disallow-tabs.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+
+    getOptionName: function() {
+        return 'disallowTabs';
+    },
+
+    check: function(file, errors) {
+        file.getLines().forEach(function(line, i) {
+            if (line.match(/\t/)) {
+                errors.add('Tab found', i + 1, 0);
```
markelog: Maybe we could be more specific then that?

markelog:
I guess this duplicates in some way duplicates
```js
{ 
  "validateIndentation": 4
}
```

But not fully

mikesherov:
@ottumm thanks for contributing? what about tabs within strings or comments?

ottumm:
@markelog yeah, would it be better to extend validateIndentation to handle this? My use case is to completely disallow tabs. I was finding that validateIndentation wasn't doing that and in fact was exhibiting problems when tabs were present. 

@mikesherov you're referring to literal tabs within strings, right (not '\t')? I was under the impression that tabs need to be escaped in JavaScript strings. 

Regarding comments, I could add an option to ignore tabs within comments.

lib/rules/disallow-tabs.js
ottumm: Yes, gladly. Did you have something in mind?

ottumm: Maybe "Tab characters are not allowed"?

markelog:
> yeah, would it be better to extend validateIndentation to handle this? My use case is to completely disallow tabs.

I think it is better to have a different rule for it.

>  I was finding that validateIndentation wasn't doing that and in fact was exhibiting problems when tabs were present.

What problems?

lib/rules/disallow-tabs.js
markelog: I meant put a column number there,

ottumm:
I filed issue #1762 to track the validateIndentation issue I'm seeing.

lib/rules/disallow-tabs.js
ottumm: Ah, gotcha, ok.

ottumm: https://github.com/ottumm/node-jscs/commit/e0dad2ebd3c1089410e950febd76d2a7c17ea8af addresses this.

markelog:
Landed!

ottumm:
Awesome, thanks!

PR #1761, opened 2015-09-08
js-file: do not interpret html as grit instructions

### LINKED ISSUE #1578: Autofixing quote is commenting out code
We have a few cases in our code base of multi line template strings like this:

```
{
  html: '\
      <iframe class="classes"\
              src="foo<%= foo %>">\
      </iframe>\
    '
}
```

When running the autofix for `validateQuoteMark`, it returns this:

```
{
  html: '\
/*  <iframe class="classes"\
              src="foo<%= foo %>*/">\
/*  </iframe>*/\
    '
}
```

How did those comment tags get in there?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25081729-autofixing-quote-is-commenting-out-code?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markelog:
Nothing to do with autofixing or with `validateQuoteMark` rule, just
an error in regexp.

Also does not catch cases when tags are actually tags not grit.
Can't do nothing about it at the moment, until we can extend parser

Fixes #1578

/cc @tonyganch

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -936,7 +936,7 @@ function parseJavaScriptSource(source, esprima, esprimaOptions) {
     var hasGritData = false;
 
     // Process grit tags like `<if ...>` and `<include ...>`
-    source = source.replace(/^\s*<\/?\s*(if|include)(?!\w)[^]*?>/gm, function(str, p1, pos) {
+    source = source.replace(/^\s*<\/?\s*(if|include)(?!\w)[^]*?>/gim, function(str, p1, pos) {
         hasGritData = true;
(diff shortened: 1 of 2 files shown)


markelog:
> Can't do nothing about it at the moment, until we can extend parser

/cc @mikesherov

mikesherov:
:+1:

PR #1764, opened 2015-09-08
Misc: add node 4.0 to travis

hzoo:


.travis.yml
```diff
@@ -12,3 +12,4 @@ node_js:
   - "0.10"
   - "0.12"
   - 'iojs'
+  - "4.0"
```
zxqfox: EOL at EOF

.travis.yml
```diff
@@ -12,3 +12,4 @@ node_js:
   - "0.10"
   - "0.12"
   - 'iojs'
```
zxqfox: Hm, we still need it?

hzoo: Yeah it won't be supported anymore - dono

zxqfox: I'm not sure that `iojs` will install 3.x ;-)

.travis.yml
hzoo: oops fixed

.travis.yml
hzoo: On this current PR `iojs` -> `Now using io.js v3.3.0`

zxqfox:
LGTM then ;-) Need to watch iojs.

mzgol:
IMO it should say `"4"`, not `"4.0"`. You'll miss minor updates to the v4 line which will arrive soon for sure. Only the newest 4.x release is supported by upstream so it doesn't make sense to test specifically on `4.0`. Node now follows semver so no breaking changes are allowed to land before Node 5.0.0.

.travis.yml
mzgol: io.js will most likely receive a minor update or two but no major ones so it will die soon anyway.

You may keep it for now but in a month or two no one will use it. :) Few devs switched from Node anyway and those who did will surely switch to Node 4 quickly.

hzoo:
Oops

.travis.yml
hzoo: Right if people are using iojs they are definitely switching to node 4 asap

Issue #1759, opened 2015-09-07
disallowQuotedKeysInObjects:allButReserved no longer includes ES3 future reserved words


Krinkle:
For example https://gerrit.wikimedia.org/r/#/c/235864/4/resources/src/mediawiki/mediawiki.jqueryMsg.js required an inline `// jscs: disable disallowQuotedKeysInObjects` to pass properly for code containing an object literal with an `"int"` key.

This used to work I believe, but regressed in v2.x.

I suspect 91e9a38c45bb41e54feccb6d004be73d54c024c4 ("introduce reservedWords instead of utils.isES3") to be the cause:
```diff
                     if (mode === 'allButReserved' &&
-                        (utils.isEs3Keyword(key.value) || utils.isEs3FutureReservedWord(key.value))		
+                        (reservedWords.check(key.value, file.getDialect()), true)
                     ) {		
```

It used to always check both ES3 keywords and future reserved keywords. Now it uses an abstraction layer that takes the "file dialect" as argument (ES3 or ES5?). I don't know where that comes from or how to control it but I guess it's set to ES5, thus causing `int` to no longer be in the effective list of reserved words.

How do I enable ES3 mode?

Krinkle:
Ah, found it.

Changelog -> "es3" -> "Docs: add es3 option to OVERVIEW.md"

PR #1770, opened 2015-09-09
Wikimedia preset: Add es3: true

jdforrester:
Our browser support matrix (https://www.mediawiki.org/wiki/Compatibility#Browser_support_matrix) covers IE8+ for JS, so we'll need to maintain ES3 compatibility (and stop jscs from thinking we're erring) for a while yet.

Follows-up #1759

### DIFF
--- a/presets/wikimedia.json
+++ b/presets/wikimedia.json
@@ -1,4 +1,5 @@
 {
+    "es3": true,
     "requireCurlyBraces": [
         "if",
         "else",


Krinkle:
:+1: At this time, our default run-time environment requires ES3-compatibility. Individual products may feature-test out of this, but that requires an explicit action.

markelog:
Landed, btw, would do think of creating your own npm preset package, if you would, we could just make it a dependency of jscs so you can update it independently?

jdforrester:
> btw, would do think of creating your own npm preset package, if you would, we could just make it a dependency of jscs so you can update it independently?

Happy to do so; is there a naming convention you like? `wikimedia-jscs-preset`?

hzoo:
I think `jscs-preset-wikimedia` would be good (easy to search for on npm/github) https://www.npmjs.com/search?q=jscs-preset-

jdforrester:
Done: https://github.com/wikimedia/jscs-preset-wikimedia

PR #1777, opened 2015-09-11
Improve *SpacesInsideParentheses rules

### LINKED ISSUE #836: Ignore/Except Quotes in requireSpacesInsideParentheses

markelog:
* Exclude quotes
* Introduce the `ignoreParenthesizedExpression` option for `requreSpacesInsideParentheses` rule

Fixes #836
Ref #1069

### DIFF
--- a/lib/rules/disallow-spaces-inside-parentheses.js
+++ b/lib/rules/disallow-spaces-inside-parentheses.js
@@ -1,7 +1,7 @@
 /**
  * Disallows space after opening round bracket and before closing.
  *
- * Types: `Object` or `Boolean`
+ * Types: `Boolean` or `Object`
  *
  * Values: Either `true` or Object with `"only"` property as an array of tokens
  *
@@ -20,12 +20,12 @@
  * ##### Valid for `only` value
  *
  * ```js
- * "disallowSpacesInsideParentheses": { "only": [ "{", "}" ] }
+ * "disallowSpacesInsideParentheses": { "only": [ "{", "}", "\"" ] }
  * ```
- *
  * ```js
  * var x = ( 1 + 2 );
  * var x = foo({});
+ * var x = foo("1");
  * ```
  *
  * ##### Invalid
@@ -60,13 +60,24 @@ module.exports.prototype = {
             assert(false, error);
         }
 
+        this._onlySingleQuote = false;
+        this._onlyDoubleQuote = false;
+        this._only = null;
+
         if (option.only) {
             this._only = {};
+
             (option.only).forEach(function(value) {
+                if (value === '\'') {
+                    this._onlySingleQuote = true;
+                }
+
+                if (value === '"') {
+                    this._onlyDoubleQuote = true;
+                }
+
                 this._only[value] = true;
             }, this);
-        } else {
-            this._only = null;
         }
     },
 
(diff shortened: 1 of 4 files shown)


markelog:
@gibson042 with this change, we can include `disallowSpacesInsideParenthesizedExpression: true` and `requireSpacesInsideParentheses: {"all": true, "ignoreParenthesizedExpression": true}` for jQuery preset.

You still want to introduce such change to jQuery code style? Do you want to introduce other parentheses rules?

gibson042:
@markelog I like the result, but not the configuration syntax. Wouldn't `allExcept: ["ParenthesizedExpression"]` be more in line with recent deprecations than introducing independent `ignore` properties?

> Do you want to introduce other parentheses rules?

I do, but haven't gotten around to it yet. However, allowing production-based exceptions like this for `requreSpacesInsideParentheses` seems to make such rules unnecessary for jQuery (but then again, _many_ JSCS rules would become redundant through expansion of `only` and `allExcept` on more general forms).

P.S. Are you planning to introduce categorization-based exceptions in `disallowSpacesInsideParentheses` here as well, to maintain parity?

markelog:
> Wouldn't allExcept: ["ParenthesizedExpression"] be more in line with recent deprecations than introducing independent ignore properties?

The thing is, `allExcept` for this rule accepts array of token values, whereas this value wouldn't be a token. So i'm not sure which is better - polymorhic configuration value or new value.

> However, allowing production-based exceptions like this for requreSpacesInsideParentheses seems to make such rules unnecessary

Yeah, my thought exactly

> (but then again, many JSCS rules would become redundant through expansion of only and allExcept on more general forms).

I guess it depends on what is easier to support and implement some check - new additional rule value or new rule.

> P.S. Are you planning to introduce categorization-based exceptions in disallowSpacesInsideParentheses here as well, to maintain parity?

We tried the parity approach, but it generated rules or rule values that no one uses, so we would ask a contributor to do something time consuming that no one would use, that seems wrong, consistent, but wrong.

Rule values are in chaotic state at the moment, we have a lot of rules and the more rules we have the more chaotic it gets.

Hoping to straight it up in 3.0 - new config, removing unused or deprecated values/rules, etc.Yeah, hoping...

gibson042:
Well, this rule satisfies jQuery, but it's going to start looking ugly when others need to ignore spaces inside declaration/invocation/statement parentheses. Maybe 4 cases are manageable, though.

markelog:
Okay, lets see how it would go, will you open an issue in github.com/jquery/contribute.jquery.org?

PR #1779, opened 2015-09-13
Support autofixing for semicolon rules

markelog:


markelog:
Needed for other changes, landed, if there would be any issue will revert

Issue #1771, opened 2015-09-10
Non-stylistic rules


markelog:
Do we want it? It seems we are slowly getting to that point.

We could implement our own rules or use eslint/jshint/jslint as a dependency using their own set of rules or both.

With "dependency" approach, we would need to figure out bridge to it and settle on documentation issues.

/cc @nzakas, @rwaldron, @mdevils

hzoo:
Not sure exactly what should be done but :+1: for better integration of tools at least.

ESLint can consume/pass ASTs http://eslint.org/blog/2015/08/eslint-v1.3.0-released/#using-asts-with-the-nodejs-api

ronkorving:
Why bother complicating things? I already use ESLint for safety checks and JSCS for style checks. I'm absolutely fine using 2 tools for 2 different purposes. I don't need them mixed up. Next year, someone will make "ESLint 2000 Pro" and I may want to use that instead of ESLint. That should have no impact on how I use JSCS.

The *only* added value I can see is sharing ASTs, so performance could be boosted. But then I would suggest releasing a module that parses a file's AST and then passes it through a JSCS API and an ESLint API. At least then you're not creating dependencies between ESLint and JSCS.

mrjoelkemp:
That's the ideal usage: use jscs for style checks and another linter for
other linting needs.

Honestly, it's painful to see communities reinventing the wheel when it
comes to linting rules. Everyone loses when we stall progress implementing
the same features across numerous tools (eslint, babel, esformatter).

We've heard from the eslint community and our own that folks like the
convenience of one tool that covers linting needs. We have the machinery to
do non-style rules already but have punted those rules to eslint for
semantic adherence.

To my earlier point about not duplicating effort, using eslint as a
dependency makes most sense.

If most of the parties are in agreement, then we can list out the problem
areas bridging the two tools.

ronkorving:
I'm glad you don't want to reinvent the wheel. But what's the purpose of adding ESLint as a dependency at all? Do people really want to deal with a single tool, period? It's not gonna make configuration any easier (documentation will still be split) and if I want to upgrade ESLint to the latest release, I'm now stuck waiting for JSCS to be updated first. This adds negative value to me as a user, I see nothing positive about it.

Again, if people want 1 tool so badly, why not make a separate package for this that has both ESLint and JSCS as a dependency? Why clutter the jscs code base?

Just my 2 cents, coming from a long time fan and user of jscs.

PR #1784, opened 2015-09-13
New rule: disallowUnusedParams

markelog:
Ref #1771

### DIFF
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -533,7 +533,7 @@ TokenAssert.prototype.noTrailingSpaces = function(options) {
                 currentLineNumber--;
             }
 
-            if (typeof precendingToken === 'undefined') {
+            if (precendingToken === null) {
                 targetToken = this._file.getFirstToken();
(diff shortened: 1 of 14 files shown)


hzoo:
Interesting so we're adding escope

lib/token-assert.js
```diff
@@ -570,7 +570,7 @@ TokenAssert.prototype.noTrailingSpaces = function(options) {
                 }
             }
 
-            if (typeof targetToken !== 'undefined' && !fixed) {
+            if (typeof targetToken !== null && !fixed) {
```
mikesherov: missed one.

mikesherov:
If we're adding this, also remove `unused` from `jshintrc` please

markelog:
> If we're adding this

We discussed this couple times in gitter, it seems we were generally liking the idea. Plus it is pretty useful to have is as autofix.

> also remove unused from jshintrc please

`unused` also catches the vars, whereas this rule works only for params

PR #1789, opened 2015-09-15
Misc: start to decompose validate indentation logic

markelog:
We stopped developing indentation logic, since bus factor of it is very low it is hard to understand and maintain, we need to decompose that logic.

WIP


mikesherov:
I'll review. I had a certain vision for how this should work, so I'll add more thoughts here about it as well.

hzoo:
I guess we can close this since it was rewritten in 3.0 anyway?

PR #1791, opened 2015-09-17
Docs: remove unnecessary paragraph

dsine-de:
On the "rules" page (http://jscs.info/rules), the word "Expression" is the only text describing the rule and I don't think it adds something or even belongs here, probably this was a copying error.


hzoo:
Thanks @dsine-de! Yeah it was probably copied from `functionExpression` or just a typo.

PR #1794, opened 2015-09-17
Docs: fix Markdown syntax in requireBlocksOnNewline

marian-r:


hzoo:
Does 
```
##### Valid for mode `{
```

work? If not I would just change all of those to be "```javascript"

hzoo:
Thanks I just finished up the modifications.

Issue #1787, opened 2015-09-14
Exponentiation Operator


alawatthe:
I'd like to include the [exponentiation operator](https://github.com/rwaldron/exponentiation-operator) (```**``` and ```**=```) into the [binary operator list](https://github.com/jscs-dev/node-jscs/blob/master/lib/utils.js#L175). And wanted to use the opportunity to ask, at which stage new JavaScript features have to be, that we include rules for them. i.e. is it too soon to include the stage 3 proposal exponentiation operator?

markelog:
> I'd like to include the exponentiation operator (** and **=) into the binary operator list. 

Yeah sure, lets do it

> use the opportunity to ask

Stage 2 i suppose, rules with earlier stages could be made as a plugin.

hzoo:
Should be easy to include this with tests

alawatthe:
Yes, will do this later today.
And I agree with the stage 2 plan.

PR #1797, opened 2015-09-18
Added the exponentiation operator to the binary operators list - fixe…

alawatthe:
…s #1787 

I am not really a fan of having to add ```esnext: true``` everywhere. But that was the best idea I had. I'm open to suggestions.


hzoo:
You will need to put it in every configure or put it in the `beforeAll`

hzoo:
:+1: after moving `esnext` config stuff up a level

alawatthe:
Thank you for reviewing :-D
I put the configuration into the beforeEach. 

The tests for require-space-before-binary-operators had an issue: it claimed that the option is set to ```value```but it was set to ```[operator]```. I fixed that, too.
Same for require-space-after-binary-operators. This test furthermore had the ```'should report sticky operator for ' + sticked + ' with true option'``` test, which is already covered by the (now) first test in the second call of ```[[operator], true].forEach( ```. I removed the obsolete test.

I hope my chaotic explanations of this make sense :-)

zxqfox:
Now we have `esnext: true` for all tests. But guess it's fine. LGTM.

hzoo:
Yeah now it's technically not tested with `esnext: false` anymore but yeah should be ok.

hzoo:
@alawatthe Ok so we noticed `npm test` was taking long and travis was taking much longer so we're going to revert this for now.

Was getting 14.92s before this PR and 66.93 after

markelog:
@alawatthe these changes makes our build significantly slower, (10 times slower in fact), we have concerns that it might reflect on real use-cases, so we had to revert it and release without it.

Could you check this what the issue is?

alawatthe:
No worries, I agree with reverting the commit. Any ideas how we can make these tests faster? The only one I have is to only enable esnext in those tests, where the operator is ```**``` or ```**=```. I'll update the pull request later today and we'll see if there is a speed up.

zxqfox:
The only idea I have is using matrix: http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

@hzoo Do we have an issue for speeding up tests?

PR #1803, opened 2015-09-22
[WIP] jsDoc: add generator and update rule set

zxqfox:
/cc @markelog @hzoo

### DIFF
--- a/lib/rules/jsdoc.js
+++ b/lib/rules/jsdoc.js
@@ -22,8 +22,8 @@
  *
  * There are 3 presets for `Closure Compiler`, `JSDoc3` and `JSDuck5`.
  *
- * By default it allows any tag of mixed set. You can pass `Object`
- * to select preset with `preset` field and add custom tags with `extra` field.
+ * By default it allows any tag of mixed set. You can pass `Object` to select
+ * preset with `preset` field and add custom tags with `extra` field.
  *
  * Type: `Boolean` or `String` or `{"preset": String, "extra": Object}`
  * (see [tag values](#user-content-tag-values))
@@ -95,6 +95,51 @@
  *
  * ```js
  * /** @still-invalid *\/
+ * ```
+ *
+ * ### checkParamExistence
+ *
+ * Ensures all parameters are documented.
+ *
+ * Type: `Boolean`
+ *
+ * Values: `true`
+ *
+ *
+ * #### Example
+ *
+ * ```js
+ * "checkParamExistence": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * /**
+ *  * @param {string} message
+ *  * @return {string}
+ *  *\/
+ * function _f ( message ) {
+ *   return true;
+ * }
+ *
+ * /**
+ *  * @inheritdoc
+ *  *\/
+ * function _f ( message ) {
+ *   return true;
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * /**
+ *  * @return {string}
+ *  *\/
+ * function _f ( message ) {
+ *   return true;
+ * }
  * ```
  *
  * ### checkParamNames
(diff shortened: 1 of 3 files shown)


lib/rules/jsdoc.js
```diff
@@ -97,6 +97,51 @@
  * /** @still-invalid *\/
  * ```
  *
+ * ### checkParamExistence
+ *
+ * Ensures all parameters are documented.
```
hzoo: Do we want to just say `Checks`?

lib/rules/jsdoc.js
```diff
@@ -22,8 +22,8 @@
  *
  * There are 3 presets for `Closure Compiler`, `JSDoc3` and `JSDuck5`.
  *
- * By default it allows any tag of mixed set. You can pass `Object`
- * to select preset with `preset` field and add custom tags with `extra` field.
+ * By default it allows any tag of mixed set. You can pass `Object` to select
+ * preset with `preset` field and add custom tags with `extra` field.
```
hzoo: Is there another word for `tag` or is that the correct jsdoc terminology? Maybe `By default, it allows for any combination of tags?`

`to select preset` -> `to select a preset`?

lib/rules/jsdoc.js
```diff
@@ -209,8 +254,8 @@
  *
  * ### checkReturnTypes
  *
- * Reports discrepancies between the claimed in jsdoc and actual type if
- * both exist (code scan)
+ * Reports discrepancies between the claimed in jsdoc and actual type if both
+ * exist (code scan)
```
hzoo: `Checks for differences between the jsdoc and actual return types if both exist.`

lib/rules/jsdoc.js
```diff
@@ -336,11 +381,11 @@
  *
  * Reports invalid types for bunch of tags
  *
- * In `strictNativeCase` mode ensures that case of natives is the same as in
- * this list: `boolean`, `number`, `string`, `Object`, `Array`, `Date`, `RegExp`.
+ * In `strictNativeCase` mode ensures that case of natives is the same as in this
+ * list: `boolean`, `number`, `string`, `Object`, `Array`, `Date`, `RegExp`.
```
hzoo: In `strictNativeCase` mode -> The `strictNativeCase` mode

lib/rules/jsdoc.js
```diff
@@ -336,11 +381,11 @@
...
  *
- * In `capitalizedNativeCase` mode ensures that first letter in all native types and
- * primitives is uppercased except the case with `function` in google closure
+ * In `capitalizedNativeCase` mode ensures that first letter in all native types
```
hzoo: `In capitalizedNativeCase mode` -> `The capitalizedNativeCase mode`

`that first letter ` -> `that the first letter `

publish/helpers/plugins.js
```diff
@@ -0,0 +1,78 @@
+var fs = require('fs');
+var path = require('path');
+
+// yo
```
hzoo: extra comments?

lib/rules/jsdoc.js
zxqfox: > Is there another word for tag

According to http://usejsdoc.org/about-block-inline-tags.html tag is correct.

> Maybe `By default, it allows for any combination of tags?`

Does this meant that the user can use any tag of any existing preset? Atm preset field works like clarification. And by default it allows any tag of combined/merged tag set. Not sure that your version is more correct. But feels like the current one is even worse.

There is one more problem with this file because it's generated by the script at the end. If we want to make here changes we should do it also in jsdoc repo. Just fyi. ;-) Guess would be better to check the whole docs for grammar.

lib/rules/jsdoc.js
zxqfox: Guess, yes. But then we need to change it everywhere. Right?

lib/rules/jsdoc.js
hzoo: I guess I didn't understand what `mixed` set meant.

hzoo: I guess you are just saying it allows all the tags from the preset you picked as well as any you add in the extra field?

zxqfox: Mixed set is a merged set of all tag files (presets). Union in terms of Set theory.

PR #1806, opened 2015-09-24
New Rule: (disallow|require)MultiLineTernary

### LINKED ISSUE #1206: New Rule: (disallow|require)MultiLineTernary
For node-style-guide: https://github.com/felixge/node-style-guide#use-multi-line-ternary-operator

```js
var foo = (a === b)
  ? 1
  : 2;
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10086636-new-rule-disallow-require-multilineternary?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bjdixon:
New rules for disallowMultiLineTernary and requireMultiLineTernary

Fixes #1206

### DIFF
--- a/lib/rules/disallow-multi-line-ternary.js
+++ b/lib/rules/disallow-multi-line-ternary.js
@@ -1,0 +1,60 @@
+/**
+ * Disallows the test, consequent and alternative to be on separate lines when using the ternary operator.
+ *
+ * Types: `Boolean`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowMultiLineTernary": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var foo = (a === b) ? 1 : 2;
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var foo = (a === b)
+ *   ? 1
+ *   : 2;
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === true,
+            this.getOptionName() + ' option requires a true value or should be removed'
+        );
+    },
+
+    getOptionName: function() {
+        return 'disallowMultiLineTernary';
+    },
+
+    check: function(file, errors) {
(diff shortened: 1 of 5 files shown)


lib/rules/disallow-multi-line-ternary.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * Disallows the test, consequent and alternative to be on separate lines when using the ternary operator.
```
hzoo: alternative -> alternate

hzoo:
Nice job @bjdixon - the rule doesn't seem to bad with the new assert methods haha!

Maybe a test for nested ternaries? What should happen there?

bjdixon:
Good point @hzoo  I'll add some tests for nested ternaries. 

I assume disallowMultiLineTernary should be 1 line regardless of any nesting. 

```
var foo = (a === b) ? (x > y) ? 1 : 2 : (c === d) ? 3 : 4
```

For requireMultiLineTernary I think each nested ternary should be a line per test, consequent and alternate, no?

```
var foo = (a === b)
    ? (x > y) 
        ? 1
        : 2
    : (c === d)
        ? 3
        : 4
```
I wouldn't usually use nested ternaries and requireMultiLineTernary is not a style I'd personally employ, so guidance would definitely be appreciated =)

hzoo:
Yeah that makes sense to me.

Maybe we should make a separate rule to `disallowNestedTernaries`

bjdixon:
@hzoo if you want to go ahead with the separate rule for ``` disallowNestedTernaries ``` I'd be happy to implement that.

Is there something I need to do so that AppVeyor build doesn't fail?

hzoo:
We can just merge it since it's ok on travis and the error message is a timeout (something is taking too long) - just need another review

markelog:
LGTM

hzoo:
Thanks @bjdixon! Will release soon

PR #1807, opened 2015-09-24
Preset: try global node module if no local node module

nantunes:


hzoo:
@markelog was this not fixed before?

markelog:
@nantunes Could you explain what you trying to do here?

nantunes:
@markelog I've installed [jscs-preset-pentaho](https://www.npmjs.com/package/jscs-preset-pentaho) globally, so I don't need to install it locally in every project.

Without my patch doing ```jscs .``` results in:
```
Preset "pentaho" does not exist
```

Outputing the exception reveals:
```
[Error: Cannot find module 'jscs-pentaho' from '/Users/nantunes/dev']
[Error: Cannot find module 'jscs-preset-pentaho' from '/Users/nantunes/dev']
[Error: Cannot find module 'jscs-config-pentaho' from '/Users/nantunes/dev']
```

With my patch it does find it. However I noticed now that it does so because it starts looking from ```/usr/local/lib/node_modules/jscs/lib```.

So this is in fact related to substack/node-resolve#68, not JSCS. Sorry.

Should I close the PR or keep it as a pending issue?

hzoo:
If you find it is a jscs problem later we can reopen - I guess close for now?

nantunes:
Ok.

PR #1815, opened 2015-09-28
Chai

mdevils:
PR for Chai integration.

Around 3000 tests were automatically converted. 20 required manual fixes in `JSCS` core.

The quality of the tests is higher now, we check for errors more precisely.


lib/js-file.js
```diff
@@ -350,6 +350,8 @@ JsFile.prototype = {
                 return this._tokens[index];
             }
         } while (--index >= 0);
+
+        return null;
```
markelog: Is this covered by tests?

lib/rules/require-capitalized-comments.js
```diff
@@ -202,7 +202,7 @@ module.exports.prototype = {
         var firstToken = file.getFirstNodeToken(comment);
         var otherToken = file.getPrevToken(firstToken, { includeComments: true });
 
-        return otherToken.loc.start.line === firstToken.loc.start.line;
+        return otherToken ? otherToken.loc.start.line === firstToken.loc.start.line : false;
```
markelog: This is one of the bugs you were talking about?

package.json
```diff
@@ -65,15 +65,17 @@
     "chalk": "~1.1.0",
     "cli-table": "~0.3.1",
     "commander": "~2.8.1",
+    "cst": "0.0.11",
```
markelog: Is this used in here?

package.json
```diff
@@ -86,11 +88,12 @@
...
     "vow": "~0.4.8",
     "vow-fs": "~0.3.4",
-    "xmlbuilder": "^2.6.1",
-    "htmlparser2": "3.8.3"
+    "xmlbuilder": "^2.6.1"
   },
   "devDependencies": {
     "browserify": "^11.0.0",
+    "chai": "^3.3.0",
+    "chai-string": "^1.1.3",
```
markelog: Default functionality of the chai is not enough? like `match` or `string`?

test/lib/assertHelpers.js
```diff
@@ -36,13 +36,13 @@ var AssertHelpers = {
             });
 
             it('report', function() {
-                assert(checker.checkString(options.input).getErrorCount() === options.errors);
+                expect(checker.checkString(options.input)).to.have.error.count.equal(options.errors);
```
markelog: To have error count equal? It seems chai tries to emulate real language, whereas this sentence structure looks out of place

test/chai-extensions.js
```diff
@@ -0,0 +1,154 @@
+var chai = require('chai');
```
markelog: I would move this module to different package

test/mocha.opts
```diff
@@ -1,3 +1,5 @@
+test/chai-extensions.js
 test/specs/**/*.js
+--timeout 3000
```
markelog: in the next lines we use abbreviation flags, could you change them all then? Or use `-t`?

test/specs/checker.js
```diff
@@ -39,37 +39,36 @@ describe('checker', function() {
...
+                expect(errors[0].getErrorList().length).to.equal(0);
             });
         });
 
         it('should throw an exception when path doesn\'t exists', function() {
             return checker.checkPath('./test/non-exists').then(function() {
-                assert(false);
+                throw new Error();
             }, function() {
-                assert(true);
```
markelog: So this is an empty function now?

test/specs/cli.js
```diff
@@ -195,7 +195,7 @@ describe('cli', function() {
         });
 
         return result.promise.fail(function(status) {
-            assert(status);
+            expect(!!status).to.equal(true);
```
markelog: Was this done with automatic or you did it by hand?

test/specs/cli.js
```diff
@@ -181,7 +181,7 @@ describe('cli', function() {
             config: 'test/data/cli/cli.json'
         });
         return result.promise.then(function() {
-            assert(result.checker.getProcessedConfig().requireCurlyBraces);
+            expect(!!result.checker.getProcessedConfig().requireCurlyBraces).to.equal(true);
```
markelog: `to have property` would be better though :-)

markelog:
That's really impressive

test/mocha.opts
mdevils: Sure, no problem.

test/chai-extensions.js
mdevils: What do you mean? Create a new npm package for our internal assertions?

test/lib/assertHelpers.js
mdevils: Can you try and spell it better?

test/chai-extensions.js
markelog: It could be helpful for the plugin writers too, plus we could add tests for those assertions

test/specs/checker.js
mdevils: Yep, `assert(true)` didn't do anything at all.

test/specs/cli.js
mdevils: Automatic.

lib/rules/require-capitalized-comments.js
mdevils: Yep, there was an error here.

package.json
mdevils: I used `startsWith`. Couldn't use `match` with automatic test rewriting (i.e. converting to regexp is tricky).

test/chai-extensions.js
mdevils: I think we can do it as a next step, not necessary before merging this.

lib/js-file.js
mdevils: I don't think so

package.json
markelog: How about `string`?

package.json
mdevils: I will remove this before merging. I have it for the test rewriting tool.

test/chai-extensions.js
markelog: Sure thing, would you mind creating a ticket after this will land?

test/lib/assertHelpers.js
markelog: `.to.equal.errors`, /cc @zxqfox

test/specs/checker.js
markelog: I mean, can we remove second function?

test/specs/cli.js
mdevils: It was converted automatically (there are around 100 cases like this) and `to have property` does not match for most. Quite often you have that property, but with negative value.

test/chai-extensions.js
mdevils: Yes, sure.

package.json
mdevils: `string` I use in case if just "contain". But we have like 20 cases where we check if string starts with another string.

test/specs/checker.js
mdevils: Oh, yes, sure.

package.json
markelog: They can be replaced with `string`, i suspect that is not a popular use-case in other libraries too, that is why those functions are not included in main dist.

My concern is solely becase `chai-string` is not very popular module, which raise flag for me, since might not be stable enough

test/lib/assertHelpers.js
zxqfox: @markelog It works like: `expect(${ACTUAL}).to.${ACTION}(${EXPECTED})`. Not sure we can do it better ;-)

mdevils:
Merged to master.

test/specs/checker.js
mdevils: Fixed.

test/mocha.opts
mdevils: Fixed.

Issue #1831, opened 2015-10-02
False positive: disallowParenthesesAroundArrowParam rest-params


SimenB:
If the only arg is a rest-param (to get `arguments` as a real array)

Using 2.2.1

https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&code=export%20default%20(...stuff)%20%3D%3E%20%7B%0A%20%20if%20(stuff.length%20%3C%202)%20return%20null%3B%0A%0A%20%20verdier.splice(-1%2C%201)%3B%0A%0A%20%20return%20verdier.reduce((streng%2C%20memo)%20%3D%3E%20%60%24%7Bmemo%7D%20%24%7BescapeExpression(streng)%7D%60%2C%20'')%3B%0A%7D%3B

hzoo:
Thought I checked for that :cry: - looks like I only added it for the require rule lol - https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/require-parentheses-around-arrow-param.js#L51

PR #1832, opened 2015-10-02
Fix: `disallowParenthesesAroundArrowParam` - account for non-identifiers (RestElement, ArrayPattern) correctly

hzoo:
@SimenB 

#1831

### DIFF
--- a/lib/rules/disallow-parentheses-around-arrow-param.js
+++ b/lib/rules/disallow-parentheses-around-arrow-param.js
@@ -58,21 +58,22 @@ module.exports.prototype = {
             if (node.params.length !== 1) {
                 return;
             }
+            var firstParam = node.params[0];
 
             // Old Esprima
             var hasDefaultParameter = node.defaults && node.defaults.length === 1;
             // ESTree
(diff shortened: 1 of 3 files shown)


lib/rules/disallow-parentheses-around-arrow-param.js
```diff
@@ -58,21 +58,22 @@ module.exports.prototype = {
...
+            var firstParam = node.params[0];
 
             // Old Esprima
             var hasDefaultParameter = node.defaults && node.defaults.length === 1;
             // ESTree
-            var hasDefaultParameterESTree = node.params[0].type === 'AssignmentPattern';
-            var hasDestructuring = node.params[0].type === 'ObjectPattern';
+            var hasDefaultParameterESTree = firstParam.type === 'AssignmentPattern';
+            var hasDestructuring = firstParam.type === 'ObjectPattern' || firstParam.type === 'ArrayPattern';
+            var hasRestElement = firstParam.type === 'RestElement';
```
hzoo: Maybe it would be easier to just check if the first param is not an identifier (and return) instead of all the checks?

SimenB:
This works for my code :smile: 
No comment on the implementation, I don't know AST enough

PR #1837, opened 2015-10-05
requireDotNotation: fix non-latin bug by improving identifier support

### LINKED ISSUE #920: requireDotNotation breaks for non-latin variable name
Example (copy to the [demo](http://jscs.info/demo.html), jQuery preset)
```
obj["π"] = 1;
obj["test"] = 2;
```
The requireDotNotation error is only reported for the last line, but should also be reported for the first.

Reason for the bug is in
https://github.com/jscs-dev/node-jscs/blob/master/lib/utils.js#L66

The RegExp
```
/^[a-zA-Z_$][a-zA-Z0-9_$]*$/;
```
is not describing all possible JavaScript identifiers see [section 7.6 in the ES5 spec](http://es5.github.io/#x7.6).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7757712-requiredotnotation-breaks-for-non-latin-variable-name?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wachunga:
The regex is not pretty. We should consider using a library instead,
eg https://github.com/joelarson4/CharFunk.
This fix revealed an incorrect test case: `while` is not a valid identifier
so it needs to be made a literal and should not cause an error.

Fixes #920


alawatthe:
I'm a bit surprised, that we don't do this correctly already... So I think we should do this.

But I don't like to have the giant regexp in the code, as there is no way to check if it is correct and debugging would be a nightmare. We already have a dependency on regenerator, so it should be easy to do something like: https://gist.github.com/mathiasbynens/6334847

wachunga:
Completely agree that the regex is gross. I didn't see the existing dependency on `regenerate`... I'll give it a whirl.

wachunga:
I can't repro that build failure on node 0.10.40 or 4.1.2. I assume it's a flaky test that times out sometimes.

hzoo:
Yeah don't worry about the time out ones - either flaky or we have a perf issue - we should figure that out / increase the timeout.

hzoo:
Don't want to leave this PR hanging but I think someone else should review since I don't know much about it - although tests look good

alawatthe:
Sorry, I meant to review this sooner :-/

The [getDialect method](https://github.com/jscs-dev/node-jscs/blob/master/lib/js-file.js#L677-L687) can return 'es3', 'es5' and 'es6', but you are only handling 'es5' and 'es6', see [here](https://github.com/wachunga/node-jscs/commit/61722ddba25c61cfbf94b9e9a563d2c2317158c0#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR58). So I think we should also have an ```generateIdentifierRegexES3``` method, which is generating the appropriate ES3 Regexes.

Apart from that, everything looks good to me.

hzoo:
Nice work!

PR #1843, opened 2015-10-06
New Rule: disallowNestedTernaries

### LINKED ISSUE #1824: New Rule: disallowNestedTernaries
disallowNestedTernaries/disallowNestedConditionalExpressions

Don't know if we want to make a parameter for a certain maxLevel of nesting since that might be too much right now?

Ref https://github.com/jscs-dev/node-jscs/pull/1806#issuecomment-142786850
@bjdixon

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27069749-new-rule-disallownestedternaries?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bjdixon:
Fixes #1824

### DIFF
--- a/lib/rules/disallow-nested-ternaries.js
+++ b/lib/rules/disallow-nested-ternaries.js
@@ -76,13 +76,14 @@ module.exports.prototype = {
             var getLevel = function(currentNode) {
                 if (currentNode.parentNode && currentNode.parentNode.type === 'ConditionalExpression') {
                     level += 1;
+                    if (level > maxLevel) {
+                        errors.add('Illegal nested ternary', node.loc.start.line, node.loc.start.column);
+                        return;
+                    }
                     getLevel(currentNode.parentNode);
                 }
             };
             getLevel(node);
-            if (level > maxLevel) {
-                errors.add('Illegal nested ternary', node.loc.start.line, node.loc.start.column);
-            }
         });
     }
 
--- a/lib/rules/disallow-nested-ternaries.js
+++ b/lib/rules/disallow-nested-ternaries.js
@@ -1,0 +1,89 @@
+/**
+ * Disallows nested ternaries.
+ *
+ * Types: `Boolean`, `Integer`
+ *
+ * Values: `true` or an Integer that describes the maximum levels of nesting to be allowed.
+ *
+ * #### Examples
+ *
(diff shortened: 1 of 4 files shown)


hzoo:
Wow awesome - looks pretty good! Nice idea to have a param for a max level of nesting. I guess you could also change it to maybe skip ConditionalExpressions that have already been visited due to nesting but it's probably not necessary as people don't normally nest that much

markelog:
When we will have CST, that would be even better for autofixing.

@bjdixon would you like us to ping you when it will happen?

lib/rules/disallow-nested-ternaries.js
```diff
@@ -0,0 +1,89 @@
...
+
+    getOptionName: function() {
+        return 'disallowNestedTernaries';
+    },
+
+    check: function(file, errors) {
+        var maxLevel = this._maxLevel;
+        file.iterateNodesByType('ConditionalExpression', function(node) {
+            var level = 0;
+            var getLevel = function(currentNode) {
```
markelog: Is this helper really needed?

markelog: Oh, i see, yeah, cool

bjdixon:
@hzoo Maybe I should return a value from ```getLevel``` and cache that? I think that might reduce the redundant walking, but it would introduce a little more complexity. I'm probably missing a simpler strategy that you have in mind. I'm happy to implement this differently. Let me know.

@markelog yeah, definitely ping me for any updates.

hzoo:
Yeah I figured maybe the complexity isn't worth it? Or, maybe it should exit early based on maxLevels so you don't need to check? Not sure

bjdixon:
Agreed, returning early would be better. There's really no point in checking past the boundary. I'll update the PR.

hzoo:
ok nice @markelog?

markelog:
LGTM

hzoo:
Awesome stuff again @bjdixon!

Issue #1845, opened 2015-10-06
Small doc typo in rule disallowMultipleVarDecl


ValYouW:
Hi,

The doc for rule `disallowMultipleVarDecl` states that the `allExcept` array can get a `required` value, while actually is should be `require`.
The doc:
https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-multiple-var-decl.js#L13

The code:
https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-multiple-var-decl.js#L99

alawatthe:
Thanks for the report @ValYouW, do you want to send a pull request for this?

PR #1847, opened 2015-10-06
Fix typo in disallowMultipleVarDecl rule

ValYouW:
The doc for rule disallowMultipleVarDecl states that the allExcept array can get a required value, while actually is should be require - fix that.


alawatthe:
Thank you :-)
By the way, if you write something like "fixes #1845" in the commit message, the issue gets updated with a link to this pull request.

ValYouW:
done :)

hzoo:
Cool thanks for the report/fix @ValYouW!

PR #1856, opened 2015-10-09
requireDollarBeforeJqueryAssignment: fix Only valiates the first key

### LINKED ISSUE #1688: requireDollarBeforeJqueryAssignment: Only validates the first key
Failing test:

```js
            it('should look at keys besides the first', function() {
                assert.equal(checker.checkString('var x = { bar: 1, foo: $(".foo") }').getErrorCount(), 1);
            });
```

Passing assertion:

```js
assert.equal(checker.checkString('var x = { foo: $(".foo") }').getErrorCount(), 1);
```

Problem line is here: https://github.com/jscs-dev/node-jscs/blob/7c2d0dd9d848cf0de4eeabb52a2514b480179956/lib/rules/require-dollar-before-jquery-assignment.js#L96

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25843048-requiredollarbeforejqueryassignment-only-validates-the-first-key?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bjdixon:
Added a loop to check each property on an object and stuffed the logic
that checks whether a varName should start with a dollar sign into a
helper function as we may be needing it more than once per
file.iterateNodesByType cycle.

Fixes #1688

### DIFF
--- a/lib/rules/require-dollar-before-jquery-assignment.js
+++ b/lib/rules/require-dollar-before-jquery-assignment.js
@@ -75,6 +75,41 @@ module.exports.prototype = {
             var varName;
             var right;
 
+            function checkIfVarNameShouldStartWithDollar(left, varName, right) {
+
+                if (varName.indexOf('$') === 0 || varName.indexOf('_$') === 0) {
+                    return;
+                }
+
+                if (!right || right.type !== 'CallExpression') {
+                    return;
+                }
+
+                var nextToken = file.getTokenByRangeStart(right.callee.range[0]);
+                if (nextToken.value !== '$') {
+                    return;
+                }
+
+                nextToken = file.getNextToken(nextToken);
+                if (nextToken.value !== '(') {
+                    return;
+                }
+
+                while (!(nextToken.type === 'Punctuator' && nextToken.value === ')')) {
+                    nextToken = file.getNextToken(nextToken);
+                }
+
+                nextToken = file.getNextToken(nextToken);
+
+                if (!nextToken || !(nextToken.type === 'Punctuator' && nextToken.value === '.')) {
+                    errors.add(
+                        'jQuery identifiers must start with a $',
+                        left.loc.start.line,
+                        left.loc.start.column
+                    );
+                }
+            }
+
             if (type === 'VariableDeclarator') {
                 if (token.id.type === 'ObjectPattern') {
                     return;
@@ -83,7 +118,7 @@ module.exports.prototype = {
                 left = token.id;
                 varName = left.name;
                 right = token.init;
-
+                checkIfVarNameShouldStartWithDollar(left, varName, right);
             } else if (ignoreProperties) {
                 return;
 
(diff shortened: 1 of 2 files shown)


lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -75,6 +75,41 @@ module.exports.prototype = {
             var varName;
             var right;
 
+            function checkIfVarNameShouldStartWithDollar(left, varName, right) {
+
+                if (varName.indexOf('$') === 0 || varName.indexOf('_$') === 0) {
```
mrjoelkemp: Non-blocker: a regex for the dollar matchers would combine conditions

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -75,6 +75,41 @@ module.exports.prototype = {
             var varName;
             var right;
 
+            function checkIfVarNameShouldStartWithDollar(left, varName, right) {
```
mrjoelkemp: Why is var not first in the list? The function name talks about var, not left or right

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -75,6 +75,41 @@ module.exports.prototype = {
...
+
+                if (varName.indexOf('$') === 0 || varName.indexOf('_$') === 0) {
+                    return;
+                }
+
+                if (!right || right.type !== 'CallExpression') {
```
mrjoelkemp: Is there a quick code sample where the right is not there? Either a test or comment would be good. The test is preferable though

lib/rules/require-dollar-before-jquery-assignment.js
```diff
@@ -75,6 +75,41 @@ module.exports.prototype = {
...
+                    return;
+                }
+
+                while (!(nextToken.type === 'Punctuator' && nextToken.value === ')')) {
+                    nextToken = file.getNextToken(nextToken);
+                }
+
+                nextToken = file.getNextToken(nextToken);
+
+                if (!nextToken || !(nextToken.type === 'Punctuator' && nextToken.value === '.')) {
```
mrjoelkemp: Test or comment for the case when nextToken is not defined?

mrjoelkemp:
Lgtm. Great work @bjdixon! 

I know the body of that function was an extraction of the same code, but the comments (if possible) would help readability of the rule. It's hard to visualize programs that satisfy parts of the code.

lib/rules/require-dollar-before-jquery-assignment.js
bjdixon: There is one test already that should cover this case:
```
it('should not report for loops', function() {
    expect(checker.checkString('for (var prop in rawVars) {}')).to.have.no.errors();
});
```
But I've added another that hopefully adds some clarity:
```
it('should not report for var declaration without assignment', function() {
    expect(checker.checkString('var x;')).to.have.no.errors();
});
```

lib/rules/require-dollar-before-jquery-assignment.js
bjdixon: I can't think of a condition where it get's to this point with ``` !nextToken ``` being true. If you can give me a clue, I'm happy to add the test =)

bjdixon:
Thanks for the review @mrjoelkemp! I've updated the PR to comply with most of your comments. However, I'm struggling in coming up with the conditions for the last test mentioned. Happy to update the PR again if you or anyone else can point me in the right direction.

mrjoelkemp:
:+1: lgtm

hzoo:
Awesome work again @bjdixon!

PR #1857, opened 2015-10-09
disallowSpaceAfterObjectKeys: fix #1818

### LINKED ISSUE #1818: disallowSpaceAfterObjectKeys - aligned: "Illegal space after key" without spaces
Input:
```javascript
var f = {
    "name": 1,
    "x": 2
};
```

Config:
```json
{
    "disallowSpaceAfterObjectKeys": {
        "allExcept": ["aligned"]
    }
}
```

output:
```
Illegal space after key at test.js :
     1 |var f = {
     2 |    "name": 1,
     3 |    "x": 2
---------------^
     4 |};
     5 |

1 code style error found.
```

There is no space after the key, so I would expect that to be valid, ie. "aligned" should only matter when there're spaces so that the colons are aligned.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27027658-disallowspaceafterobjectkeys-aligned-illegal-space-after-key-without-spaces?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dreef3:
Here's my attempt to fix #1818. Adds the following valid case to `disallowSpaceAfterObjectKeys` with `aligned` option:

#### Config
```javascript
disallowSpaceAfterObjectKeys: { allExcept: ['aligned'] }
```

#### Valid
```javascript
var f = {
    "name": 1,
    "x": 2
};
```

Maybe code that counts spaces could be moved to some utils file but I'm not sure where it should belong.

### DIFF
--- a/lib/rules/disallow-space-after-object-keys.js
+++ b/lib/rules/disallow-space-after-object-keys.js
@@ -163,9 +163,25 @@ module.exports.prototype = {
                 tokens.push(keyToken);
             });
 
+            var noSpace = true;
+            if (exceptAligned) {
+                var withoutSpace = 0;
+                var alignedOnColon = 0;
+                tokens.forEach(function(key) {
+                    var colon = file.getNextToken(key);
+                    var spaces = Math.abs(colon.range[0] - key.range[1]);
+                    if (spaces === 0) {
+                        withoutSpace++;
+                    } else if (spaces === maxKeyEndPos - key.loc.end.column) {
+                        alignedOnColon++;
+                    }
+                });
+
+                noSpace = withoutSpace > alignedOnColon;
+            }
+
             tokens.forEach(function(key) {
                 var colon = file.getNextToken(key);
-                var noSpace = Math.abs(colon.range[0] - key.range[1]) === 0;
                 var spaces = (exceptAligned && !noSpace) ? maxKeyEndPos - key.loc.end.column : 0;
                 errors.assert.spacesBetween({
                     token: key,
--- a/lib/rules/disallow-space-after-object-keys.js
+++ b/lib/rules/disallow-space-after-object-keys.js
@@ -165,7 +165,8 @@ module.exports.prototype = {
 
             tokens.forEach(function(key) {
                 var colon = file.getNextToken(key);
-                var spaces = exceptAligned ? maxKeyEndPos - key.loc.end.column : 0;
+                var noSpace = Math.abs(colon.range[0] - key.range[1]) === 0;
+                var spaces = (exceptAligned && !noSpace) ? maxKeyEndPos - key.loc.end.column : 0;
                 errors.assert.spacesBetween({
                     token: key,
                     nextToken: colon,
--- a/test/specs/rules/disallow-space-after-object-keys.js
+++ b/test/specs/rules/disallow-space-after-object-keys.js
@@ -159,6 +159,16 @@ describe('rules/disallow-space-after-object-keys', function() {
                         '};'
                     )).to.have.no.errors();
                 });
+
+                it('should report objects with both keys without spaces and aligned on colon #1818', function() {
+                    expect(checker.checkString(
+                      'var f = {\n' +
+                      '  "n": 1,\n' +
+                      '  "xasdf"   : 2,\n' +
+                      '  "fyfyasdf": 0\n' +
+                      '};'
+                    )).to.have.error.count.equal(1);
+                });
             });
 
             describe('with method value', function() {
--- a/test/specs/rules/disallow-space-after-object-keys.js
+++ b/test/specs/rules/disallow-space-after-object-keys.js
@@ -150,6 +150,15 @@ describe('rules/disallow-space-after-object-keys', function() {
                         '};'
                     )).to.have.no.errors();
                 });
+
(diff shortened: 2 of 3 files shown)


hzoo:
Looks good to me. @davidrapin @avdv - does this PR work for yall?

davidrapin:
Perfect :)

avdv:
I think this would erroneously permit:

```javascript
var f = {
    "name": 1,
    "x"            : 2
};
```
because you're completely disabling the check for individual lines without spaces. You would have to check whether all colons on all lines are at the same column or none of the lines actually has a space after the key.

dreef3:
@avdv ,

I tried to fix your case. Here's one more that arises if we allow either all keys without space or all keys aligned on colon in a single object:

##### Invalid
```javascript
var x = {
  a : 1,
  foo : function() {},
  bcd : 2
};
```

##### Valid (no space after all keys)
```javascript
var x = {
  a: 1,
  foo: function() {},
  bcd: 2
};
```

##### Valid (aligned on colon)
```javascript
var x = {
  a  : 1,
  foo: function() {},
  bcd: 2
};
```

I propose to restrict `aligned` option to keys aligned by colon only as it was before. Another reason could be that these two styles of alignment can be treated as mutually exclusive and you always can use the rule without `aligned` to get keys without spaces after them.

Another option would be to always prefer second variant with colons aligned in cases like above.

davidrapin:
@dreef3, restricting `aligned` to aligned-only is not acceptable solution imho. As I explain in #1818, this would not be consistent with the semantic of the `allExcept` option. 

In this case the user wants to disallow all spaces between the key name and `:`, **except when the `:` are aligned**.

Real-world cases are mostly mixed between cases where colons are aligned and cases without spaces before keys at all because aligning colons wouldn't make sense in that particular case.

Is there a particular difficulty in implementing this?

dreef3:
@davidrapin,

I implemented the desired behavior in pull request except the case when it's unclear if object should have all keys without spaces or it should have colons aligned. Please see my previous comment for example. Right now such object would be treated as "no space after keys" case.

davidrapin:
@dreef3 sorry, I misunderstood the issue. 

I'm actually okay with example (1) being invalid ("no space allowed after key"), example (2) being valid (reason behind #1818) and example (3) being valid (no regression).

dreef3:
@davidrapin, @avdv , @hzoo 

As I understand the way JSCS works it will treat only one of these two valid examples as valid at the same time. So right now example 3 is not valid anymore.

avdv:
@dreef3 LGTM

Just for the sake of nit-picking, the implementation seems a bit overly complicated now and one doesn't easily see what's going on. Here's what I wrote last evening (too late to make a PR):

```javascript
        var exceptSingleline = this._exceptSingleline;
        var exceptMultiline = this._exceptMultiline;
        var exceptAligned = this._exceptAligned;
        var exceptMethod = this._exceptMethod;

            var multiline = node.loc.start.line !== node.loc.end.line;
            if (exceptSingleline && !multiline) {
                return;
            }
            if (exceptMultiline && multiline) {
                return;
            }

            var maxKeyEndPos = 0;
            var colonPos = -1;
            var isAligned = true;
            var tokens = [];
                if (property.shorthand || property.kind !== 'init' ||
                    (exceptMethod && property.method) ||
                    node.type === 'SpreadProperty') {
                    return;
                }

                var keyToken = file.getLastNodeToken(property.key);
                if (property.computed === true) {
                    keyToken = file.getNextToken(keyToken);
                }

                if (exceptAligned) {
                    maxKeyEndPos = Math.max(maxKeyEndPos, keyToken.loc.end.column);
                    var colon = file.getNextToken(keyToken);
                    if (colonPos < 0) {
                        colonPos = colon.loc.start.column;
                    } else if (colon.loc.start.column !== colonPos) {
                        isAligned = false;
                    }
                }
                tokens.push(keyToken);
            });

            if (exceptAligned && isAligned && maxKeyEndPos === colonPos) {
                return;
            }

            tokens.forEach(function(key) {
                var colon = file.getNextToken(key);

                errors.assert.spacesBetween({
                    token: key,
                    nextToken: colon,
                    exactly: 0,
                    message: 'Illegal space after key',
                    disallowNewLine: true
                });
            });
```

IMO, this way it's easier to see what the exception of the rule is (so, no checking needed) and the "no space after key" no longer is an exemption of the exemption.

markelog:
> Oh, just in case I wasn't clear enough, your change is fine for me as is. 

Cool then :-)

hzoo:
Going to release today - if you want to refactor @dreef3 /@avdv to be clearer later send another PR! Thanks

avdv:
> if you want to refactor @dreef3 /@avdv to be clearer later send another PR if we haven't already merged

@dreef3 if you want, you can use my code and update your PR with it; that would be OK with me. Either way, thanks for your fix. Looking forward for the release...

hzoo:
Thanks @dreef3! Although crap, when I squashed it took the other account as the author

PR #1870, opened 2015-10-15
Fix: do not extract/lint html by default

hzoo:
Fixes #1828
Closes gh-1870

PR #1871, opened 2015-10-15
Fix: do not extract/lint html by default

### LINKED ISSUE #1828: disallowTabs fails with TypeError, Extracting HTML by default
```
TypeError: Cannot call method 'replace' of undefined
    at renderLine (/home/vagrant/www/node_modules/jscs/lib/errors.js:250:17)
    at Object.Errors.explainError (/home/vagrant/www/node_modules/jscs/lib/errors.js:181:13)
    at /home/vagrant/www/node_modules/jscs/lib/reporters/console.js:16:36
    at Array.forEach (native)
    at /home/vagrant/www/node_modules/jscs/lib/reporters/console.js:14:35
    at Array.forEach (native)
    at Object.module.exports [as writer] (/home/vagrant/www/node_modules/jscs/lib/reporters/console.js:9:22)
    at /home/vagrant/www/node_modules/jscs/lib/cli.js:151:18
    at Array.0 (/home/vagrant/www/node_modules/vow/lib/vow.js:711:56)
    at Object.callFns [as _onImmediate] (/home/vagrant/www/node_modules/vow/lib/vow.js:23:35)
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27081681-disallowtabs-fails-with-typeerror-extracting-html-by-default?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1828
Closes gh-1870

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -299,7 +299,7 @@ Values: A single file extension or an Array of file extensions, beginning with a
 
 Set list of glob patterns for files which embeded JavaScript should be checked.
 
-Type: `Array` or `false`
+Type: `Array` or `Boolean`
 
 Values: Array of file matching patterns
 
@@ -311,9 +311,9 @@ JavaScript extracting from files that doesn't match to [fileExtensions](#fileExt
 "extract": ["*.htm", "*.html"]
 ```
 
-#### Default
+#### Value `true`
 
-Default value is `["**/*.+(htm|html|xhtml)"]`. So JavaScript is extracted from files with `.htm`, `.html` or `.xhtml` extension.
+JavaScript is extracted from files with `.htm`, `.html` or `.xhtml` extension with value `true`.
 
 ### maxErrors
 Set the maximum number of errors to report (pass -1 or null to report all errors).
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -490,10 +490,6 @@ Configuration.prototype._processConfig = function(config) {
 
     if (options.hasOwnProperty('extract')) {
         this._loadExtract(options.extract);
-
-    // Set default masks if there is no presets that could define their own
-    } else if (!options.hasOwnProperty('preset')) {
-        this._loadExtract(this._defaultExtractFileMasks);
     }
 
     if (options.hasOwnProperty('fileExtensions')) {
@@ -745,12 +741,13 @@ Configuration.prototype._loadExcludedFiles = function(masks) {
  * @protected
  */
 Configuration.prototype._loadExtract = function(masks) {
-    if (masks === false) {
+    if (masks === true) {
+        masks = this._defaultExtractFileMasks;
+    } else if (masks === false) {
         masks = [];
     }
 
-    assert(Array.isArray(masks), '`extract` option should be array of strings or false');
-
+    assert(Array.isArray(masks), '`extract` option should be array of strings');
     this._extractFileMasks = masks.slice();
     this._extractFileMatchers = this._extractFileMasks.map(function(fileMask) {
         return new minimatch.Minimatch(path.resolve(this._basePath, fileMask), {
--- a/test/specs/checker.js
+++ b/test/specs/checker.js
@@ -131,20 +131,7 @@ describe('checker', function() {
     });
 
     describe('extract', function() {
-        it('should extract from *.htm, *.html and *.xhtml by default', function() {
-            return checker.checkPath('./test/data/extract').then(function(errors) {
-                expect(errors.length).to.equal(3);
-                expect(errors[0].getErrorList().length).to.equal(2);
-                expect(errors[1].getErrorList().length).to.equal(2);
-                expect(errors[2].getErrorList().length).to.equal(0);
-            });
-        });
-
-        it('should not extract js when set to false', function() {
-            checker.configure({
-                extract: false
-            });
-
+        it('should not extract anything by default', function() {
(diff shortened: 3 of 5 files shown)


hzoo:
@lahmatiy turns off extract by default and removes boolean option.

test/specs/checker.js
```diff
@@ -177,16 +173,6 @@ describe('checker', function() {
             });
         });
 
-        it('extractFile should return null if file doesn\'t check', function() {
-            checker.configure({
-                extract: false
```
lahmatiy: this test should be ok, just remove `checker.configure`

lib/config/configuration.js
```diff
@@ -489,11 +489,8 @@ Configuration.prototype._processConfig = function(config) {
     }
 
     if (options.hasOwnProperty('extract')) {
-        this._loadExtract(options.extract);
-
-    // Set default masks if there is no presets that could define their own
-    } else if (!options.hasOwnProperty('preset')) {
-        this._loadExtract(this._defaultExtractFileMasks);
+        assert(Array.isArray(options.extract), '`extract` option requires array value');
+        this._loadExtract(options.extract.concat(['**/*.+(htm|html|xhtml)']));
```
lahmatiy: It's wrong. Should pass `options.extract` as is, with no extra.

lahmatiy: Moreover, this mask stores in `_defaultExtractFileMasks`

lib/config/configuration.js
```diff
@@ -745,11 +742,7 @@ Configuration.prototype._loadExcludedFiles = function(masks) {
  * @protected
  */
 Configuration.prototype._loadExtract = function(masks) {
-    if (masks === false) {
```
lahmatiy: I think should be possible to pass `true` value to use default mask for html files.

lahmatiy:
> Default value is ["**/*.+(htm|html|xhtml)"]

Also should be removed from docs

test/specs/extract-js.js
```diff
@@ -38,7 +38,8 @@ describe('extract-js', function() {
     describe('whitespaces', function() {
         it('should be no errors on disallowTrailingWhitespace', function() {
             checker.configure({
-                disallowTrailingWhitespace: true
+                disallowTrailingWhitespace: true,
+                extract: []
```
lahmatiy: Looks strange. `extract: true` would be better

hzoo:
If they specify an array at all - would we want to add `.html`?
Although If we only support html anyway might as well only allow true

hzoo:
@lahmatiy ok made changes

lahmatiy:
> Although If we only support html anyway might as well only allow true

Nope. For example your project can has templates or other html file with extensions other than defaults.

lib/config/configuration.js
```diff
@@ -745,12 +741,15 @@ Configuration.prototype._loadExcludedFiles = function(masks) {
...
+    if (masks === true) {
+        masks = this._defaultExtractFileMasks;
+    } else if (masks === false) {
         masks = [];
+    } else {
+        assert(Array.isArray(masks), '`extract` option should be array of strings');
+        masks = masks.concat(this._defaultExtractFileMasks);
```
lahmatiy: You still concat default masks. It's not a good idea. Left users to specify exact set of masks.

hzoo: So you're saying we should only use the defaults for value `true`?

OVERVIEW.md
```diff
@@ -313,7 +313,7 @@ JavaScript extracting from files that doesn't match to [fileExtensions](#fileExt
 
 #### Default
 
-Default value is `["**/*.+(htm|html|xhtml)"]`. So JavaScript is extracted from files with `.htm`, `.html` or `.xhtml` extension.
+JavaScript is extracted from files with `.htm`, `.html` or `.xhtml` extension by default. If an array is passed those extensions are concatenated.
```
lahmatiy: It's not correct. By default it doesn't extract anything.
Array contains file masks, not extensions.
`true` value is use default html mask.

lib/config/configuration.js
lahmatiy: Yes, exactly.

lahmatiy:
> we only support html

It's for now. But I believe someday it will be expanded for other formats.

hzoo:
@lahmatiy ok changes made

test/specs/config/configuration.js
```diff
@@ -354,21 +354,39 @@ describe('config/configuration', function() {
     });
 
     describe('shouldExtractFile', function() {
-        it('should be check *.htm, *.html, *.xhtml by default', function() {
+        it('should not check anything by default', function() {
             configuration.load({});
+            expect(!!configuration.shouldExtractFile('file.htm')).to.equal(false);
```
lahmatiy: `shouldExtractFile` always returns boolean, maybe remove `!!` and `!`?

lahmatiy:
Looks good :+1:

test/specs/config/configuration.js
hzoo: yeah otherwise looks confusing

PR #1872, opened 2015-10-16
Misc: fix various typos

wachunga:


hzoo:
Thanks a lot @wachunga - we can always use the help!

PR #1874, opened 2015-10-16
Remove duplicate entry in airbnb.json

dduugg:
The entry also appears on line 37 (-> 36).


hzoo:
Thanks for the report/PR!

PR #1876, opened 2015-10-17
Misc: increase mocha timeout

hzoo:
Not sure how much

Pretty annoying when docs changes cause failures https://travis-ci.org/jscs-dev/node-jscs/builds/85757477

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,5 +1,5 @@
 test/chai-extensions.js
 test/specs/**/*.js
--t 3000
+-t 4000
 -R progress
 -u bdd


hzoo:
The irony is that the PR failed so that means 3500 isn't enough

mrjoelkemp:
Do we know what's actually taking a long time during these tests? Otherwise, double it and call it a day?

hzoo:
Something with autofixing is taking a long time

hzoo:
Ok I didn't know about https://mochajs.org/#suite-level - we should add that for autofixing tests if we can

PR #1888, opened 2015-10-19
Fix: `disallowSpaceAfterKeywords` - fix issue with using default keyword list

### LINKED ISSUE #1885: `return` should not be in the list of `spacedKeywords`, `function` also has an issue
Hello,

I'm running JSCS with this rule enabled:
`disallowSpaceAfterKeywords: true`.

As such, JSCS is telling me that this:
`return obj`
should be:
`returnobj`.

Which is obviously invalid.
I expect setting `disallowSpaceAfterKeywords` to `true` to not affect `return` statements.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27517520-return-should-not-be-in-the-list-of-spacedkeywords-function-also-has-an-issue?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Fixes #1885
Closes gh-1888

cc @Francisc

### DIFF
--- a/lib/rules/disallow-space-after-keywords.js
+++ b/lib/rules/disallow-space-after-keywords.js
@@ -65,8 +65,8 @@ module.exports.prototype = {
         file.iterateTokensByTypeAndValue('Keyword', this._keywords, function(token) {
             var nextToken = file.getNextToken(token);
 
-            // Make an exception if the next token is also a keyword (a space would be required).
-            if (nextToken.type === 'Keyword') {
+            // Make an exception if the next token is not a Punctuator such as a Keyword or Identifier
+            if (nextToken.type !== 'Punctuator') {
                 return;
             }
             errors.assert.noWhitespaceBetween({
--- a/lib/rules/require-space-after-keywords.js
+++ b/lib/rules/require-space-after-keywords.js
@@ -29,7 +29,9 @@
  * ##### Valid
  *
  * ```js
- * return true;
+ * if (x) {
+ *     x++;
+ * }
  * ```
  *
  * ##### Invalid
--- a/test/specs/rules/disallow-space-after-keywords.js
+++ b/test/specs/rules/disallow-space-after-keywords.js
@@ -3,6 +3,7 @@ var expect = require('chai').expect;
 
 describe('rules/disallow-space-after-keywords', function() {
     var checker;
+    var rule = 'disallowSpaceAfterKeywords';
 
     beforeEach(function() {
         checker = new Checker();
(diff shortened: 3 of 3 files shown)


lib/rules/disallow-space-after-keywords.js
```diff
@@ -65,8 +67,8 @@ module.exports.prototype = {
         file.iterateTokensByTypeAndValue('Keyword', this._keywords, function(token) {
             var nextToken = file.getNextToken(token);
 
-            // Make an exception if the next token is also a keyword (a space would be required).
-            if (nextToken.type === 'Keyword') {
+            // Make an exception if the next token is not a Punctuator such as a Keyword or Identifier
+            if (nextToken.type !== 'Punctuator') {
```
hzoo: The previous fix for `else if` was adding `nextToken.type === 'Keyword'`

I was going to check `nextToken.type == 'Identifier'` but I figured maybe this could be better?

@mikesherov

lib/rules/disallow-space-after-keywords.js
```diff
@@ -51,7 +51,9 @@ module.exports.prototype = {
         );
 
         if (keywords === true) {
-            keywords = defaultKeywords;
+            keywords = defaultKeywords.filter(function(keyword) {
+                return ['function', 'return'].indexOf(keyword) === -1;
```
gibson042: I would say it's more complicated than this... some keywords require immediately following parentheses (`if`, `for`, `catch`, etc.), intervening whitespace of which can always be removed. Others (`return`, `throw`, `function`, etc.) require whitespace only when necessary to separate them from keywords/identifiers (so `return[]` and `function(){}` are fine but `thrownewError()` and `returnfoo` and `functionbar(){}` aren't). I think you want to leverage and maybe extend/expose more of https://github.com/hzoo/node-jscs/blob/1885/lib/token-categorizer.js , which also deals with such cases (cf. its top-level regular expressions). But you _definitely_ don't want to ignore these keywords outright.

hzoo: Oops actually I mean't to remove this after changing `nextToken.type !== 'Punctuator'`

hzoo:
@gibson042 what do you think after removing that snippet?

lib/rules/disallow-space-after-keywords.js
hzoo: Although actually - maybe it shouldn't check `function` since we have different rules to check function spacing - and other stuff like `function *` which should be a rule

hzoo: Yeah even in the example it only has a few keywords (the ones that initially make sense) - not every single keyword. `if, else, for, while, do, switch, try, catch, finally` sound good

gibson042:
I think there are still a few edge cases around [ASI-restricted keywords](http://www.ecma-international.org/ecma-262/6.0/#sec-rules-of-automatic-semicolon-insertion) and EOF like
```js
// false positive: keyword, line terminator, punctuator
yield
"ignored statement with leading punctuator"
```
and
```js
// false negative: keyword, whitespace, line terminator or EOF, identifier or keyword or EOF
break	// ← tab
undeclared
```

…but they may never come up.

PR #1895, opened 2015-10-20
disallowSpaceAfterComma: New option

### LINKED ISSUE #1889: Missing `disallowSpaceAfterComma`
Hello,

There's `disallowSpaceBeforeComma`, `requireSpaceBeforeComma` and `requireSpaceAfterComma`.
However, there isn't a `disallowSpaceAfterComma` property.

As an example, I would like to be able to make this invalid:
```
[1, 2, 3]
```
And his valid:
```
[1,2,3]
```

Is there a way to validate this given current options?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27522904-missing-disallowspaceaftercomma?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Francisc:
Disallows spaces after commas, for example `[1, 2, 3]` is invalid and
`[1,2,3]` is valid.

Fixes #1889

### DIFF
--- a/grouping.json
+++ b/grouping.json
@@ -22,6 +22,7 @@
     "validateLineBreaks",
     "validateIndentation",
     "disallowSpaceBeforeComma",
+    "disallowSpaceAfterComma",
     "disallowSpaceBeforeSemicolon",
     "disallowPaddingNewLinesAfterBlocks",
     "disallowPaddingNewLinesAfterUseStrict",
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -1066,6 +1066,7 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/disallow-space-before-semicolon'));
 
     this.registerRule(require('../rules/disallow-space-before-comma'));
+    this.registerRule(require('../rules/disallow-space-after-comma'));
 
     this.registerRule(require('../rules/require-space-before-comma'));
     this.registerRule(require('../rules/require-space-after-comma'));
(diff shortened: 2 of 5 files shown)


hzoo:
I think this should of been asked in the other rules - but can you add a test for new lines between commas instead of a space? We could also have a comment in between. Otherwise looks good

Francisc:
OK, I'll do that now.

Francisc:
Newline char is allowed by this rule.
Which makes sense since I'm looking to disallow the space and tab chars.
Do you want me to add a test that validates that newline characters are OK?

hzoo:
Yeah just checks that it doesn't error or something

Francisc:
Hope this is better.
I've also updated the tests for `disallowSpaceBeforeComma`.

hzoo:
Yeah we just want to cover as many cases as we can think of to catch any potential bugs - looks good. I'm probably going to wait on merging new rules but this should be part of 2.4.0 which we should release before next monday at least.

Francisc:
Alright, thanks.

hzoo:
Check out our https://github.com/jscs-dev/node-jscs/blob/master/CONTRIBUTING.md and other rules if you want to understand how JSCS works - it's usually pretty simple to add rules. I usually use astexplorer.net to check the code's AST

Francisc:
I did read the CONTRIBUTING.md file! :)
Thanks for the http://astexplorer.net, didn't know about it.
Should come in handy.

hzoo:
Thanks again!

PR #1901, opened 2015-10-20
requireAlignedMultilineParams: new rule

### LINKED ISSUE #1894: Feature Request: requireAlignedMultilineParams
Currently it doesn't seem like there's a way to enforce indentation of parameters when a function takes up multiple lines. This rule would attempt to address the issue. I have a working implementation, and am posting this mainly so that I can see if it's something there's interest in, and for any feedback.

Option can be one of:
  - true
  - a number
  - "firstParam"

I've broken down the three cases, to show what's valid/invalid for each:

### when option === true

Parameters that continue past the first line of a function definition must be aligned with the body of the function. For example:

```javascript
function addThreeNumbers(one,
  two,
  three) {
  return one + two + three;
}
```

would be considered valid, as would

```javascript
function addThreeNumbers(one,
    two, three) {
    return one + two + three;
}
```

Note that there are no restrictions on the number of parameters per line, just that the first parameter on each line from the second on, is aligned with the body. This also applies for anonymous functions passed as an argument:

```javascript
beforeEach(inject(function(one, two, three,
    four, five, six,
    seven) {
    one = 1;
    // Do some more stuff
}));
```

Below is an example of code that would NOT pass the check

```javascript
function addThreeNumbers(one,
    two,
    three) {
  return one + two + three;
}
```

It will not pass, because the parameters are indented two spaces past the body of the function.

### when typeof option === 'number'

You can also pass a number to the rule, which will be the number of columns you want the params to be shifted right, relative to the body of the function. For example, if option is 2, the following is valid:

```javascript
function addThreeNumbers(one,
    two,
    three) {
  return one + two + three;
}
```

Note that passing 0, is also the same as passing true, which I'm not a huge fan of.

### when option === 'firstParam'

Here, each parameter of a newline must be in the same column as the first parameter. For example:

```javascript
function addThreeNumbers(one,
                                             two, three) {
  return one + two + three;
}
```

would be valid, while

```javascript
function addThreeNumbers(one,
  two,
  three) {
  return one + two + three;
}
```

would not.

Thanks for taking the time to read this. I hope we can work on getting a rule like this merged, and would appreciate any feedback you have on if you think it'll be useful, any functionality improvements, as well as your thoughts on the options.

Cheers

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27535606-feature-request-requirealignedmultilineparams?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JGulbronson:
enforce indentation of parameters in multiline functions

Closes gh-1896
Fixes #1894

Currently, it seems there's a gap in jscs with enforcing parameter alignment. The closest rule I could find is `validateAlignedFunctionParameters`, though the name doesn't seem to be an accurate description of what it does, and it isn't very customizable. I'm proposing an additional rule, which, when used along with `validateAlignedFunctionParameters` would allow users greater control over the indentation of their code.

### DIFF
--- a/lib/rules/require-aligned-multiline-params.js
+++ b/lib/rules/require-aligned-multiline-params.js
@@ -44,6 +44,16 @@
  * };
  * ```
  *
+ * ##### Invalid for `0`
+ *
+ * ```js
+ * var test = function(one, two,
+ *     three, four, five,
+ *     six, seven, eight) {
+ *   console.log(a);
+ * };
+ * ```
+ *
  */
 
 var assert = require('assert');
--- a/lib/rules/require-aligned-multiline-params.js
+++ b/lib/rules/require-aligned-multiline-params.js
@@ -1,0 +1,117 @@
+/**
+ * Enforces indentation of parameters in multiline functions
+ *
+ * Types: `Boolean`, `String`, `Number`
+ *
+ * Values:
+ *  - `true` to require parameters are aligned with the body of the function
+ *  - `'firstParam'` to require parameters to be aligned with the first parameter
+ *  - `Number` for the number of columns the parameters should be indented past the function body
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowSpaceBeforeComma": true
+ * ```
+ *
+ * ##### Valid for `true`
+ *
+ * ```js
+ * var test = function(one, two,
+ *   three, four, five,
+ *   six, seven, eight) {
+ *   console.log(a);
+ * };
+ * ```
+ *
+ * ##### Valid for `2`
+ *
+ * ```js
+ * var test = function(one, two,
+ *     three, four, five,
+ *     six, seven, eight) {
+ *   console.log(a);
+ * };
+ * ```
+ *
+ * ##### Valid for `'firstParam'`
+ *
+ * ```js
+ * var test = function(one, two,
+ *                     three, four, five,
+ *                     six, seven, eight) {
+ *   console.log(a);
+ * };
+ * ```
+ *
+ */
+
+var assert = require('assert');
+
+module.exports = function() {
+};
+
+module.exports.prototype = {
+
+    configure: function(option) {
+        if (typeof option === 'number') {
+            this._indentationLevel = option;
+        } else if (typeof option === 'string') {
+            assert(
+                option === 'firstParam',
+                this.getOptionName() + ' option requires string value to be "firstParam"'
+            );
+
+            this._alignWithFirstParam = true;
+        } else if (option === true) {
+            this._indentationLevel = 0;
+        } else {
+            assert(
+                false,
+                this.getOptionName() + ' option requires a valid option'
+            );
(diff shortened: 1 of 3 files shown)


JGulbronson:
@hzoo I see you added the 2.4.0 milestone to this. What are the next steps I have to take to make sure it makes it in to that release?

markelog:
> Coverage decreased (-0.01%) to 98.373%

Would you mind fixing that? You can check out the coverage by executing `nm run coverage-html && open coverage.html` command

lib/rules/require-aligned-multiline-params.js
```diff
@@ -0,0 +1,117 @@
...
+ *     three, four, five,
+ *     six, seven, eight) {
+ *   console.log(a);
+ * };
+ * ```
+ *
+ * ##### Valid for `'firstParam'`
+ *
+ * ```js
+ * var test = function(one, two,
```
markelog: Could you also add invalid case?

lib/rules/require-aligned-multiline-params.js
```diff
@@ -0,0 +1,117 @@
...
+
+var assert = require('assert');
+
+module.exports = function() {
+};
+
+module.exports.prototype = {
+
+    configure: function(option) {
+        if (typeof option === 'number') {
```
markelog: We want to switch to https://github.com/mafintosh/is-my-json-valid, to simplify those assertions, we can either leave this as is or you can help us out and simplify live of developers after you.

JGulbronson: I'd lean towards leaving it as is, since I don't see the benefit of switching to `is-my-json-valid` for this particular case. What do you think is the main benefit of switching here?

lib/rules/require-aligned-multiline-params.js
JGulbronson: Added!

lib/rules/require-aligned-multiline-params.js
markelog: In your case, close to none, since you already wrote the logic and tests for it, for supporting it, there some benefit and for developers that could use this as an example for the new rules

JGulbronson: In that case, I'd prefer to leave it as is. The tests/logic is already done, and I don't think this is the best example to use it in. Here, we're only validating the type/value of the `option` parameter. It would feel like overkill to use that library in this scenario, as we're not checking JSON.

markelog: Soon we will start to use it even in most simplest cases, no sweat though, as-is is fine, will try it out on the next rule :-)

PR #1903, opened 2015-10-21
Introduce "fix" as an option

markelog:


markelog:
We can release, right after that

hzoo:
Looks good maybe someone else can look at it

PR #1904, opened 2015-10-21
Configuration: improve resolve of additionalRules

markelog:
Since "configPath" option could point to folder, but not the file


hzoo:
Do we want to increase the timeout for that test and/or look into it again (appveyor)

```
1) modules/config/node-configuration load error filter should accept `errorFilter` from node_modules:
     Error: timeout of 3000ms exceeded. Ensure the done() callback is being called in this test.
```

markelog:
Increased the timeout, but it didn't help, honestly, i would just remove appveyor, don't remember if it did help us in any capacity

PR #1913, opened 2015-10-22
Fix: block statements rules - account for bare blocks

### LINKED ISSUE #1328: Does not play well with blocks
Consider the following code:

```js
/// some code here

{
  let prototype = NamedNodeMap.prototype;
  while (prototype) {
    for (const name of Object.getOwnPropertyNames(prototype)) {
      reservedNames.add(name);
    }
    prototype = Object.getPrototypeOf(prototype);
  }
}

/// more code here
```

I get the following warnings:

```
One space required before opening brace for block expressions at lib/jsdom/living/attributes.js :
   104 |};
   105 |
   106 |exports.NamedNodeMap = NamedNodeMap;
--------------------------------------------^
   107 |
   108 |{

Newline before curly brace for block statement is disallowed at lib/jsdom/living/attributes.js :
   104 |};
   105 |
   106 |exports.NamedNodeMap = NamedNodeMap;
--------------------------------------------^
   107 |
   108 |{
```

Can we make an exception to allow bare blocks while still requiring such style rules for `if`s etc.?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12774539-does-not-play-well-with-blocks?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Ref https://github.com/tmpvar/jsdom/pull/1266

Fixes #1328
Closes gh-1913

### DIFF
--- a/lib/rules/disallow-newline-before-block-statements.js
+++ b/lib/rules/disallow-newline-before-block-statements.js
@@ -185,6 +185,10 @@ module.exports.prototype = {
     check: function(file, errors) {
         var setting = this._setting;
         file.iterateNodesByType('BlockStatement', function(node) {
+            if (isBareBlock(node)) {
+                return;
+            }
+
             if (setting === true || setting.indexOf(getBlockType(node)) !== -1) {
                 var openingBrace = file.getFirstNodeToken(node);
                 var prevToken = file.getPrevToken(openingBrace);
@@ -199,6 +203,15 @@ module.exports.prototype = {
     }
 };
 
+function isBareBlock(node) {
+    var parentNode = node.parentNode;
+
+    return parentNode &&
+    parentNode.type === 'BlockStatement' ||
+    parentNode.type === 'Program' ||
+    parentNode.body && parentNode.body.type === 'BlockStatement' && Array.isArray(parentNode.body);
+}
+
 function getBlockType(node) {
     var parentNode = node.parentNode;
     switch (parentNode.type) {
@@ -206,6 +219,7 @@ function getBlockType(node) {
             return (parentNode.alternate === node) ? 'else' : 'if';
         case 'FunctionDeclaration':
         case 'FunctionExpression':
+        case 'ArrowFunctionExpression':
             return 'function';
         case 'ForStatement':
         case 'ForInStatement':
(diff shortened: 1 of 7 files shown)


lib/rules/disallow-newline-before-block-statements.js
```diff
@@ -199,6 +203,11 @@ module.exports.prototype = {
     }
 };
 
+function isBareBlock(node) {
+    return node.parentNode &&
+    node.parentNode.type === 'BlockStatement' || node.parentNode.type === 'Program';
```
Sebmaster: What if there's a bare block in a function declaration like this:
```js
function foo() {
  {
    let key = 'bar';
  }
  // no more key
}
```

Wouldn't the parentNode be != Program then and it'd be the same issue?

hzoo: Yeah good point it would be `FunctionDeclaration` I guess I can add something like `|| node.parentNode.body && node.parentNode.body.type === 'BlockStatement';`?

Sebmaster: Looks like that would work.

hzoo: Ok that actually broke a lot of tests - we need to also add something like ` && Array.isArray(parentNode.body)` since even `FunctionDeclaration` has a set `body` property that is a `BlockStatement` but it's an object as compared to the `IfStatement` `body` which is an array that can contain multiple statements.

markelog:
> Coverage decreased (-0.02%) to 98.391%

?

hzoo:
It says `https://github.com/jscs-dev/node-jscs/pull/1913/files#diff-04405089acfc36b61502d0c5f5a54f83R236` was not covered anymore for some reason

hzoo:
removed that line - it's not needed right? Also added arrowfunctionexpression

PR #1918, opened 2015-10-23
JUnit Reporter: Changing JUnit XML to match Jenkins JUnit XML Schema

### LINKED ISSUE #1917: JUnit xml report does not get parsed correctly on Jenkins
JUnit xml report does not get parsed correctly on Jenkins.
The JUnit XML that is generated does not match Jenkins JUnit XML Schema, making it (Jenkins) return an error when generating the report.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27650351-junit-xml-report-does-not-get-parsed-correctly-on-jenkins?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

L3O:
Fixes #1917

### DIFF
--- a/lib/reporters/jenkinsjunit.js
+++ b/lib/reporters/jenkinsjunit.js
@@ -10,14 +10,13 @@ module.exports = function(errorCollection) {
     errorCollection.forEach(function(errors) {
         var errorsCount = errors.getErrorCount();
         var testcase = testsuite.ele('testcase', {
-            name: errors.getFilename(),
-            failures: errorsCount
+            name: errors.getFilename()
         });
 
         i += errorsCount;
 
         errors.getErrorList().forEach(function(error) {
-            testcase.ele('failure', {}, errors.explainError(error));
+            testcase.ele('failure', { 'message': errors.explainError(error)});
         });
     });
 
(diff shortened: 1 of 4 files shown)


markelog:
Do you have a link to the schema?

L3O:
This is the schema I looked at:
https://svn.jenkins-ci.org/trunk/hudson/dtkit/dtkit-format/dtkit-junit-model/src/main/resources/com/thalesgroup/dtkit/junit/model/xsd/junit-4.xsd

markelog:
Current schema is compatible with ant, if we would accept your proposition it would mean we wouldn't support it but would support jenkins one. 

I'm not sure what is more popular and what should be used by default. But since this is first request for format change since this reporter was introduced (for two years that is) and because format change would warrant major version i'm leaning to recommend package this reporter as jscs plugin.

L3O:
So what you are saying is create another jscs plugin for this schema ?
Or create new reporter on the project for this schema ?

markelog:
> So what you are saying is create another jscs plugin for this schema ?

Yeah

L3O:
What about a new reporter for jenkins ? 
Like the one in the commit `jenkinsjunit.js` that export an xml on the schema expected from jenkins.

markelog:
I would recommend to do it as a plugin first, since we already have an overwhelming amount of default reporters.

If it will became popular we can consider adding it to the dist.

PR #1920, opened 2015-10-24
requireSpaceAfterComma

### LINKED ISSUE #1915: error loop regarding javasript object in array
Error rendered by [linter-jscs](https://github.com/AtomLinter/linter-jscs) Atom text editor library, but believe the problem is node-jscs.

Below is a screenshot of the issue I was having, where it indicates to me that there is a missing comma before closing beracket.

image of issue

So I added the the missing comma, so that line 10 has the error fixed:
image of step 2

I then get a second error, 'Space required after comma', as indicated.

Once I fixed that, I get the last error:
last error

As you can see, this is a loop of errors since I remove the space, I get back to the second error.

Let me know if it's a user problem.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27631007-error-loop-regarding-javasript-object-in-array?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bjdixon:
New rule value: exceptTrailingCommas will allow commas to be followed by
either } or ] as an alternative to a space.

Fixes #1915

### DIFF
--- a/lib/rules/require-space-after-comma.js
+++ b/lib/rules/require-space-after-comma.js
@@ -1,27 +1,45 @@
 /**
  * Requires space after comma
  *
- * Types: `Boolean`
+ * Types: `Boolean`, or `String`
  *
- * Values: `true` to require a space after any comma
+ * Values:
+ *  - `Boolean`: `true` to require a space after any comma
+ *  - `String`: `"exceptTrailingCommas"` to require a space after all but trailing commas
  *
  * #### Example
  *
  * ```js
  * "requireSpaceAfterComma": true
  * ```
+ * ```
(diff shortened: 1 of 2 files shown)


markelog:
How about we make it as 
```js
allExcept: ['trailing']
```

?

bjdixon:
@markelog makes sense. That will be more consistent with ``` disallowCommaBeforeLineBreak ```

PR #1922, opened 2015-10-25
New Rule:requireSpacesInGenerator /n Fixes #1879

### LINKED ISSUE #1879: New Rule: (require|disallow)GeneratorSpacing
Ref https://github.com/jscs-dev/node-jscs/issues/1175

```js
function * ()
function* ()
function *()
function*()
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27486409-new-rule-require-disallow-generatorspacing?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

stefania11:
Fixes #1879, worked @hzoo and @chrismlee. We left out invalid examples for documentation because they self-apparent for people that use generators.

### DIFF
--- a/lib/rules/require-spaces-around-generators.js
+++ b/lib/rules/require-spaces-around-generators.js
@@ -1,5 +1,5 @@
 /**
- * Requires space before and after `*` in generator functions  ([named](#requirespacesinnamedfunctionexpression)
+ * Requires space before and after `*` in generator functions
  *
  * Types: `Object`
  *
@@ -38,7 +38,7 @@ module.exports.prototype = {
     configure: function(options) {
         assert(
             typeof options === 'object',
-            this.getOptionName() + ' option must be the object'
+            this.getOptionName() + ' option must be an object'
         );
 
         if ('beforeStar' in options) {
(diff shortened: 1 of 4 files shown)


lib/rules/require-spaces-around-generators.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * Requires space before and after `*` in generator functions  ([named](#requirespacesinnamedfunctionexpression)
```
hzoo: Don't think you need to do the `([named](#requirespacesinnamedfunctionexpression)` part

lib/rules/require-spaces-around-generators.js
```diff
@@ -0,0 +1,128 @@
...
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            typeof options === 'object',
+            this.getOptionName() + ' option must be the object'
```
hzoo: `the object` to `an object`

hzoo:
Awesome! Thanks for this awesome contribution! It was really fun being a part of the open source meetup and hope yall learned something! Happy to see more contributions!

lib/rules/require-spaces-around-generators.js
```diff
@@ -0,0 +1,121 @@
...
+
+            // shorthand or constructor methods
+            if (parent.method || parent.type === 'MethodDefinition') {
+                functionNode = parent.key;
+            }
+
+            var currentToken = file.getFirstNodeToken(functionNode);
+
+            if (!node.generator) {
+                return;
```
mrjoelkemp: You should be able to do this early return at the top of the function no?

mrjoelkemp:
Why does this skip async?

hzoo:
Because in `async function * () {}`, async is the first node token and we
want to skip that and check star spacing

lib/rules/require-spaces-around-generators.js
hzoo: Yeah true

PR #1926, opened 2015-10-26
fix wrong text in test case

Gerhut:


hzoo:
I think you have more changes than you intended in this PR? or should this be closed in favor of the other one you already made?

PR #1927, opened 2015-10-26
add allExcept: ["moduleExports"] in requireMatchingFunctionName rule

Gerhut:
When allExcept: ['moduleExports'] set, ignore no matching names of function declearation assignment to module.exports.

Valid:

```js
module.exports = function foo() {}

module['exports'] = function bar() {}

k.module.exports = function exports() {}
```

Invalid:

```js
k.module.exports = function baz() {}
```

### DIFF
--- a/lib/rules/require-matching-function-name.js
+++ b/lib/rules/require-matching-function-name.js
@@ -2,11 +2,12 @@
  * Requires function names to match member and property names.
  *
  * It doesn't affect anonymous functions nor functions assigned to members or
- * properties named with a reserved word.
+ * properties named with a reserved word. Assigning to `module.exports` is also
+ * ignored, unless `includeModuleExports: true` is configured.
  *
- * Type: `Boolean`
+ * Types: `Boolean` or `Object`
  *
- * Value: `true`
+ * Values: `true` or Object with `includeModuleExports: true`
  *
  * #### Example
  *
@@ -30,6 +31,14 @@
  * var test = {foo: function foo() {}};
  * ```
  *
+ * ```js
+ * module.exports = function foo() {};
+ * ```
+ *
+ * ```js
+ * module['exports'] = function foo() {};
+ * ```
+ *
  * ##### Invalid
  *
  * ```js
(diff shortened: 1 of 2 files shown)


lib/rules/require-matching-function-name.js
```diff
@@ -93,56 +124,59 @@ module.exports.prototype = {
...
 
-function checkForMember(assignment, skip, errors) {
-    var _name = _resolvePatternName(assignment.left);
-    if (_name === false || skip(_name, assignment.right)) {
-        return;
-    }
+        function checkForMember(assignment, skip, errors) {
+            if (_this._exceptModuleExports && _isModuleExports(assignment.left)) {
+                return;
+            }
```
hzoo: I don't think it was necessary to move all the functions into the object. You could just call this in the https://github.com/jscs-dev/node-jscs/pull/1927/files#diff-105ac0edfe6913192922933858c13d7fR82 so you wouldn't have to create `_this` right?

Gerhut: Fixed and sorry for my poor coding skill. However, [The line you referenced](https://github.com/jscs-dev/node-jscs/pull/1927/files#diff-105ac0edfe6913192922933858c13d7fR82) is also in a function scope https://github.com/jscs-dev/node-jscs/pull/1927/files#diff-105ac0edfe6913192922933858c13d7fR77 so I just cache the `_exceptModuleExports`.

markelog:
Sounds like this should be enabled by default? i.e. no need to introduce another option?

> Coverage decreased (-0.04%) to 97.809%

? You can check the coverage by running `npm run coverage-html && open coverage.html`

hzoo:
Or the link https://coveralls.io/builds/3958392/source?filename=lib%2Frules%2Frequire-matching-function-name.js

Gerhut:
Coverage test passed, Thank you for your help.

markelog:
What about making this behaviour enabled by default?

Gerhut:
done

lib/rules/require-matching-function-name.js
```diff
@@ -54,10 +54,16 @@ module.exports = function() {};
 
 module.exports.prototype = {
     configure: function(requireMatchingFunctionName) {
-        assert(
-            requireMatchingFunctionName === true,
-            'requireMatchingFunctionName option requires true value or should be removed'
-        );
+        if (typeof requireMatchingFunctionName === 'object') {
+            assert(requireMatchingFunctionName.includeModuleExports === true,
```
markelog: Could you document such value?

Gerhut:
Documented

markelog:
Thank you!

PR #1935, opened 2015-10-27
Whitespace token instead of `whitespaceBefore` token property

markelog:
WIP


markelog:
Okay, since @mdevils should be almost finish with it, i'm closing it

PR #1938, opened 2015-10-28
Update: `requireCurlyBraces`: create exceptions for if and else statements

### LINKED ISSUE #244: requireCurlyBraces one-statement / one-expression exception
"requireCurlyBraces" is a great option but needs an exception. The only time it's acceptable to ignore it is for early returns or continuations.

[Example from our Fabric.js](https://github.com/kangax/fabric.js/blob/master/src/canvas.class.js#L341):

```js
 if (!target) return;
```

[Or with `continue`](https://github.com/kangax/fabric.js/blob/master/src/canvas.class.js#L1086):

```js
if (!this._objects[i] || !this._objects[i].active) continue;
```

[Example from Underscore](https://github.com/jashkenas/underscore/blob/master/underscore.js#L82):

```js
if (iterator.call(context, obj[i], i, obj) === breaker) return;
```

[Example from Angular](https://github.com/angular/angular.js/blob/master/src/Angular.js#L836):

```js
if (key.charAt(0) === '$' || isFunction(o1[key])) continue;
```

etc.

Most of the time, I want `if` statements to have curly braces, except when there's 1 statement in its body, such as `return` or `continue`.

I couldn't find a way for JSCS to ignore these, hence the request.

I could think of 2 levels of allowance:

- Only allow "empty" statements, such as  `return;`, `continue;`, `break;`
- Only allow one statement, such as `return false;`, `return foo();`, etc.

I'm leaning towards 2nd. I noticed that both [Underscore](https://github.com/jashkenas/underscore/blob/master/underscore.js#L97) and [Angular](https://github.com/angular/angular.js/blob/master/src/Angular.js#L837) use the latter as well.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2009589-requirecurlybraces-one-statement-one-expression-exception?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hzoo:
Make an exception for return, break, continue statements

Fixes #244
Closes gh-1938

```js
    "requireCurlyBraces": {
        "allExcept": ["return", "continue", "break"],
        "keywords": ["if", "else", "for", "while", "do", "switch", "return", "try", "catch"]
    },
```

cc @kangax - Although we don't have an autofix for this yet - hopefully in 3.0 with https://github.com/cst/cst

### DIFF
--- a/lib/rules/require-curly-braces.js
+++ b/lib/rules/require-curly-braces.js
@@ -1,9 +1,17 @@
 /**
  * Requires curly braces after statements.
  *
- * Types: `Array` or `Boolean`
+ * Types: `Array` or `Boolean` or `Object`
  *
- * Values: Array of quoted keywords or `true` to require curly braces after the following keywords:
+ * Values:
+ *     - Array of quoted keywords
+ *     - `true` to require curly braces after the following keywords
+ *     - `Object`
+ *         - `'keywords'`
+ *             - Array of quoted keywords
+ *         - `'allExcept'`
+ *             - Array of keywords inside of the block that would allow curly braces
+ *             - Ex: ["return" , "continue", "break"]
  *
  * JSHint: [`curly`](http://jshint.com/docs/options/#curly)
  *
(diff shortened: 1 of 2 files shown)


pomeh:
Thanks @hzoo for this addition. I have a question: why the exceptions list is only used for `if` and `else` statements ? I think it should be possible to have code like this too:

```js
function foo (bar) {
    switch (bar)
    {
        case 'foo': return 'bar';
        case 'baz':
                var baz = 'qux';
                // more statements
                return baz;
        default: return 0;
    }
}
```

In this case, the exceptions for `case` statement would make sense. What do you think ?

Issue #1805, opened 2015-09-22
ES6 Object literal function short-hand and requireSpacesInAnonymousFunctionExpression { beforeOpeningRoundBrace: true } 


indexzero:
This code:

``` js
var anObject = {
  ok() {
    console.log('ok');
  }
};
```

With this setting:

```
{
  "requireSpacesInAnonymousFunctionExpression": {
    "beforeOpeningRoundBrace": true
  }
}
```

Causes this output:

```
Missing space before opening round brace at wtf.js :
     1 |
     2 |var anObject = {
     3 |  ok() {
------------^
     4 |    console.log('ok');
     5 |  }

1 code style error found.
```

From an AST point of view I can see why this would be the case, but from a style point of view you can see how this is very close to a named function expression. Would an exception for this case be excepted?

indexzero:
It is also worth mentioning that since there is no `function` keyword here. This can cause an unlintable code situation when used in conjunction with other rules.

hzoo:
Yeah this should be fixed - although I just found https://github.com/jscs-dev/node-jscs/issues/1513 - we actually have a option for shorthand methods now - http://jscs.info/rule/requireSpacesInAnonymousFunctionExpression

We should an example to the docs though

hzoo:
Although maybe this should be by default

PR #1941, opened 2015-10-29
Yandex preset: add "allExcept": ["shorthand"] to requireSpacesInAnony…

ikokostya:
…mousFunctionExpression rule

See https://github.com/jscs-dev/node-jscs/issues/1805

I can't understand how I can add test for this rule to https://github.com/jscs-dev/node-jscs/blob/master/test/data/options/preset/yandex.js, because it works only in `ES6`.

### DIFF
--- a/presets/yandex.json
+++ b/presets/yandex.json
@@ -47,7 +47,8 @@
     },
     "requireSpacesInAnonymousFunctionExpression": {
         "beforeOpeningRoundBrace": true,
-        "beforeOpeningCurlyBrace": true
+        "beforeOpeningCurlyBrace": true,
+        "allExcept": ["shorthand"]
     },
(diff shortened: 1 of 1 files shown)


tarmolov:
:ok:

markelog:
> I can't understand how I can add test for this rule to https://github.com/jscs-dev/node-jscs/blob/master/test/data/options/preset/yandex.js, because it works only in ES6.

Like that - https://github.com/wikimedia/jscs-preset-wikimedia/blob/master/test/specs/preset-wikimedia.js#L28-L33, it would be better though to move this preset to its own repo.

ikokostya:
See https://github.com/jscs-dev/node-jscs/pull/1957

Issue #1812, opened 2015-09-28
Discussing Tags for commits


hzoo:
Although this isn't as important with a more manual changelog

Suggestions (and examples):

- `Fix` | `Bugfix`
- `Docs`
  - ```Docs: `disallowParenthesesAroundArrowParam` fixes```
- `New` | `New Rule`
  - ```New: `disallowArrowFunctions` ```
- `Update` | `Rule Update` - new rule values, support that isn't a bug fix
  - ``` Update: `disallowMultiLineTernary` - add `nested` option ```
- `Internal`: - something for jsfile, string-checker
  - ``` Internal: `CLI` - implement `extract` option ```
- `Config` | `Configuration` - change related to .jscsrc
  - ```Config: strip BOM from config files```
- `Dependency` | `Package` | `Upgrade` - updating a dependency
  - ```Upgrade: esprima@2.6.0```
- `Chore` | `Build` - related to dev-ops, CI, or dependencies
- `Misc`
- `Preset`
- `Test`
- `Feat`
- `Perf`
- `Cleanup` | `Refactor`
- `Security`
- ` Breaking`

Don't need that many but just some thoughts

zxqfox:
I'm just looking at changelog. Not sure that we need so detailed list of tags. `Breaking` can be a flag, like: `Fix: BREAKING etc.`

Docs: Documentation only changes
Feat: A new feature
Fix: A bug fix
Chore: Changes to the build process or auxiliary tools and libraries such as documentation generation
Preset
New Rule
Update Rule
Perf
Cleanup | Refactor

Misc ← Test, Config, Internal, Security
Chore ← Dependency | Package | Upgrade, ?

hzoo:
Less categories is good - just shouting out possibilities.

General: Docs, Fix, Chore, Misc
JSCS: New Rule, Update Rule, Preset

markelog:
We would need to use those tags with changelog generator, since primarily we want to do it for changelog itself.

Like what it would generate if we would use `Fix:Breaking: ...`?

Also we need to make sure we can add such checks in githook.

hzoo:
Similar to https://github.com/jscs-dev/node-jscs/issues/407 so goina close that one again.

https://github.com/jzaefferer/commitplease, http://gitcop.com/, etc

hzoo:
[...]

PR #1942, opened 2015-10-29
Whitespace token

mdevils:
Implementation of Whitespace-token in JSCS.

This is a prerequirement for CST integration.

/cc @hzoo @markelog

### DIFF
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -37,9 +37,11 @@ var JsFile = function(params) {
     this._lineBreaks = null;
     this._lines = this._source.split(/\r\n|\r|\n/);
 
+    var hasErrors = false;
     try {
         this._tree = parseJavaScriptSource(this._source, params.esprima, params.esprimaOptions);
     } catch (e) {
+        hasErrors = true;
         this._parseErrors.push(e);
     }
 
@@ -47,8 +49,8 @@ var JsFile = function(params) {
     this._scope = null;
 
     this._tokens = this._buildTokenList(this._tree.tokens, this._tree.comments);
-    this._addEOFToken();
-    this._applyWhitespaceData(this._tokens, this._source);
+    this._addEOFToken(hasErrors);
+    this._tokens = this._addWhitespaceTokens(this._tokens, this._source);
 
     this._setTokenIndexes();
 
@@ -134,6 +136,50 @@ JsFile.prototype = {
     },
 
     /**
+     * Sets whitespace before specified token.
+     *
+     * @param {Object} token
+     * @param {String} whitespace
+     */
+    setWhitespaceBefore: function(token, whitespace) {
+        var whitespaceToken = this.getPrevToken(token, {includeWhitespace: true});
+        if (whitespaceToken && whitespaceToken.type === 'Whitespace') {
+            if (whitespace === '') {
+                this.removeToken(whitespaceToken);
+            } else {
+                whitespaceToken.value = whitespace;
+            }
+        } else {
+            if (whitespace === '') {
+                return;
+            }
+            var tokenIndex = this._tokens.indexOf(token);
+            this._tokens.splice(tokenIndex, 0, {
+                type: 'Whitespace',
+                value: whitespace,
+                isWhitespace: true
+            });
+            for (var i = tokenIndex, l = this._tokens.length; i < l; i++) {
+                this._tokens[i]._tokenIndex = i;
+            }
+        }
+    },
+
+    /**
+     * Returns whitespace before specified token.
+     *
+     * @param {Object} token
+     */
+    getWhitespaceBefore: function(token) {
+        var whitespaceToken = this.getPrevToken(token, {includeWhitespace: true});
+        if (whitespaceToken && whitespaceToken.type === 'Whitespace') {
+            return whitespaceToken.value;
+        } else {
+            return '';
+        }
+    },
+
+    /**
      * Remove some entity (only one) from array with predicate
      *
      * @param {Array} entities
@@ -150,7 +196,7 @@ JsFile.prototype = {
     },
 
     /**
-     * Remove token from token list
+     * Remove token from token list.
      *
      * @param {Object} token
      */
@@ -243,9 +289,15 @@ JsFile.prototype = {
         for (var i = 0, l = tokens.length; i < l; i++) {
             var token = tokens[i];
 
+            token._tokenIndex = i;
+
+            if (token.type === 'Whitespace') {
+                continue;
+            }
+
             // tokens by range
-            tokenRangeStartIndex[token.range[0]] = i;
-            tokenRangeEndIndex[token.range[1]] = i;
+            tokenRangeStartIndex[token.range[0]] = token;
+            tokenRangeEndIndex[token.range[1]] = token;
 
             // tokens by line
             var lineNumber = token.loc.start.line;
@@ -254,8 +306,6 @@ JsFile.prototype = {
             }
 
             tokensByLineIndex[lineNumber].push(token);
-
-            token._tokenIndex = i;
         }
 
         return {
@@ -271,8 +321,7 @@ JsFile.prototype = {
      * @returns {Object|null}
      */
     getTokenByRangeStart: function(start) {
-        var tokenIndex = this._tokenRangeStartIndex[start];
-        return tokenIndex === undefined ? null : this._tokens[tokenIndex];
+        return this._tokenRangeStartIndex[start] || null;
     },
 
     /**
@@ -281,8 +330,7 @@ JsFile.prototype = {
      * @returns {Object|null}
      */
     getTokenByRangeEnd: function(end) {
-        var tokenIndex = this._tokenRangeEndIndex[end];
-        return tokenIndex === undefined ? null : this._tokens[tokenIndex];
+        return this._tokenRangeEndIndex[end] || null;
     },
 
     /**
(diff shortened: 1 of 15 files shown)


lib/rules/disallow-multiple-line-breaks.js
```diff
@@ -46,9 +46,13 @@ module.exports.prototype = {
...
         // Iterate over all tokens (including comments)
-        file.getTokens().forEach(function(token, index, tokens) {
+        file.getTokens().forEach(function(token) {
+            if (token.type === 'Whitespace') {
+                return;
+            }
+
             // If there are no trailing tokens, exit early
-            var nextToken = tokens[index + 1];
+            var nextToken = file.getNextToken(token, {includeComments: true});
```
markelog: Is this a relevant change?

lib/rules/maximum-number-of-lines.js
```diff
@@ -49,7 +49,7 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
-        var firstToken = file.getFirstToken();
+        var firstToken = file.getFirstToken({includeComments: true});
```
markelog: ?

lib/rules/disallow-multiple-line-breaks.js
mdevils: Yes, sure. Now token list contains both `Code`, `Comment` and `Whitespace` tokens.

markelog: Oh, okay

lib/rules/require-spaces-inside-parenthesized-expression.js
```diff
@@ -97,6 +97,7 @@ module.exports.prototype = {
             });
         });
 
+        var d = Math.random();
```
markelog: ?

lib/rules/disallow-multiple-line-breaks.js
hzoo: I feel like we almost always want to add includeComments or something?

lib/rules/maximum-number-of-lines.js
mdevils: What do you question?

test/specs/token-categorizer.js
```diff
@@ -34,7 +34,7 @@ describe('token-categorizer', function() {
         });
 
         it('should limit input to open parentheses', function() {
-            var nonParen = sharedFile.getPrevToken(sharedFile.getLastToken());
+            var nonParen = sharedFile.getPrevToken(sharedFile.getLastToken({includeComments: true}));
```
markelog: So why change this file?

lib/rules/disallow-multiple-line-breaks.js
mdevils: Actually not always according to the code :)

lib/rules/maximum-number-of-lines.js
markelog: This change is not relevant to introducing of whitespace token isn't it?

test/specs/token-categorizer.js
mdevils: I needed `getFirstToken` to ignore `Whitespace`, but just silently having this as a default behaviour seemed not so good to me. So I changed `getFirstToken` and `getLastToken` signatures to match `getPrev/NextToken` signatures.

markelog: Okay, gotcha

lib/js-file.js
```diff
@@ -134,6 +136,50 @@ JsFile.prototype = {
     },
 
     /**
+     * Sets whitespace before specified token.
+     *
+     * @param {Object} token
+     * @param {String} whitespace
+     */
+    setWhitespaceBefore: function(token, whitespace) {
```
markelog: I'd say this method is a bit complicated, would recommend to consider a decomposition

lib/rules/disallow-multiple-line-breaks.js
hzoo: More like we forget

lib/js-file.js
```diff
@@ -387,6 +441,32 @@ JsFile.prototype = {
...
+     * Returns the first token before the given.
+     *
+     * @param {Object} token
+     * @param {Object} [options]
+     * @param {Boolean} [options.includeComments=false]
+     * @param {Boolean} [options.includeWhitespace=false]
+     * @returns {Object|null}
+     */
+    getPrevToken: function(token, options) {
+        return this._getPrevTokenFromIndex(token._tokenIndex - 1, options);
```
markelog: So we only need to those private helpers for increment/decrement token index?

lib/js-file.js
```diff
@@ -588,6 +668,23 @@ JsFile.prototype = {
     },
 
     /**
+     * Executes callback for each token in token list.
+     *
+     * @param {Function} cb
+     * @private
+     */
+    _forEachToken: function(cb) {
```
markelog: For speed?

lib/js-file.js
```diff
@@ -134,6 +136,50 @@ JsFile.prototype = {
...
+            if (whitespace === '') {
+                return;
+            }
+            var tokenIndex = this._tokens.indexOf(token);
+            this._tokens.splice(tokenIndex, 0, {
+                type: 'Whitespace',
+                value: whitespace,
+                isWhitespace: true
+            });
+            for (var i = tokenIndex, l = this._tokens.length; i < l; i++) {
```
markelog: They say v8 already do the optimization for caching `length`

markelog: Just fyi btw, i'm still using it :-)

lib/js-file.js
mdevils: Nope, for iterating the list which changes during the iteration. I had problems with it. Because if you just use `forEach`, then `index` becomes invalid once you push items (like adding some whitespace before the token).

lib/js-file.js
mdevils: How would you propose to decompose?

lib/js-file.js
mdevils: Actually we can have just 1 private helper and pass the direction: `1` or `-1`.

lib/js-file.js
markelog: There is two primary code paths - if we have whitespace and when we need to add a whitespace token, i would propose to try to decompose it by those points

lib/js-file.js
mdevils: Lol, I wrote this just by the habit, will fix this :)

markelog:
Is there any speed penalty?

markelog:
> Coverage decreased (-0.3%) to 97.554%

I know we gonna change this soon, but would still recommend to go down with this metric

lib/rules/disallow-multiple-line-breaks.js
markelog: Maybe this is something we should consider for 3.0

markelog: It is really seems that we need to make it enabled by default, but since now we can ignore whitespaces too, it might look inconsitent

hzoo:
Code coverage decrease is for the new methods `setWhitespaceBefore, `getWhitespaceBefore`, `_getNextTokenFromIndex`, `_getPrevTokenFromIndex`. Looks good but I guess it can be harder to read through because of changes that don't seem immediately relevant do the whitespace chane

lib/js-file.js
mdevils: Tried to decompose, but it looked ugly. I've simplified a bit and added comments instead.

mdevils:
Got increased coverage :)

lib/rules/disallow-multiple-line-breaks.js
mdevils: This will be fixed with CST which has different properties for these cases.

mdevils:
I finished fixing according to the comments.

lib/js-file.js
```diff
@@ -134,6 +136,53 @@ JsFile.prototype = {
...
+            for (var i = tokenIndex; i < this._tokens.length; i++) {
+                this._tokens[i]._tokenIndex = i;
+            }
+        }
+    },
+
+    /**
+     * Returns whitespace before specified token.
+     *
+     * @param {Object} token
```
hzoo: this returns a string right?

lib/js-file.js
```diff
@@ -308,50 +359,67 @@ JsFile.prototype = {
...
-    getPrevToken: function(token, options) {
-        var index = token._tokenIndex - 1;
-        if (index < 0) {
-            return null;
-        }
-
-        if (options && options.includeComments) {
-            return this._tokens[index] || null;
-        }
+    _getTokenFromIndex: function(index, direction, options) {
```
hzoo: Not a big deal but I don't really like the `direction `1` - forward or `-1` - backwards`
since you can't tell what it means from looking at `this._getTokenFromIndex(this._tokens.length - 1, -1, options);` Oh but I see it's an internal method anyway

lib/js-file.js
mdevils: Yes, I will add jsdoc.

lib/js-file.js
mdevils: I thinks it's fine for internal method including the fact that it will be removed soon.

markelog: Well, we could use keywords, like "left", "right", but yeah, if we remove it soon, then no big deal, but i always afraid phrase - "There is nothing more permanent than temporary"

lib/js-file.js
markelog: I guess it is fine for temporary measure, but we might need to start about limit for cyclomatic complexity

lib/js-file.js
mdevils: In this case CST will completely replace it.

markelog:
Don't forget to apply git tags next time. 

@hzoo you wanted to add commitplease right?

hzoo:
I guess I don't care that much anymore @markelog haha. We rewrite the changelog anyway but if we want to do it still ok.

mdevils:
@hzoo thank you!
@markelog which git tags? I have probably missed something.

markelog:
@hzoo 

> We rewrite the changelog anyway but if we want to do it still ok.

Well, we still experimenting at this point, with tags we could create more clean changelog.

@mdevils https://github.com/jscs-dev/node-jscs/issues/1812, bureaucracy at its best in other words :-)

Issue #907, opened 2015-01-13
Operator is not printed properly in error message


ptomato:
**Platform:** Mac OSX 10.10
**Version:** 1.8.1

With this config file
```json
{
    "requireSpaceAfterBinaryOperators": ["="]
}
```
and this source file
```js
var thing =1;
```
I get the following output:
```
Operator undefined should not stick to following expression at example.js :
     1 |var thing =1;
-------------------^
     2 |


1 code style error found.
```
I expect that should be something like `Operator "=" should not stick...`, i.e. the operator should be printed in the error message rather than `undefined`.

ptomato:
Could reproduce in 1.9.0 as well.

PR #1948, opened 2015-11-02
requireSpaceBeforeBinaryOperators: report "operator =" instead of "undefined"

Rob--W:
Similar to #907, but for the spacing before the = operator.
Before: "Operator undefined should not stick to preceding expression"
After : "Operator = should not stick to preceding expression"

### DIFF
--- a/test/specs/rules/require-space-before-binary-operators.js
+++ b/test/specs/rules/require-space-before-binary-operators.js
@@ -89,7 +89,8 @@ describe('rules/require-space-before-binary-operators', function() {
 
     it('should report for assignment expression', function() {
         checker.configure({ requireSpaceBeforeBinaryOperators: ['='] });
-        expect(checker.checkString('var x=1')).to.have.one.validation.error.from('requireSpaceBeforeBinaryOperators');
+        expect(checker.checkString('var x=1'))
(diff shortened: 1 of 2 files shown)


hzoo:
Thanks!

test/specs/rules/require-space-before-binary-operators.js
```diff
@@ -89,7 +89,12 @@ describe('rules/require-space-before-binary-operators', function() {
 
     it('should report for assignment expression', function() {
         checker.configure({ requireSpaceBeforeBinaryOperators: ['='] });
-        expect(checker.checkString('var x=1')).to.have.one.validation.error.from('requireSpaceBeforeBinaryOperators');
+        var errors = checker.checkString('var x=1');
+        expect(errors).to.have.one.validation.error.from('requireSpaceBeforeBinaryOperators');
+        var error = errors.getErrorList()[0];
+        expect(error.line).to.equal(1);
```
markelog: You can replace this with
```js
expect(errors).to.have.one.error('Operator = should not stick to preceding expression');
```

Rob--W: Done!

Issue #1916, opened 2015-10-23
Oddness with requireSpacesInsideParentheses, quotes, and multiple args


texel:
Per https://github.com/rwaldron/idiomatic.js/#spacing, I'm trying to support this:

```javascript
// Function call with multiple arguments
foo( "bar", baz );

// Single argument string literal, no space
foo("bar");
```

With the current rules, I can use `"except": ['"']` but that will result in the following fixes:

```javascript
foo("bar", baz) // becomes
foo("bar", baz )

foo(baz, "bar") // becomes
foo( baz, "bar")
```

It seems like we might need explicit exceptions for function arguments. I'm fine with digging in and attempting a fix, but I wanted to get the maintainers' opinions first on how best to tackle this.

markelog:
How about we make it as separate rule? Two rules even, since one would require to have spaces -

```js
foo( baz, "bar" )
```
and the other would disallow it - 
```js
foo("bar")
```

Does it sound doable?

texel:
Sure, that sounds reasonable.

markelog:
Would you like to take a stab at it?

texel:
Sorry for the delay... yeah I'd like to.

hzoo:
I'm putting an [`orphaned`](https://github.com/jscs-dev/node-jscs/issues?q=label%3Aorphaned+is%3Aclosed) label to note this issue isn't planned to be worked on, but might be a good contribution as a custom rule (not in core). We can all check back on this list for future rule ideas/options.

We released [3.0](https://github.com/jscs-dev/node-jscs/releases/tag/v3.0.0) recently! We also wrote a [post](https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.oot3wvgpu) about how 3.0 will be our last major release (since we are deprecating the project) and that we are [merging with ESLint](http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint)!

Thus we won't be working on any new features moving forward which includes new rules/options. We will continue to fix bugs and work on CST/better autofixing for current rules in the upcoming months both to help JSCS users and to gain experience that we can hopefully use in ESLint. But otherwise we both teams will be working on getting [ESLint feature parity with JSCS](https://github.com/eslint/eslint/milestones/JSCS%20Compatibility) and just a nice transition for our users.

[...]

PR #1951, opened 2015-11-02
Add rules for spaces inside function calls [WIP]

texel:
Speculative work to address issue #1916


lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowSpacesInsideFunctionCalls": true
+ *
+ * // or
+ *
+ * "disallowSpacesInsideFunctionCalls": {
+ *     "allExcept": [ "'", "\'" ]
```
zxqfox: `"'" === "\'"` right?

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+ * - true
+ * - `Object`:
+ *      - "only": An array containing tokens to disallow.
+ *      - "onlySingleArgument": true
+ *      - "onlyMultipleArguments": true
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowSpacesInsideFunctionCalls": true
```
zxqfox: What is the default?

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
+/**
+ * Disallows space after opening and before closing function call parentheses.
+ *
+ * Types: `Boolean` or `Object`
+ *
+ * Values:
+ * - true
+ * - `Object`:
+ *      - "only": An array containing tokens to disallow.
```
zxqfox: I'd say we don't need too much options. Can we left just 2 params: `allExcept` with tokens array and, let's say, `arity` with 2 variants: single, multiple or even a Number.

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(value) {
+        var isObject = typeof value === 'object';
+
+        var error = this.getOptionName() + ' rule requires string value true or object';
```
zxqfox: "string value true"? nooo.

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+        var validFlags = [
+            'only',
+            'onlySingleArgument',
+            'onlyMultipleArguments'
+        ];
+
+        if (isObject) {
+            var hasValidFlags;
+
+            validFlags.forEach(function(flag) {
```
zxqfox: Better to check keys to exist in validFlags. Like `assert(Object.keys(value).every(function (key) { return validFlags.indexOf(key) !== -1; }))`

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+                hasValidFlags = hasValidFlags || flag in value;
+            });
+
+            assert(hasValidFlags, error);
+        } else {
+            assert(value === true, error);
```
zxqfox: Why not
```
if (!isObject) {
  assert(value === true, error);
  return;
}

// do the rest
```
? ;-) We usually like early returns.

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+ * ```
+ */
+
+var assert = require('assert');
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(value) {
+        var isObject = typeof value === 'object';
```
zxqfox: Better to check for isArray also.

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+                    token: argLastToken,
+                    nextToken: closeParens,
+                    message: 'Space before closing parenthesis'
+                });
+            } else {
+                var callToken = file.getTokenByRangeStart(functionCall.range[0]);
+                var openParen = file.getNextToken(callToken);
+                var closeParen = file.getNextToken(openParen);
+
+                errors.assert.noWhitespaceBetween({
```
zxqfox: same here — better to move to the top of the function, right after single/multiple checks (or even before them).

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+
+    check: function(file, errors) {
+        var only = this._only;
+        var onlySingleArgument = this._onlySingleArgument;
+        var onlyMultipleArguments = this._onlyMultipleArguments;
+
+        file.iterateNodesByType('CallExpression', function(functionCall) {
+            var functionArgs = functionCall.arguments;
+
+            if (onlySingleArgument && functionArgs.length !== 1) {
```
zxqfox: What about noArgs case? Is it fine here to have a space between parens?

lib/rules/disallow-spaces-inside-function-calls.js
```diff
@@ -0,0 +1,143 @@
...
+            } else if (onlyMultipleArguments && functionArgs.length === 1) {
+                return;
+            }
+
+            if (functionArgs.length) {
+                var firstArg = functionArgs[0];
+                var lastArg = functionArgs[functionArgs.length - 1];
+                var argFirstToken = file.getTokenByRangeStart(firstArg.range[0]);
+                var argLastToken = file.getTokenByRangeEnd(lastArg.range[1]);
+                var openParens = file.getPrevToken(argFirstToken);
```
zxqfox: openParens or openParen?

gibson042:
I'm really sorry to jump in so late on this, but I have some major reservations.

# The Name
Consistency with #1254 would use **SpacesInsideParenthesizedArguments** if optimizing for alphabetical grouping, or [**SpacesInsideCall**(**Expression**)?**Arguments**](https://github.com/jscs-dev/node-jscs/pull/1254#issuecomment-133056398) if optimizing for unambiguous comprehension. Which way to go remains an open question, since it was asked in https://github.com/jscs-dev/node-jscs/pull/1254#issuecomment-133107488 but never resolved. But **SpacesInsideFunctionCalls** is in my opinion _far_ too similar to the existing **[SpacesInCallExpression](http://jscs.info/rule/disallowSpacesInCallExpression)** rules (`foo‸(a, b, …)`).

# The Implementation
1acc0c0b4c6afc1d2fab9257712977b76e887450 introduced `TokenCategorizer.categorizeOpenParen` and `TokenCategorizer.categorizeCloseParen` for just such purposes, and all parentheses rules should use them, lest we miss edge cases in unpredictable and inconsistent ways. For example,  `invocation( (parenthesizedExpression) )` has an open parenthesis preceding `file.getTokenByRangeStart(firstArg.range[0])` that is **not** wrapping an argument list, which will trip up the code in this PR.

markelog:
@texel Friendly ping :-)

markelog:
Okay, i'm closing it for now, ping us up if you want to continue work on this

PR #1957, opened 2015-11-03
Preset: use yandex preset as dependency

ikokostya:
cc @markelog


markelog:
What happened here?

ikokostya:
@markelog https://github.com/jscs-dev/node-jscs/issues/1958

PR #1959, opened 2015-11-04
Preset: WordPress requireBlocksOnNewline

GaryJones:
See https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/:

> `if`/`else`/`for`/`while`/`try` blocks should always use braces, and *always go on multiple lines*.


markelog:
LGTM

PR #1966, opened 2015-11-08
Enforce LF in gitattributes and editorconfig

ficristo:
After a fresh clone I had problem with the line breaks on my Windows PC.
Since they are already enforced as LF in the .jscsrc file I enforced them even in gitattributes and editorconfig.


markelog:
Good one! Thanks

PR #1974, opened 2015-11-11
require-camelcase-or-uppercase-identifiers strict mode

janpieterz:
Added a strict mode, rendering capitals as a first character invalid for
an identifier.

@markelog this 'should' add the strict mode we talked about on gitter.

### DIFF
--- a/lib/rules/require-camelcase-or-uppercase-identifiers.js
+++ b/lib/rules/require-camelcase-or-uppercase-identifiers.js
@@ -63,12 +63,24 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(options) {
-        assert(
-            options === true || options === 'ignoreProperties',
-            this.getOptionName() + ' option requires a true value or `ignoreProperties`'
-        );
+        if (typeof options !== 'object') {
+            assert(
+              options === true || options === 'ignoreProperties',
+              this.getOptionName() + ' option requires a true value or `ignoreProperties`'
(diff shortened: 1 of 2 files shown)


bjdixon:
Hi @janpieterz , just my opinion but I think the docstring should have 
* `Object` added to the **Types** and **Values** sections
* a note about the `"ignoreProperties"` being deprecated when used as a string 
* some examples showing the configuration object in the **Example** section or written out in full in the **Valid/Invalid for mode** sections would be helpful

There are some examples in the docstrings of `requireSpaceAfterLineComment` and `disallowQuotedKeysInObjects`

janpieterz:
All totally valid remarks, will get on that tonight, thanks for the feedback.

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -79,7 +111,13 @@ module.exports.prototype = {
         file.iterateTokensByType('Identifier', function(token) {
             var value = token.value;
             if (value.replace(/^_+|_+$/g, '').indexOf('_') === -1 || value.toUpperCase() === value) {
-                return;
+                if (this._strict) {
```
zxqfox: to reduce nested if's you can put here `if (!this._strict) { return }`

janpieterz:
Fixes implemented.

markelog:
Good job everyone!

PR #1978, opened 2015-11-14
disallowTrailingWhitespace: Fix autofixing whitespace on last line

lukeapage:


lukeapage:
I'm not sure how coverage is setup because I added lines to a function that clearly *is* tested, but coveralls is reporting the whole function untested ..
https://coveralls.io/builds/4154698/source?filename=lib%2Ftoken-assert.js#L518

lib/token-assert.js
```diff
@@ -584,6 +584,14 @@ TokenAssert.prototype.noTrailingSpaces = function(options) {
                 fixed = true;
             }
 
+            if (!fixed && precendingToken && precendingToken.type === 'EOF') {
+                precendingToken = this._file.getPrevToken(precendingToken, { includeWhitespace: true });
+                if (precendingToken && precendingToken.isWhitespace) {
```
zxqfox: You have 3 tests in outer if and 2 in inner, are you sure you need all of them?
Can you cover it? That's why coveralls reporting about coverage decrease.

lukeapage: no I'm not sure I need them. are you sure I don't? Its the classic push-pull between defensive coding and code coverage. If you know better tell me what to change or a test that checks it. Or alternatively, maybe you prefer my more radical approach in #1981?

zxqfox: Well, you don't need test `precendingToken` here if there is no purpose ;-).

I can't get how to break it if you remove `precendingToken`.

lukeapage: I'm waiting on making this change to see if the better approach in #1981 will be considered.

PR #1979, opened 2015-11-14
disallowTrailingWhitespace: fix autofix of multiple trailing spaces in a comment

lukeapage:


lukeapage:
coveralls still showing wrong coverage: https://coveralls.io/builds/4154949/source?filename=lib%2Fjs-file.js#L865

whole function down as not tested..

markelog:
> coveralls still showing wrong coverage: https://coveralls.io/builds/4154949/source?filename=lib%2Fjs-file.js#L865

and

"All checks have passed"?

lukeapage:
> and
> 
> "All checks have passed"?

just because the coverage went up, it doesn't mean its not wrong

markelog:
Fair enough, i have ping @mdevils about it, as he is maintainer of the `unit-coverage` tool

mdevils:
Hello @lukeapage. I will soon look into this issue and respond.

mdevils:
According to the idea of `unit-coverage`, to increase the coverage of `token-assert`, you need to improve tests of `token-assert` instead of testing it indirectly.

lukeapage:
so are the existing tests all in the wrong place or is the code in the
wrong place?

markelog:
You would need to add to add tests for `TokenAssert#noTrailingSpaces` not to the rule that use it, although having tests for rule is also good

lukeapage:
yeah but at the moment something is in
the wrong place.  is it a.  the tests or b.  the code

its a bit harsh that fixing a bug needs me to refactor one or the other
because you've decided the integration style of your own tests doesn't
count towards coverage.

markelog:
>  is it a.  the tests or b.  the code

Not sure what do you mean by that, if you move tests in other file nothing will change.

> its a bit harsh that fixing a bug needs me to refactor one or the other...

Not exactly, if in the future, rule `disallowTrailingWhitespace` will be removed, `TokenAssert#noTrailingSpaces` method will be left untested. Or if there will be a bug in the `TokenAssert#noTrailingSpaces` developer will be shown error in tests of the rule not of the method.

This is more strict testing policy, i give you that, but more maintainable.

lukeapage:
i think the code is in the wrong place, it doesn't seem right to assert a
whole rule in token assert.

markelog:
Other rules could use that method, that is why it is in the common place

lukeapage:
no other rule is going to check trailing space on the whole file - that is the rule!!

markelog:
We can probably move this logic to the `_fix` property, note that `token-assert` module was introduced before the `_fix`.

PR #1980, opened 2015-11-14
disallowTrailingWhitespace: fix autofixing CRLF and CR files

lukeapage:
it would be slightly better if this used the line endings option, but that doesn't autofix yet.

thinking about this and the line endings option, have you thought about/would you accept a pr that instead of altering tokens, added a post output hook to fix ? trailing whitespace and eol would be a doddle to fix post output, instead of all the complications in finding the right token..

### DIFF
--- a/lib/token-assert.js
+++ b/lib/token-assert.js
@@ -517,6 +517,7 @@ TokenAssert.prototype.noTokenBefore = function(options) {
 TokenAssert.prototype.noTrailingSpaces = function(options) {
     var ignoreEmptyLines = options.ignoreEmptyLines;
     var lines = this._file.getLines();
+    var linebreak = this._file.getLineBreaks()[0] || '\n';
 
     for (var i = 0, l = lines.length; i < l; i++) {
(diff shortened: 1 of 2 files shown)


lukeapage:
as noted in #1978 I think something isn't set up right with your coverage.

lib/token-assert.js
```diff
@@ -517,6 +517,7 @@ TokenAssert.prototype.noTokenBefore = function(options) {
 TokenAssert.prototype.noTrailingSpaces = function(options) {
     var ignoreEmptyLines = options.ignoreEmptyLines;
     var lines = this._file.getLines();
+    var linebreak = this._file.getLineBreaks()[0] || '\n';
```
zxqfox: Can you cover this branch: `|| '\n'`?

lukeapage: there is already a test without any line breaks in it.

lukeapage: if you see the coverage saying that isn't tested can you point me at it.

Issue #1854, opened 2015-10-08
Version 3.0


mdevils:
This is a roadmap ticket for version `3.0` which we are starting to work on intensively today.

The `master` branch is now for the new `3.0` version. For `2.x`-hotfixes we now have `2.x` branch.

### CST (WIP PR: #1977) (Merged #2101)
The `3.0` version consists of:
 * [x] Using `CST` for checking.
 * [ ] Using `CST` for fixing.
 * [ ] *Deprecating* most of `JsFile` functions (they are duplicative to `CST`).
 * [ ] Operating using `elements` instead of `line, column` in error reporting.
 * [ ] Document what is considered a public API (#1084, #1821)
 - [ ] Fixup skipped tests/all tests pass
 - [ ] Fix TODO comments

### CLI
* [ ] --verbose = true by default (so easier to debug issue by rule name) - https://github.com/jscs-dev/node-jscs/issues/1522
- [ ] Update the internal error message to supply version number, debug info, sample code, etc
- [x] Remove esnext, es6, esprima, esprimaOptions values (on by default / use cst) #2102 #2103 #2104

### Deprecations
* [x] remove deprecated (JSX) [XJS visitors](https://github.com/jscs-dev/node-jscs/blob/9fd87060c16e596c4f9be75e3224d2e569436891/lib/tree-iterator.js#L6-L17)
* [ ] remove all `*deprecated*` options #2110
- [ ] remove all options that aren't `allExcept: {}` and replace them with `allExcept`

### Wishlist
- ~~[ ] Rewrite in ES6 (#1469)~~
- ~~[ ] Perf metrics on rules, parsing, file handling, etc #1043~~
- [ ] A way to create an issue from the cli (with issue template or automated)

### Config Wishlist
- ~~[ ] New structure - https://github.com/jscs-dev/node-jscs/issues/698~~
- ~~[ ] Provide option for config rewrite (basically a command to upgrade a users config from 2.x to 3 automatically so there's a simple upgrade path, same with deprecated)~~
* [ ] Rules rename - https://github.com/jscs-dev/node-jscs/issues/1834  
* [ ] Use json schema for rule input validation - https://github.com/jscs-dev/node-jscs/issues/480 

/cc @markelog @hzoo @zxqfox @mrjoelkemp @mikesherov

markelog:
What is the progress on those points?

> Using CST for fixing.

Can we leave some parts on token manipulating? I think no, right?

[...]

PR #1981, opened 2015-11-14
disallowTrailingWhitespace: fix bugs by post processing

lukeapage:
As noted in #1980 I think this is a better way of doing it.

All tests and fixes from #1978, #1979, #1980 are present and also add autofixing of the validate line breaks rule.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -30,8 +30,9 @@ Errors.prototype = {
      * @param {String} message
      * @param {Number|Object} line
      * @param {Number} [column] optional if line is an object
+     * @param {Boolean} [fixed] optional if error is fixed
      */
-    add: function(message, line, column) {
+    add: function(message, line, column, fixed) {
         if (typeof line === 'object') {
             column = line.column;
             line = line.line;
@@ -40,7 +41,8 @@ Errors.prototype = {
         var errorInfo = {
             message: message,
             line: line,
-            column: column
+            column: column,
+            fixed: fixed
         };
 
         this._validateInput(errorInfo);
--- a/lib/js-file.js
+++ b/lib/js-file.js
@@ -862,6 +862,18 @@ JsFile.prototype = {
     },
 
     /**
+     * Registers a function to be run when rendering which can modify the output
+     *
+     * @param {Function} postProcessor
+     */
+    registerPostRenderProcessor: function(postProcessor) {
+        if (!this.postRenderProcessors) {
+            this.postRenderProcessors = [];
+        }
+        this.postRenderProcessors.push(postProcessor);
+    },
+
+    /**
      * Renders JS-file sources using token list.
      *
      * @returns {String}
@@ -886,6 +898,12 @@ JsFile.prototype = {
 
                 default:
                     result += token.value;
+            }
+        }
+
+        if (this.postRenderProcessors) {
+            for (i = 0; i < this.postRenderProcessors.length; i++) {
+                result = this.postRenderProcessors[i](result);
             }
         }
 
--- a/lib/rules/validate-line-breaks.js
+++ b/lib/rules/validate-line-breaks.js
@@ -69,12 +69,22 @@ module.exports.prototype = {
             return;
         }
 
+        var needsFixing = false;
+        var allowedLineBreak = this._allowedLineBreak;
+
         file.getLineBreaks().some(function(lineBreak, i) {
-            if (lineBreak !== this._allowedLineBreak) {
-                errors.add('Invalid line break', i + 1, lines[i].length);
+            if (lineBreak !== allowedLineBreak) {
+                needsFixing = true;
+                errors.add('Invalid line break', i + 1, lines[i].length, true);
                 return this._reportOncePerFile;
             }
         }, this);
-    }
 
+        if (needsFixing) {
+            file.registerPostRenderProcessor(function(output) {
+                var lines = output.split(/\r\n|\n|\r/);
+                return lines.join(allowedLineBreak);
+            });
+        }
+    }
 };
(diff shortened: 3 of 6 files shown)


markelog:
We already use two ways of autofixing - `token-assert` and `_fix` property, i would rather not introduce the third one.

Would you mind moving your autofixing logic to `_fix`, like that - https://github.com/jscs-dev/node-jscs/blob/ec8c0b0eda7cc1aa427c1a6c77ded42adab39c8c/lib/rules/validate-quote-marks.js#L136-L141 and https://github.com/jscs-dev/node-jscs/blob/ec8c0b0eda7cc1aa427c1a6c77ded42adab39c8c/lib/rules/validate-quote-marks.js#L146-L151?

markelog:
> All tests and fixes from #1978, #1979, #1980

So we don't need those pulls? We can just land this one?

lukeapage:
> We already use two ways of autofixing - token-assert and _fix property, i would rather not introduce the third one.
> 
> Would you mind moving your autofixing logic to _fix, like that - https://github.com/jscs-dev/node-jscs/blob/ec8c0b0eda7cc1aa427c1a6c77ded42adab39c8c/lib/rules/validate-quote-marks.js#L136-L141 and https://github.com/jscs-dev/node-jscs/blob/ec8c0b0eda7cc1aa427c1a6c77ded42adab39c8c/lib/rules/validate-quote-marks.js#L146-L151?

Looking at the _fix I don't see how that applies - _fix modifies tokens and the whole point of this is moving the autofixing out of tokens because modifying whitespace on these rules is alot more complicated with tokens (as you can see from the bugs I fixed in the other PR's).

if you don't want to accept a much simpler way of fixing these rules, we'll have to go back to fixing the buggy behaviour of the token fixer in #1978, #1979 and #1980 and abandon the fixing of line endings.

If I was going to fix line endings by modifying tokens I would first abstract the logic in js-file that finds the token to edit whitespace.. but then you would have "another way of autofixing" which you don't want - so that seems a bit pointless.

lukeapage:
look at where `_fix` is done - https://github.com/jscs-dev/node-jscs/blob/3e13e682d008b5cdb8b8cb00190c8dbb570f61c3/lib/string-checker.js#L266

What you should be doing is joining together your two ways of autofixing, which look in most cases to be just simple refactorings of each other and allowing a 2nd way of autofixing which just changes the file after the fact.

There is a 3rd rule that will benefit from autoprocessing autofixing - disallowTabs. I don't think it really applies to anything else.

markelog:
Postprocessing would be nice for it, but it kinda contradicts the idea of using AST/CST/tokens or parser itself. I'm not sure why we would need it.

As of the modifying tokens or AST with `_fix`, it is not hard, for example, for the file like that -
```js
var a;\n 
```

You will have following tokens array (with couple of our extension btw) - 
```js
[ Token {
    type: 'Keyword',
    value: 'var',
    start: 0,
    end: 3,
    loc: SourceLocation { start: [Object], end: [Object] },
    range: [ 0, 3 ] },
  { type: 'Whitespace', value: ' ', isWhitespace: true },
  Token {
    type: 'Identifier',
    value: 'a',
    start: 4,
    end: 5,
    loc: SourceLocation { start: [Object], end: [Object] },
    range: [ 4, 5 ] },
  Token {
    type: 'Punctuator',
    value: ';',
    start: 5,
    end: 6,
    loc: SourceLocation { start: [Object], end: [Object] },
    range: [ 5, 6 ] },
  { type: 'Whitespace', value: '\n \n', isWhitespace: true },
  { type: 'EOF',
    value: '',
    range: [ 9, 10 ],
    loc: { start: [Object], end: [Object] } } ]
```
So, you can just modify the `Whitespace` tokens, for which, we already have helper functions like [setWhitespaceBefore](https://github.com/jscs-dev/node-jscs/blob/91cc6d647e506f63103bb5bfc3560ea82b1d50e0/lib/js-file.js#L144), without introducing additional complexity to the system like post processing

lukeapage:
> but it kinda contradicts the idea of using AST/CST/tokens or parser itself. 

It doesn't contradict it, it adds to it. Just because you have a complex way of doing something, why should you do it the complex way if there is a far simpler way of dong it?

Most of the style-checks in jscs make sense to use an AST, but these 3 rules don't.

> I'm not sure why we would need to go this way if we have such powerful tools already.

Look at this PR. The AST way of doing it is 64 lines of code. Its difficult to understand and its very buggy. Even with my 3 other fixes I am not convinced it will always work.
The post processing is 10 lines of code. It will always work.

Its not about whats more powerful, its what makes sense for the job. File formatting like end of line and trailing whitespace does not make sense within the AST. If it did you would detect whitespace that way too - but you don't, you shortcut back to the source.

lukeapage:
and this isn't coming from a dislike of AST or that way of doing it - I've made significant contributions to jshint and less.js, I know alot about parsing and processing and modifying AST's.

hzoo:
I guess we just need to decide what we want to do while we move to eventually use the CST to check/fix in 3.0 https://github.com/jscs-dev/node-jscs/issues/1854

markelog:
Yeah, that is why i said that post processing is nice for such cases, but AST/CST/tokens exist to simplify any approach for modification or search of specific patterns, whitespaces are not that different from token `Punctuator` or `VariableDeclarator` AST node.

We don't use regexps so we wouldn't duplicate parser efforts and work not with source but with its representation of it. Which is much simpler approach in the long run.

Of course, i might be wrong, so in order to be sure. Let's ask other maintainers.

@hzoo, @mrjoelkemp, @mdevils thoughts?

markelog:
> I guess we just need to decide what we want to do while we move to eventually use the CST to check/fix in 3.0 #1854

Not really CST related question, post processing source output with regexps still a viable option

mdevils:
From my point of view, future of autofixing should be just an option `autofix` on the `check` method of a rule. Having `CST`, we can apply changes during checking (if necessary) and nothing will break. So I would not yet add more complexity to the system and I'd like to keep it simple till we integrate `CST`.

PR #1985, opened 2015-11-15
Add `requireEnhancedObjectLiterals`

seanpdoyle:
When `{ requireEnhancedObjectLiterals: true }`, assert that all
object declarations use the new [ES6 enhanced object literals][babel].

[babel]: http://babeljs.io/docs/learn-es2015/#enhanced-object-literals


hzoo:
We might want to think about splitting the rule into 2 separate rules later

mmun:
@hzoo Big :+1: to separating this into two rules. Is there a path to this? Perhaps `requireEnhancedObjectLiterals: 'functions' | 'properties' | true`?

markelog:
You welcome to give a try :-), for both proposals

mmun:
Thanks for the green light.

PR #1986, opened 2015-11-15
Add `requireObjectDestructuring` rule

seanpdoyle:
When `{ requireObjectDestructuring: true }`, assert that variable
declarations use object destructuring when assigning via object access.


markelog:
Would you mind adding rules documentation to your pulls?

seanpdoyle:
@markelog :ok_hand: is something like https://github.com/jscs-dev/node-jscs/blob/master/lib/rules/disallow-anonymous-functions.js#L1-L37 sufficient?

markelog:
Yep :)

seanpdoyle:
@markelog I've updated all PRs

Issue #1223, opened 2015-04-01
New Rule: requireEarlyReturn


hzoo:
From https://github.com/jscs-dev/node-jscs/issues/1071#issuecomment-74583649 (node-style-guide and idiomatic preset).

Right:
```js
function isPercentage(val) {
  if (val < 0) {
    return false;
  }

  if (val > 100) {
    return false;
  }

  return true;
}
```
Wrong:

```js
function isPercentage(val) {
  if (val >= 0) {
    if (val < 100) {
      return true;
    } else {
      return false;
    }
  } else {
    return false;
  }
}
```

Name? Maybe `requireEarlyReturnFromFunctions`?
Not sure how we would go about doing this? Checking for return's in else statements?

zxqfox:
:+1: for this.

`requireEarlyReturn` is fine for me. I think `FromFunctions` not needed because returns used in functions only and made for functions.

hzoo:
Oh haha good point.

markelog:
That one would be cool to have

hzoo:
Going to close for now.

markelog:
Okay, reopening, since it is needed for couple presets

PR #1989, opened 2015-11-19
New rule: requireEarlyReturn

ficristo:
With this PR now this config `{ requireEarlyReturn: 'noElse' }` would cause the use of else after return to be forbidden.
While this should close https://github.com/jscs-dev/node-jscs/issues/1223, it doesn't change the if conditions as seems requested in the first comment.

### DIFF
--- a/lib/rules/require-early-return.js
+++ b/lib/rules/require-early-return.js
@@ -1,3 +1,41 @@
+/**
+ * Requires to return early in a function.
+ *
+ * Types: `String`
+ *
+ * Values:
+ *  - `noElse`: disallow to use of else if the corrisponding `if` block contain a return.
+ *
+ * #### Example
+ *
+ * ```js
+ * "requireEarlyReturn": "noElse"
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * function test() {
+ *     if (x) {
+ *         return x;
+ *     }
+ *     return y;
+ * }
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * function test() {
+ *     if (x) {
+ *         return x;
+ *     } else {
+ *         return y;
+ *     }
+ * }
+ * ```
+ */
+
 var assert = require('assert');
 
 module.exports = function() {};
@@ -17,7 +55,7 @@ module.exports.prototype = {
     check: function(file, errors) {
         function addError(entity) {
             errors.add(
-                'use of else after return',
+                'Use of else after return',
                 entity.loc.start.line,
                 entity.loc.start.column
             );
@@ -34,8 +72,10 @@ module.exports.prototype = {
             }
 
             if (ifBranch.type === 'BlockStatement') {
-                if (ifBranch.body[ifBranch.body.length - 1].type === 'ReturnStatement') {
-                    return addError(node.alternate);
+                for (var i = ifBranch.body.length - 1; i >= 0; i--) {
+                    if (ifBranch.body[i].type === 'ReturnStatement') {
+                        return addError(node.alternate);
+                    }
                 }
             }
         });
--- a/lib/rules/require-early-return.js
+++ b/lib/rules/require-early-return.js
@@ -105,7 +105,7 @@ module.exports.prototype = {
                 }
             }
 
-            return addError(node.alternate);
+            return addError(file.getPrevToken(file.getFirstNodeToken(node.alternate)));
         });
     }
 };
--- a/lib/rules/require-early-return.js
+++ b/lib/rules/require-early-return.js
@@ -61,23 +61,31 @@ module.exports.prototype = {
             );
         }
 
+        function hasNodeReturn(node) {
+            if (node.type === 'BlockStatement') {
+                for (var i = node.body.length - 1; i >= 0; i--) {
+                    if (node.body[i].type === 'ReturnStatement') {
+                        return true;
+                    }
+                }
+                return false;
+            }
+            return node.type === 'ReturnStatement';
+        }
+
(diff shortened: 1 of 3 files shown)


lib/rules/require-early-return.js
```diff
@@ -0,0 +1,43 @@
+var assert = require('assert');
```
markelog: Would you mind adding some documentation for this?

markelog:
Will throw for empty `if` -

```js
if (true) {
} 
```

Will fail for `if...else` without the function scope - 
```js
      if (true) {
        return
      } else {

      }
```

Will fail if last statement is not a `return` but `return` is present in the `if` body -
```js
```

test/specs/rules/require-early-return.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    describe('with option value noElse - ', function() {
+        beforeEach(function() {
+            checker.configure({ requireEarlyReturn: 'noElse' });
+        });
+
+        var str;
+
+        // Should be errors
```
markelog: It is better to put that stuff in the description of the tests itself

lib/rules/require-early-return.js
```diff
@@ -0,0 +1,43 @@
...
+
+module.exports = function() {};
+
+module.exports.prototype = {
+    configure: function(options) {
+        assert(
+            options === 'noElse',
```
markelog: `{ requireEarlyReturn: 'noElse' }` why not `{ requireEarlyReturn: true }`?

lib/rules/require-early-return.js
```diff
@@ -0,0 +1,43 @@
...
+    },
+
+    getOptionName: function() {
+        return 'requireEarlyReturn';
+    },
+
+    check: function(file, errors) {
+        function addError(entity) {
+            errors.add(
+                'use of else after return',
```
markelog: First letter should be in uppercase

lib/rules/require-early-return.js
ficristo: I thought it was wanted some stricted behaviour, so this is to be open to add other options (eg 'strict'). Happy to change to be a boolean value.

ficristo:
> Will fail for if...else without the function scope -

Any downside to leave as is? If not I'll add some tests, otherwise which downside are there?

ficristo:
I still didn't change the value to be a boolean because I've implemented the `disallow else after return` part but not the `require early return` one.

ficristo:
This should pass, ATM it fails.
```javascript
function foo() {
  if (x) {
    bar();
  } else if (z) {
    return "foo";
  } else {
    baz(); 
  }
}
```
Need to look at it.

ficristo:
To solve the last problem i need to trasverse the tree in backwards orders.
If I'm in the `else if` I need to look at the previous `if`.
@markelog Is there an API for it?

bjdixon:
@ficristo to look at the parent ``` if ``` of the current ``` else if ``` you should be able to use ``` node.parentNode ```

ficristo:
@bjdixon Thank you.
I've update the PR with the fix and a couple of test.

lib/rules/require-early-return.js
markelog: What would `strict` option do?

I think we can switch to boolean.

lib/rules/require-early-return.js
```diff
@@ -0,0 +1,91 @@
...
+            }
+            return node.type === 'ReturnStatement';
+        }
+
+        file.iterateNodesByType('IfStatement', function(node) {
+            if (!node.alternate) {
+                return;
+            }
+
+            // Check if all the parents have a return statement, if not continue to the following IfStatement node.
```
markelog: Would you mind providing example in the comment?

lib/rules/require-early-return.js
```diff
@@ -0,0 +1,91 @@
...
+        function addError(entity) {
+            errors.add(
+                'Use of else after return',
+                entity.loc.start.line,
+                entity.loc.start.column
+            );
+        }
+
+        function hasNodeReturn(node) {
+            if (node.type === 'BlockStatement') {
```
markelog: Would be cool to have a comment here too

markelog:
Aside for couple minor things, it seems we are good to go

ficristo:
I've updated the PR.

lib/rules/require-early-return.js
ficristo: Function to check:
```javascript
function isPercentage(val) {
  if (val >= 0) {
    if (val <= 100) {
      return true;
    } else {
      return false;
    }
  } else {
    return false;
  }
}
```

With `requireEarlyReturn: 'noElse'`:
```javascript
function isPercentage(val) {
  if (val >= 0) {
    if (val <= 100) {
      return true;
    }
    return false;
  }
  return false;
}
```

With `requireEarlyReturn: 'strict' // or true`:
```javascript
function isPercentage(val) {
  if (val < 0) {
    return false;
  }

  if (val > 100) {
    return false;
  }

  return true;
}
```

markelog: How about we would use `true` value? We can introduce `strict` later. To be honest, i don't see anything criminal in the second case.

ficristo:
Now with a boolean value.

ficristo:
I've found that the column reported was wrong, now should correct.
Or at least I've copied the check from `disallow-curly-braces` file but I've found that for code like the following isn't exact:
```javascript
} // <-- There is a space after the brace.
  else
       {
```
The error is reported after the space after the first brace.
Maybe for another issue.

markelog:
Awesome work!

hzoo:
Nice!

PR #1991, opened 2015-11-19
Trailing whitespace fixes autofix, take 4

lukeapage:
Really the trailing whitespace autofixer needs to be re-written - but in a few comments like https://github.com/jscs-dev/node-jscs/pull/1981#issuecomment-158118388 people are against a simple way of doing it, preferring to instead modify tokens. It was also mentioned a few times that there will be large refactorings to fix in the CST. probably there needs to be a helper function that helps find whitespace in a way that trailing whitespace, dissallow tabs and validate end of line can all use.

Since the CST isn't available I have fixed comments on my patches to the original solution, doing the minimal amount of refactoring. I don't think this will hurt moving to the CST and the tests will be a useful addition in making sure any change is correct.

1. As noted here : https://github.com/jscs-dev/node-jscs/pull/1979#issuecomment-158210746 the unit tests don't currently count any of the logic testing trailing whitespace checking and fixing towards the code coverage. Since no other function is ever going to want to check and fix all trailing whitespace in a file other than the trailing whitespace rule, I've moved the logic to where it can count towards the code coverage, so that I can show I have increased coverage and the fixes can be merged
2. as per here: https://github.com/jscs-dev/node-jscs/pull/1978#discussion_r44961863 I've removed the check for precending token so that coverage is higher. I added a test for an empty string, just to be sure.


markelog:
Thank you the changes and for perseverance!

PR #2009, opened 2015-12-08
Update jscs-jsdoc to 1.3.0

ljharb:
This includes https://github.com/jscs-dev/jscs-jsdoc/pull/179 which includes https://github.com/yavorskiy/comment-parser/pull/24#issuecomment-159669144. Since this PR also properly includes a semver range identifier, fixes of this type (for comment-parser and jscs-jsdoc) should no longer require corresponding PRs across the dependency graph :-)

This will hopefully make jscs work on node 0.8 (that doesn't mean you have to officially support it!) and if not, will at least remove one bottleneck so I can pursue the others.

`jscs` is included because `jscs-jsdoc` apparently now requires it as a peerDep.


markelog:
> jscs is included because jscs-jsdoc apparently now requires it as a peerDep.

So be dependant on itself o_0?

markelog:
Okay, talked with @zxqfox, we decided to remove that peer dep, we will land this right after the release of the `jscs-jsdoc` 1.3.1 and push it up with jscs 2.7.

ljharb:
Awesome, that sounds great! I don't have an opinion on the peerDep, I just want 1.3.0 or higher :-D

PR #2010, opened 2015-12-08
requireCamelCaseorUpperCaseIdentifiers: Allow object restructuring with ignoreProperties set

### LINKED ISSUE #2003: requireCamelCaseOrUpperCaseIdentifiers and destructure assignment doesn't play nice
Hi!

When using the `requireCamelCaseOrUpperCaseIdentifiers` rule with the `ignoreProperties` flag turned on, JSCS still flags destructured assignments in functions

Sample JSCS config
```json
{
  "esnext": true,
  "esprima": "babel-jscs",
  "requireCamelCaseOrUpperCaseIdentifiers":{
    "ignoreProperties": true
  }
}
```

Sample failed Javascript
```javascript
  const x = ({testProperty: test_property}) => testProperty.length;
```

Results in
```
All identifiers must be camelCase or UPPER_CASE at src/test.js :
     3 |
     4 |  var x = ({testProperty: test_property}) => testProperty.length;
----------------------------------^
```

The same issue applies to default destructuring, but that is understandable.
```javascript
  const x = ({test_property, ...rest}) => test_property.length;
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28801376-requirecamelcaseoruppercaseidentifiers-and-destructure-assignment-doesn-t-play-nice?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

maherbeg:
Allows the object destructuring syntax to be used with objects that have
snake_cased properties and the `ignoreProperties` option set to true.

Fixes #2003

### DIFF
--- a/lib/rules/require-camelcase-or-uppercase-identifiers.js
+++ b/lib/rules/require-camelcase-or-uppercase-identifiers.js
@@ -129,6 +129,23 @@ module.exports.prototype = {
                     return;
                 }
 
+                /* This enables an identifier to be snake cased via the object
+                 * destructuring pattern. We must check to see if the identifier
+                 * is being used to set values into an object to determine if
+                 * this is a legal assignment.
+                 * Example: ({camelCase: snake_case}) => camelCase.length
+                 */
+                if (prevToken && prevToken.value === ':') {
+                    var node = file.getNodeByRange(token.range[0]);
+                    var parentNode = node.parentNode;
+                    if (parentNode && parentNode.type === 'Property') {
+                        var grandpa = parentNode.parentNode;
+                        if (grandpa && grandpa.type === 'ObjectPattern') {
+                            return;
+                        }
+                    }
+                }
+
                 if (prevToken && (prevToken.value === '.' ||
                     prevToken.value === 'get' || prevToken.value === 'set')) {
                     return;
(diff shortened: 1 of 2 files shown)


lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -129,6 +129,17 @@ module.exports.prototype = {
                     return;
                 }
 
+                if (prevToken && prevToken.value === ':') {
```
markelog: It get the job done! Although i wonder if there is better way. Would you mind commenting this out?

maherbeg: Of course! (I wasn't sure what the standard here was, didn't see it often).

I'm not sure if there's a better way but I'm not super deep into this code and have seen the `grandpa` pattern around. Please let me know if there is!

markelog: I guess it is fine, just add a comment and we are good to go

maherbeg: Done!

Issue #2014, opened 2015-12-09
Mention what rules do not support autofix feature


markelog:
Not that *do* support it (cause most of them are), but the ones that don't

markelog:
At this point only major and CST related bugs will be fixed.

PR #2023, opened 2015-12-15
Docs: add if autofixable to rule descriptions

hzoo:
Although this is putting if rules are fixable rather than not - https://github.com/jscs-dev/node-jscs/issues/2014

### DIFF
--- a/lib/rules/disallow-comma-before-line-break.js
+++ b/lib/rules/disallow-comma-before-line-break.js
@@ -1,5 +1,5 @@
 /**
- * Disallows commas as last token on a line in lists.
+ * Disallows commas as last token on a line in lists. (fixable)
  *
  * Type: `Boolean`, `Object`
  *
--- a/lib/rules/disallow-keywords-on-new-line.js
+++ b/lib/rules/disallow-keywords-on-new-line.js
@@ -1,5 +1,5 @@
 /**
- * Disallows placing keywords on a new line.
+ * Disallows placing keywords on a new line. (fixable)
  *
  * Types: `Array`
  *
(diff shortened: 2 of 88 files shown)


markelog:
> Although this is putting if rules are fixable rather than not - #2014

Why?

hzoo:
I think merging as is should be fine at this point @markelog?

markelog:
I think no, since this pull is pretty old and i'm would go other way around with it

hzoo:
It would make it easy to compare whats autofixable in eslint/jscs given that's how they have it

markelog:
We would need this pull to be updated then, although i'm still not sure why it can't be other way around

zxqfox:
It can be other way around, but in that way it'll take more time. Isn't it a wasting?

hzoo:
Yes unless anyone is going to actually do that work, we should just merge (my point is that it's already done)

markelog:
> my point is that it's already done

https://en.wikipedia.org/wiki/Sunk_costs

after fourth months it needs to be reevaluated anyhow, which means you have go through all the rules again and jscs has more rules that are autofixable then not

hzoo:
Why does it have to be reevaluated in four months? we aren't adding more anytime soon

markelog:
I'm not sure if all rules that right now are marked as fixable are indeed fixable, some rules didn't have any autofix tests, since they used tested versions of general methods, but with 3.0 rewrite we heavily change every single rule and removed some general methods and for some introducted `_fix` method.

Plus we need to account for new rules.

hzoo:
k

Issue #1900, opened 2015-10-20
Allow `opt_varName` like variables


tonyganch:
I would like to make a pr that allows naming optional parameters in Google style: `opt_varName`.
What is an appropriate way of reflecting this in config?
As an example: `"requireCamelCaseOrUpperCaseIdentifiers": "optionalSnakeCase"` or `"requireCamelCaseOrUpperCaseIdentifiers": "allowOpt"`.

hzoo:
If you need to add it to your own preset you can override the rule yourself http://jscs.info/overview#preset - with a value of `null` or `false` to disable the rule completely, or choose a different value.

As a separate issue, it seems like you are proposing another option - how would we determine what should or should not be allowed? Might as well add a list of exceptions.

tonyganch:
I'm proposing an extension to existing `requireCamelCaseOrUpperCaseIdentifiers`.
[Google Code Style](http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Naming#Naming) says:

> Optional function arguments start with opt_.
> Functions that take a variable number of arguments should have the last argument named var_args.

So what I'm proposing is adding a new acceptable value for `requireCamelCaseOrUpperCaseIdentifiers` option which will check that variables are named using camelCase, but also allow `opt_` prefix and `var_args` variable name.
I'm also proposing to include that new value in google preset.

tonyganch:
So, for example, these will be valid:
```
variableName
EnumName
CONST_NAME
opt_variableName
var_args
```
And these will be invalid:
```
variable_name
var_arguments
```

markelog:
How about: 
```js
requireCamelCaseOrUpperCaseIdentifiers: {
  value: '...'
  allExcept: [...]
}
```
?

Would be a good opportunity to try out [is-my-json-valid](https://github.com/mafintosh/is-my-json-valid) so you wouldn't go crazy writing all those checks in `configure`.

tonyganch:
@markelog, how should I write `allExcept` in this case for `opt_thereCanBeAnything`?

markelog:
How about glob patterns? So we wouldn't be tied to specific cases, but could be more generic, i imagine such conditions also could be used in other styles.

Also, it would be cool to justify such option in docs

tonyganch:
@markelog, do I get it right?
```
requireCamelCaseOrUpperCaseIdentifiers: {
  value: true,
  allExcept: ['opt_*', 'var_args']
}
[...]

PR #2028, opened 2015-12-18
add units suffix support to requireCamelCaseOrUpperCaseIdentifiers

pabigot:
I've got projects where I want to use `requireCamelCaseOrUpperCaseIdentifiers` but have a higher priority requirement that variables and properties that hold measurements specify the [units of measure](http://unitsofmeasure.org/ucum.html) in a suffix with an underscore separator.  For example: `ambientTemperature_cCel` or `centerFrequency_kHz`.

My solution is to augment the configuration value for `requireCamelCaseOrUpperCaseIdentifiers` to take an optional `units` property which is an array of strings or regular expressions identifying allowed units that are permitted as a underscore-separated suffix.  If an identifier has one of these suffixes and would pass the rule if the suffix was removed, then it is allowed with the suffix as well.

Is this enhancement of interest, or is there an existing way to solve this?  (I noticed issue #1900 which is similar but I specifically want the non-suffix part to still be processed by the rule.)

### DIFF
--- a/lib/rules/require-camelcase-or-uppercase-identifiers.js
+++ b/lib/rules/require-camelcase-or-uppercase-identifiers.js
@@ -115,9 +115,15 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateTokensByType('Identifier', function(token) {
             var value = token.value;
-            if (value.replace(/^_+|_+$/g, '').indexOf('_') === -1 || value.toUpperCase() === value) {
+
+            // Leading and trailing underscores signify visibility/scope and do not affect
+            // validation of the rule.  Remove them to simplify the checks.
+            var isPrivate = (value[0] === '_');
+            value = value.replace(/^_+|_+$/g, '');
+
+            if (value.indexOf('_') === -1 || value.toUpperCase() === value) {
                 if (!this._strict) {return;}
-                if (value[0].toUpperCase() !== value[0] || value[0] === '_') {
+                if (value.length === 0 || value[0].toUpperCase() !== value[0] || isPrivate) {
                     return;
                 }
             }
--- a/lib/rules/require-camelcase-or-uppercase-identifiers.js
+++ b/lib/rules/require-camelcase-or-uppercase-identifiers.js
@@ -10,6 +10,8 @@
  * - `Object`:
  *    - `ignoreProperties`:  boolean that allows an exception for object property names
  *    - `strict`: boolean that forces the first character to not be capitalized
+ *    - `allowedPrefixes`: array of String or RegExp values permitted as prefixes before `_`
+ *    - `allowedSuffixes`: array of String or RegExp values permitted as suffixes after `_`
  *
  * JSHint: [`camelcase`](http://jshint.com/docs/options/#camelcase)
  *
@@ -19,6 +21,10 @@
  * "requireCamelCaseOrUpperCaseIdentifiers": true
  *
  * "requireCamelCaseOrUpperCaseIdentifiers": {"ignoreProperties": true, "strict": true}
+ *
+ * "requireCamelCaseOrUpperCaseIdentifiers": {"allowedPrefixes": ["opt", /pfx\d+/]}
+ *
+ * "requireCamelCaseOrUpperCaseIdentifiers": {"allowedSuffixes": ["dCel", /[kMG]?Hz/]}
  * ```
  *
  * ##### Valid for mode `true`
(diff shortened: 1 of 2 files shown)


pabigot:
This could also be modified to use `allowedSuffixes` instead of `units` with the same behavior, and add `allowedPrefixes` for similar markings at the start of the identifier (e.g. `opt_`).

pabigot:
On further consideration I've decided that the generalization to prefixes and suffixes makes sense, so have updated the pull request with a version that doesn't name the options in a way that limits them to units.

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -115,9 +190,78 @@ module.exports.prototype = {
...
+             * an acceptable identifier into an unacceptable
+             * identifier so we can continue with the normal
+             * verification of whatever it produces. */
+            var i;
+            var len;
+            if (this._allowedPrefixes) {
+                for (i = 0, len = this._allowedPrefixes.length; i < len; ++i) {
+                    var prefix = this._allowedPrefixes[i];
+                    if (typeof prefix === 'string') {
+                        if (((1 + prefix.length) < value.length) &&
```
markelog: You gotta decompose here, three included `if`'s with pretty complicated conditions, whereas two of them are included in `for`.

Cyclomatic complexity should be much, much lower.

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -115,9 +190,78 @@ module.exports.prototype = {
...
+            if (match) {
+                isPrivate = true;
+                value = value.substr(match[0].length);
+            }
+            match = /_+$/.exec(value);
+            if (match) {
+                value = value.substr(0, match.index);
+            }
+
+            /* Strip at most one prefix permitted text and underscore
```
markelog: Use line instead of the block comments

markelog:
> Coverage decreased (-0.01%) to 99.291%

?

lib/rules/require-camelcase-or-uppercase-identifiers.js
pabigot: Upcoming rework factors out the conditions and bounds calculations.  The condition in for in condition remains but I can refactor further if it's still too complex.

lib/rules/require-camelcase-or-uppercase-identifiers.js
pabigot: Misled by existing block comment; fixed in rework.

pabigot:
Resubmitted with comments addressed.

Coverage decrease was probably due to dc146d8b930b9d2be99d3efcf54beed631ef6557 using an overly verbose replacement of the leading/trailing underscore strip formerly applied only within a condition.  I've refactored so it's back to a single line, but is still done separately and first on the argument that scope markers apply after prefix/suffix markers.

There was no change in unit tests for the prefix/suffix underscore stripping because it was intended to preserve existing behavior.  I've added commit bd3b160046bf4bd01a311d97b605111d09dafb4f explicitly calls out an inconsistency in the unit tests.  If the existing behavior is wrong, I'll rework; if the test description was wrong, hopefully coverage will be back up where it used to be.

pabigot:
`

pabigot:
I see the issue with coverage and will fix it when I get a chance to figure out why the obvious test doesn't fail.

pabigot:
@markelog current pull should be up to date and addresses all comments so far.

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -113,11 +188,104 @@ module.exports.prototype = {
...
+            if (((1 + prefix.length) < value.length) &&
+                (value.charAt(prefix.length) === '_') &&
+                (value.substr(0, prefix.length) === prefix)) {
+                start = 1 + prefix.length;
+            }
+            return start;
+        }
+
+        // Return undefined or the start of the unprefixed value
+        function startAfterRegExpPrefix(value, prefix) {
```
markelog: We can move those function from `check ` method

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -113,11 +188,104 @@ module.exports.prototype = {
     },
 
     check: function(file, errors) {
+
+        // Return undefined or the start of the unprefixed value
+        function startAfterStringPrefix(value, prefix) {
+            var start;
+            if (((1 + prefix.length) < value.length) &&
```
markelog: Instead of one `if` would you mind trying something like 
```js
var start = prefix.length + 1;

if (len >= value.length)) {
  return;
}
...

return start;
...

return start;
```
?

pabigot:
@markelog Good comments; the refactoring is more clear and exposed gaps in the coverage tests.

I don't believe the travis failure on Node.js 4 is related to my changes.

pabigot:
@markelog I'm going to need to implement `allExcept` as well, which will complete the functionality envisioned in #1900.  OK to add that as a new commit on the branch underlying pr, or do you want it in a different pr?

markelog:
Totally forgot about that ticket, can we introduce `allExcept` instead of adding two new options?

pabigot:
We could.  I didn't because there wasn't an obvious way to support prefixes and suffixes that still allowed the remainder of the identifier to be validated.  E.g. the proposed `opt_*` would, naively, allow `opt_this_thing` where the semantics I want for a prefix would disallow that but permit `opt_thisThing`.  Treating prefix and suffix explicitly also eliminates the need for start-of-match and end-of-match identifiers in regular expression exceptions.

So I think there's really three cases.  They could be combined into one option:

    "allExcept": {
       "match": [ "var_args", /^ignore/ ],
       "prefix": [ "opt" ],
       "suffix": [ "re", /[kMG]Hz/ ]
    }

with:

      "allExcept": [ "var_args", /^ignore/ ]

accepted when only `match` exceptions are present.

pabigot:
@markelog Looking at my last comment, there is a problem: I've been assuming I could use regular expressions, and that won't work with JSON configuration structures.  I still think the three cases are worth supporting independently, but I'll need to rework the implementation to use glob, which does seem to have the expressibility I want.

I'll convert the implementation to glob after we've resolved what the configuration should look like.

pabigot:
@markelog I think this implementation needs significant rework which depends on agreeing on the specification and semantics, so I've updated #2029 with a proposal and request for input and am closing this PR.  Thank you for your comments so far; I'll submit a new solution after we've worked out the details in the issue.

Issue #1956, opened 2015-11-03
disallowNewlineBeforeBlockStatements and multiline statements


ficristo:
Usually I write code like this:
`if (conditions) {`
so I enabled the rule `disallowNewlineBeforeBlockStatements`.

But sometimes, if the conditions are too long, I would like to enforce the brace to new line:
```javascript
if (cond1 && cond2 &&
    cond3...)
{
```

The same thing should work for `for`
```javascript
for (inits;
     conds;
     steps)
{
```

and in this case would be nice to enforce that `inits`, `conds` and `steps` are on different lines.

markelog:
```js 
if (cond1 && cond2 &&
    cond3...)
```
Check it out - http://jscs.info/rule/requireOperatorBeforeLineBreak

> The same thing should work for for

Not sure by what condition we could decide if body is too long.

ficristo:
Sorry I should have be clearer. I'm am not interested in the operators but in the curly bracket.
Usually I want it on the same line of the round bracket but when the conditions are on different lines (that was what I meant with too long...) they should be on a new line.

So if I write the if like this:
```javascript
if (a==b &&
    b==c)
```
the curly bracket should be on new line:
```javascript
if (a==b &&
    b==c)
{
```

Please reopen.

markelog:
Would you like to try to implement it?

ficristo:
@markelog I have some trouble to run `npm test` on my windows machine so If I resolve them maybe I can take a look in the future.

ficristo:
Initially I thought to implement this as an exception for both `disallowNewlineBeforeBlockStatements` and `requireNewlineBeforeBlockStatements`, but maybe it is confusing?
If I'll implement this with a new rule how should it interact with the previous `disallowNewlineBeforeBlockStatements` and `requireNewlineBeforeBlockStatements`?
@markelog

ficristo:
@markelog @hzoo gentle ping for previous comment.

ficristo:
Any ideas for the prevoius comments?

hzoo:
We want rules to be independent of each other - I think its fine to make it as an exception. Basically in the test part of the `IfStatement` or anything like that is multiline, then the brace should be on a new line

ficristo:
Is there an api to get the position of the round braces?
I tried something like this but seems very fragile...
```javascript
file.iterateNodesByType(['BlockStatement', 'ClassBody'], function(node) {
[...]

PR #2035, opened 2015-12-23
disallowNewlineBeforeBlockStatements: add allExcept multiLine exception

ficristo:
This should fix https://github.com/jscs-dev/node-jscs/issues/1956.

### DIFF
--- a/lib/rules/disallow-newline-before-block-statements.js
+++ b/lib/rules/disallow-newline-before-block-statements.js
@@ -247,6 +247,22 @@ module.exports.prototype = {
     check: function(file, errors) {
         var setting = this._setting;
         var hasMultiLineEx = this._hasMultiLineEx;
+
+        function assertSameLine(token, nextToken) {
+            errors.assert.sameLine({
+                token: token,
+                nextToken: nextToken,
+                message: 'Newline before curly brace for block statement is disallowed'
+            });
+        }
+        function assertDifferentLine(token, nextToken) {
+            errors.assert.differentLine({
+                token: token,
+                nextToken: nextToken,
+                message: 'Newline before curly brace for block statement is required'
+            });
+        }
+
         file.iterateNodesByType(['BlockStatement', 'ClassBody'], function(node) {
             if (isBareBlock(node)) {
                 return;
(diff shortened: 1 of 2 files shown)


lib/rules/disallow-newline-before-block-statements.js
```diff
@@ -193,11 +256,34 @@ module.exports.prototype = {
...
-                errors.assert.sameLine({
-                    token: prevToken,
-                    nextToken: openingBrace,
-                    message: 'Newline before curly brace for block statement is disallowed'
-                });
+                if (hasMultiLineEx === true) {
+                    var parentNode = node.parentNode;
+                    var parentNextToken = file.getFirstNodeToken(parentNode);
+                    var openingRoundBrace = file.findNextToken(parentNextToken, 'Punctuator', '(');
+
```
markelog: Why there is a newline?

lib/rules/disallow-newline-before-block-statements.js
```diff
@@ -193,11 +256,34 @@ module.exports.prototype = {
...
+                    var closingRoundBrace = file.findPrevToken(openingBrace, 'Punctuator', ')');
+
+                    if (openingRoundBrace && closingRoundBrace &&
```
markelog: There is a lot of nested `if` which usually not a good sign. Would you mind trying [requireEarlyReturn](http://jscs.info/rule/requireEarlyReturn)?

lib/rules/disallow-newline-before-block-statements.js
```diff
@@ -193,11 +256,34 @@ module.exports.prototype = {
...
+                        openingRoundBrace.loc.start.line !== closingRoundBrace.loc.end.line) {
```
markelog: Also, couple of comments would do good here

ficristo: Done.

lib/rules/disallow-newline-before-block-statements.js
ficristo: I changed the hasMultiLineEx condition to use it. I think is enough.

markelog:
Sweetsauce!

PR #2040, opened 2015-12-29
`maximumLineLength` - rewrite `functionSignature` handling, include `FunctionExpression`

### LINKED ISSUE #2032: maximumLineLength allExcept functionSignatures - broken in IIFE
With jscs-2.7.0, maximumLineLenght.allExcept = ['functionSignatures'] does not work if such a function is wrapped in an IIFE block.

I assume that's a bug, right?

Simple case:

`.jscsrc`
```json
{
  "maximumLineLength": {
    "value": 64,
    "allExcept": [
      "functionSignature"
    ]
  }
}
```

`bug.js`
```js
function tooLongButOK_spam_spam_spam_spam_spam_spam_spam_spam_spam() {
}

(function() {
  function tooLongAndNotOK_spam_spam_spam_spam_spam_spam_spam_spam() {
  }
})();

(function() {
  "use strict";
  function tooLongAndStillNotOK_spam_spam_spam_spam_spam_spam_spam() {
  }
})();
```

I'd expect jscs to run on this without errors, but I get 2 errors.

himdel:
This closes #2032 - `maximumLineLength` `allExcept: ['functionSignature']` - broken in IIFE.

I've essentially rewritten the functionSignature handling to use `iterateNodesByType` instead of the recursive `searchBodyForDecl`, and include `FunctionExpression` (as well as the previously checked `FunctionDeclaration` and `MethodDefiniton`)
    
Also abstracted common boilerplate into a local `removeLoc` function - the empty lines `.loc.start.line - 1` through `.loc.end.line - 1` logic is used in multiple places there and warrants a helper.

Spec changes:

   * add  `should not report functions within IIFE blocks`
   * change: `should` to `should not` in `.. report functions stored in variables`
   * extended that same spec to test for multiple variations - var, let or const, with or without parameters or name
   * add `should report functions within comments` - that one mostly so that we verify at least something can fail (which *may* previously have been the job of `should report functions stored in variables`)

### DIFF
--- a/lib/rules/maximum-line-length.js
+++ b/lib/rules/maximum-line-length.js
@@ -127,35 +127,23 @@ module.exports.prototype = {
         }
 
         if (this._allowFunctionSignature) {
-            var functionDeclarationLocs = [];
-            var searchBodyForDecl = function searchBodyForDecl(node) {
-                node.body.forEach(function(bodyNode) {
-                    if (bodyNode.type === 'FunctionDeclaration') {
-                        // get the loc for the `function Identifier` portion
-                        functionDeclarationLocs.push(bodyNode.id.loc);
-                        // get the locs for the params
-                        bodyNode.params.forEach(function(param) { functionDeclarationLocs.push(param.loc); });
-                        return;
+            file.iterateNodesByType('FunctionDeclaration', function(node) {
+                removeLoc(node.id);
+                node.params.forEach(removeLoc);
+            });
 
-                    } else if (bodyNode.type === 'ClassDeclaration') {
-                        searchBodyForDecl(bodyNode.body);
-                        return;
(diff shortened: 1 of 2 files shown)


himdel:
.. Also added support for `ArrowFunctionExpression`, so now even `fooooooooooo => baaaaaaaaar` is ignored.

markelog:
> Coverage decreased (-0.001%) to 99.327%

?

Also, would you mind adding example for this param?

himdel:
Coverage is decreased because 100% of shorter code is less code covered than 100% of longer code , apparently :).

> Also, would you mind adding example for this param?

Sure, will do..

himdel:
..added a few examples..

markelog:
Thank you!

PR #2047, opened 2016-01-05
requireCapitalizedConstructorsNew: New rule based on jshint newcap.

### LINKED ISSUE #2046: jscs "requireCapitalizedConstructors" and jshint "newcap" are different
Basically `requireCapitalizedConstructors` works by requiring constructors after the `new` keyword to be capitalized, where as `newcap` works by requiring the `new` keyword before capitalized function calls.

Here is a test case to show the difference:

`.jscsrc`
```json
{
	"requireCapitalizedConstructors": true
}
```

`.jshintrc`
```json
{
	"newcap": true
}
```

`test.js`
```js
function Test() {
	return null;
}
function test() {
	return null;
}

var testUpper = Test();
var testLower = new test();
```

**jscs output:**
```
$ jscs .
Constructor functions should be capitalized at ./test.js :
     7 |
     8 |var testUpper = Test();
     9 |var testLower = new test();
----------------------------^
    10 |


1 code style error found.
```

**jshint output:**
```
$ jshint .
test.js: line 8, col 17, Missing 'new' prefix when invoking a constructor.

1 error
```

Personally, I would like to have both in jscs. Is there an option I'm missing to enforce this, or could one be added?

AlexanderOMara:
Added a new rule that behaves closer to the behavior of "newcap" in jshint.

Fixes #2046

### DIFF
--- a/test/specs/rules/require-capitalized-constructors-new.js
+++ b/test/specs/rules/require-capitalized-constructors-new.js
@@ -1,0 +1,54 @@
+var Checker = require('../../../lib/checker');
+var expect = require('chai').expect;
+
+describe('rules/require-capitalized-constructors-new', function() {
+    var checker;
+
+    function baseCases() {
+        it('should report capitalized constructors without the "new" keyword', function() {
+            expect(checker.checkString('var x = Y();'))
+              .to.have.one.validation.error.from('requireCapitalizedConstructorsNew');
+        });
+
+        it('should not report capitalized construction', function() {
+            expect(checker.checkString('var x = new Y();')).to.have.no.errors();
+        });
+
+        it('should not report member expression construction', function() {
+            expect(checker.checkString('var x = ns.Y();')).to.have.no.errors();
+        });
+
+        it('should not report lowercase function calls', function() {
+            expect(checker.checkString('var x = y();')).to.have.no.errors();
+        });
+    }
+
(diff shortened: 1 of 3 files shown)


test/specs/rules/require-capitalized-constructors-new.js
```diff
@@ -0,0 +1,54 @@
+var Checker = require('../../../lib/checker');
+var expect = require('chai').expect;
+
+describe('rules/require-capitalized-constructors-new', function() {
+    var checker;
+
+    function baseCases() {
```
markelog: Not sure if we need that, if tests covering used code-paths we good either way

markelog: Means, that we could move those tests from the helper function and to the `'with `true` value'` `describe`

AlexanderOMara: So it's not necessary to run these tests on both the `true` and `allExcept` configurations? I based it on [test/specs/rules/require-capitalized-constructors.js](https://github.com/jscs-dev/node-jscs/blob/master/test/specs/rules/require-capitalized-constructors.js)

markelog: I'd say no, for my taste, if we covered it, it should be good

PR #2051, opened 2016-01-07
Fix examples for disallowkeywordsOnNewLine

dharkness:
Two of the examples had an extra space before the comment `*` which borked the generated documentation markup.


dharkness:
I recreated the change in a new branch with a new pull request that matches the commit message format.

PR #2052, opened 2016-01-07
disallowKeywordsOnNewLine: fix documentation format

dharkness:
Two of the examples had an extra space before the comment `*` which messes up the generated documentation.


hzoo:
Thanks again!

PR #2053, opened 2016-01-07
disallowQuotedKeysInObjects: fix Types and Values documentation

dharkness:
- Added Object to the list of Types
  Should String be removed since it's deprecated?

- Fixed Values list formatting
  Was creating an extra indentation level

### DIFF
--- a/lib/rules/disallow-quoted-keys-in-objects.js
+++ b/lib/rules/disallow-quoted-keys-in-objects.js
@@ -1,15 +1,15 @@
 /**
  * Disallows quoted keys in object if possible.
  *
- * Types: `String` or `Boolean`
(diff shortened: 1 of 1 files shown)


dharkness:
1. Should I remove deprecated types? (`String` in this case)

2. Which style of listing three or more types do you prefer?

* Boolean, String or Object
* Boolean or String or Object

3. Any preference for ordering the types?
  * match the values list
  * alphabetical
  * some other strict ordering

dharkness:
From chat with @markelog 
1. No, leave deprecated types in the list
2. Use the comma option (first)

hzoo:
Thanks!

PR #2057, opened 2016-01-07
`requireSemicolons`: check for `ClassProperty`

hzoo:
Ref https://github.com/babel/babel/pull/3225

### DIFF
--- a/lib/rules/require-semicolons.js
+++ b/lib/rules/require-semicolons.js
@@ -62,7 +62,8 @@ module.exports.prototype = {
             'ThrowStatement',
             'BreakStatement',
             'ContinueStatement',
-            'DebuggerStatement'
+            'DebuggerStatement',
+            'ClassProperty'
         ], function(node) {
             // ignore variable declaration inside for and for-in
             if (node.type === 'VariableDeclaration') {
--- a/test/specs/rules/require-semicolons.js
+++ b/test/specs/rules/require-semicolons.js
@@ -393,6 +393,38 @@ describe('rules/require-semicolons', function() {
         });
     });
 
+    describe('class property', function() {
+        valid([
+            'class A { asdf; }',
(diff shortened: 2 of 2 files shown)


markelog:
Wow, didn't even know you could do that. 
LGTM

michaelficarra:
It's going to be hilarious when the babel people figure out this is a bug and the semicolon-free people need a semicolon remover.

markelog:
Oh right, we need the same action for `disallowSemicolons` )

markelog:
When prop without semicolon will became a correct syntax that is

hzoo:
Well when that happens it sounds like a `good first bug`/`beginner friendly` issue to me! :smile:

michaelficarra:
@markelog It is already valid to omit semicolons after class properties in most cases.

markelog:
@michaelficarra

[Really](https://github.com/babel/babel/pull/3225#issuecomment-170056158)?

> semicolons are required always.

michaelficarra:
[Really](https://github.com/babel/babel/pull/3225#issuecomment-170155425).

markelog:
Don't see any clarification on the ASI in original link i provided, but judging by the interesting [original](https://github.com/jeffmo/es-class-fields-and-static-properties/issues/25) discussion and rules of ASI you should be right.

Which means babel pr was indeed falsy (not this pr though) and that we can add same opposite logic to `disallowSemicolons` rule.

Thanks for playing!

michaelficarra:
You're welcome. Don't forget about [this case](https://github.com/jeffmo/es-class-fields-and-static-properties/issues/26#issuecomment-170138501) which I hadn't thought of at first.

flying-sheep:
[here](https://github.com/jeffmo/es-class-fields-and-static-properties/issues/26#issuecomment-177720642)’s the ASI clarification

PR #2061, opened 2016-01-10
requireVarDeclFirst: Fixing an edge case

### LINKED ISSUE #2022: requireVarDeclFirst error when comment in function expression
Rule that has issue: "requireVarDeclFirst: true"
Code that will make jscs exhibit the issue:
.jscsrc file:

    {
      "requireVarDeclFirst": true
    }

test.js

    var test = function() {
      var color = "grey";
      // This is a comment
    };
    var test2 = "string";

node command:

    jscs test.js

The task will report an error for the line with the second assignment, but if I remove the comment in the function expression, the error goes away.

Using jscs 2.7.0
Operating system: Windows 7 64-bit

kepta:
In this edge case the comments from previous block scope were being taken and
a false positive error was thrown.

Fixes #2022 - issue

### DIFF
--- a/lib/rules/require-var-decl-first.js
+++ b/lib/rules/require-var-decl-first.js
@@ -187,7 +187,7 @@ function getCommentOffsetBetweenNodes(previousNode, currentNode, commentTokens, 
             // of the previous node and the previousNode is not the parent of currentNode
             break;
         }
-        
+
         if (comment.range[0] > currentNode.range[0] &&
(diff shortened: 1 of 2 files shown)


markelog:
> Coverage decreased (-0.02%) to 99.308%

You need to add tests

kepta:
Oh I am so sorry for that.

markelog:
Totally okay, just bring it up

markelog:
Nicely done, thank you!

Issue #2041, opened 2015-12-30
requireSpaceBeforeKeywords should ignore functions as arguments


radek-holy:
The `requireSpaceBeforeKeywords` reports an error:
> Missing space before function keyword

in the following code:
```javascript
[].forEach(function (arg) {
    console.log(arg);
});
```

I believe that such occurrence of a keyword should be an exception from this rule.

markelog:
Sounds like we should add an option for this, would you like to implement `allExcept` for this rule?

markelog:
See example - http://jscs.info/rule/disallowSpacesInsideParenthesizedExpression.html

radek-holy:
Hm, have you ever seen a code convention to write a space in this case? I am new to JavaScript so I have no idea.
My opinion is that this rule should not apply in this case at all. If you have aver seen such code, maybe I would add an opposite exception: "ignore this case by default unless an `includeParenthesized` option is enabled".
But it's just my opinion. Feel free to choose your solution. It will help too.

markelog:
Yep - https://contribute.jquery.org/style-guide/js/#spacing, but i agree, that is seems pretty esoteric and could be accomplished by other rules.

So i'm fine implementing this behaviour by default.

gpiress:
I'm interested in helping with this issue, as I had the same problem.

But just to make sure if what I thought is enough. I'm thinking in following @markelog  suggestion and implementing a third way of using this rule, i.e using an Object with the 'allExcept' property, and one could define it such as:

`requireSpaceBeforeKeywords: { allExcept: 'function' }`

And this would be sufficient for the issue to be solved, am I right?

radek-holy:
I still prefer if it was a default behaviour but yes, better this than nothing...

markelog:
I think we can make this as a default behaviour, if jquery followers would want such style, they could use different rules for that

PR #2063, opened 2016-01-11
fix Issue2041

gpiress:
**requireSpaceBeforeKeywords**

Added a new value option for the rule, an Object with an _allExcept_ property, to inform which keywords shouldn't be targeted by the rule.

Fixes issue #2041, as asked by user radek-holy

I went for the extra value approach as indicated in the issue, because I thought it would be more of a general approach that can be used across other rules and to behave similar as all other rules.

### DIFF
--- a/lib/rules/require-space-before-keywords.js
+++ b/lib/rules/require-space-before-keywords.js
@@ -1,9 +1,11 @@
 /**
  * Requires space before keyword.
  *
- * Types: `Array` or `Boolean`
+ * Types: `Array`, `Boolean` or `Object`
  *
- * Values: Array of quoted keywords or `true` to require all possible keywords to have a preceding space.
+ * Values: `true` to require all possible keywords to have a preceding space,
+ * Array of quoted keywords
+ * or an Object with the `allExcept` property set with an Array of quoted keywords.
  *
  * #### Example
  *
--- a/lib/rules/require-space-before-keywords.js
+++ b/lib/rules/require-space-before-keywords.js
@@ -41,12 +41,20 @@ module.exports = function() {};
 module.exports.prototype = {
 
     configure: function(keywords) {
+        var isValidObject = (keywords === Object(keywords) && keywords.hasOwnProperty('allExcept'));
+
         assert(
-            Array.isArray(keywords) || keywords === true,
-            this.getOptionName() + ' option requires array or true value');
+            Array.isArray(keywords) || keywords === true || isValidObject,
+            this.getOptionName() + ' option requires array, object with "allExcept" property or true value');
 
         if (keywords === true) {
             keywords = defaultKeywords;
+        }
+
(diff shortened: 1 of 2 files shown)


markelog:
We can leave `allExcept`, but i think `function` should be there by default, what do you think?

gpiress:
Ok, I believe we can leave the `function` keyword as a default excluded value. =)

Sorry for the delay in seeing this message.

gpiress:
Okay, just to clarify, this is the behaviour as it is in this pull request:

- If the option is defined as `requireSpaceBeforeKeywords: true`, it checks for all the defaultKeywords, EXCLUDING the `function` keyword, which means the code posted in the original issue would be valid, as required.

- If the option is defined as `requireSpaceBeforeKeywords: { allExcept: ['else'] }`, this would still validate the `function` keyword, checking for a space before it.

This way, it by default doesn't check for the `function` keyword, but it is possible to do so through the `allExcept` version of the rule.

Is it as expected?

markelog:
Awesome, thank you!

lib/rules/require-space-before-keywords.js
```diff
@@ -35,20 +37,28 @@
...
+            this.getOptionName() + ' option requires array, object with "allExcept" property or true value');
 
-        if (keywords === true) {
-            keywords = defaultKeywords;
+        var excludedKeywords = ignoredKeywords;
+        if (isValidObject) {
+            excludedKeywords = keywords.allExcept;
         }
 
+        keywords = defaultKeywords.filter(function(keyword) {
```
hzoo: Ok this should of been in the if statement above?

ljharb: perhaps this should be `(keywords === true ? defaultKeywords : keywords).filter`? #2135

hzoo: Nice we both found it! what do you think about just moving it into the if statement?

ljharb: ah, since `isValidObject` won't be true if it's an array (unless the array has an allExcept property, but since config is json that shouldn't be possible). Yes, that seems like the right approach.

hzoo: Ah, yeah just need to account for true value

gpiress: Sorry guys, it seems I really missed this scenario. My bad =(.

test/specs/rules/require-space-before-keywords.js
```diff
@@ -56,6 +56,16 @@ describe('rules/require-space-before-keywords', function() {
         expect(checker.checkString('if (true) x++;else x--;')).to.have.no.errors();
     });
 
+    it('should not report missing space on function by default', function() {
```
zxqfox: @markelog Why?

PR #2068, opened 2016-01-15
validateIndentation: no erroring when tabs are used but spaces are ex…

### LINKED ISSUE #1907: validateIndentation doesn't work with tabs
After jquery preset was updated - https://github.com/jscs-dev/node-jscs/commit/376dbc3dd3d018df438dc2711d75067b21606542, with adding this rule to it.

For example [this](https://github.com/jquery/jquery/blob/47ccf3daadc4b312f850502300129952e70f9d9d/src/event/alias.js) - 

```js
define( [
	"../core",
	"../event"
], function( jQuery ) {

jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
	"mousedown mouseup mousemove mouseover mouseout mouseenter mouseleave " +
	"change select submit keydown keypress keyup contextmenu" ).split( " " ),
	function( i, name ) {

	// Handle event binding
	jQuery.fn[ name ] = function( data, fn ) {
		return arguments.length > 0 ?
			this.on( name, null, data, fn ) :
			this.trigger( name );
	};
} );

jQuery.fn.extend( {
	hover: function( fnOver, fnOut ) {
		return this.mouseenter( fnOver ).mouseleave( fnOut || fnOver );
	}
} );

} );
```

will result in this
```js
Expected indentation of 2 characters at src/event/alias.js :
     9 | function( i, name ) {
    10 |
    11 | // Handle event binding
----------^
    12 | jQuery.fn[ name ] = function( data, fn ) {
    13 |  return arguments.length > 0 ?

Expected indentation of 2 characters at src/event/alias.js :
    10 |
    11 | // Handle event binding
    12 | jQuery.fn[ name ] = function( data, fn ) {
----------^
    13 |  return arguments.length > 0 ?
    14 |   this.on( name, null, data, fn ) :

Expected indentation of 1 characters at src/event/alias.js :
    15 |   this.trigger( name );
    16 | };
    17 |} );
```
Given amount of indentation bugs, i wonder if `validateIndetation` rule ever works right.

/cc @mikesherov 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27606966-validateindentation-doesn-t-work-with-tabs?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kylepaulsen:
…pected.

Also check to make sure that the tabs would make the code look indented
correctly.

Fixes #1907 

This might be a hasty fix but I thought I would throw it out there anyway.

### DIFF
--- a/test/specs/rules/validate-indentation.js
+++ b/test/specs/rules/validate-indentation.js
@@ -156,6 +156,23 @@ describe('rules/validate-indentation', function() {
         ]);
     });
 
+    it('should not report errors for using tabs when validateIndentation is a number', function() {
+        checker.configure({ validateIndentation: 4 });
+        expect(checker.checkString(' ' +
+            'function a (x) {\n' +
+            '\tif (true) {\n' +
+            '\t\tconsole.log("something");\n' +
+            '\t}\n' +
+            '\tswitch (x) {\n' +
+            '\t\tcase 1:\n' +
+            '\t\t\treturn 1\n' +
+            '\t\tdefault:\n' +
+            '\t\t\treturn 2\n' +
+            '\t}\n' +
+            '}'
+            )).to.have.no.errors();
+    });
+
     describe('includeEmptyLines', function() {
         it('should validate indentation on an empty line when includeEmptyLines is true', function() {
             checker.configure({
--- a/lib/rules/validate-indentation.js
+++ b/lib/rules/validate-indentation.js
@@ -264,6 +264,12 @@ module.exports.prototype = {
 
         function getIndentationFromLine(line) {
             var firstContent = line.search(rNotIndentChar);
+            var numTabs;
+            if (indentChar === ' ') {
+                // recompute firstContent assuming tabs are spaces.
+                numTabs = line.substring(0, firstContent).split('\t').length - 1;
+                firstContent = firstContent - numTabs + numTabs * indentSize;
+            }
             if (firstContent === -1) {
                 firstContent = line.length;
             }
@@ -563,7 +569,8 @@ module.exports.prototype = {
 
         var indentChar = this._indentChar;
         var indentSize = this._indentSize;
-        var rNotIndentChar = new RegExp('[^' + indentChar + ']');
+        var realRegexSrc = indentChar === ' ' ? indentChar + '\t' : indentChar;
+        var rNotIndentChar = new RegExp('[^' + realRegexSrc + ']');
 
         var indentStack = [0];
         var linesToCheck = file.getLines().map(function(line) {


test/specs/rules/validate-indentation.js
```diff
@@ -156,6 +156,23 @@ describe('rules/validate-indentation', function() {
         ]);
     });
 
+    it('should not report errors for using tabs when validateIndentation is a number', function() {
+        checker.configure({ validateIndentation: 4 });
```
markelog: Hm, if project uses tabs it should be `\t` not 4, maybe you trying to fix something else?

markelog: Checked the code from https://github.com/jscs-dev/node-jscs/issues/1907 it still shows same errors

kylepaulsen: The problem I was having was when you use a global .jscsrc file that has validateIndentation: 4 in it, but then you edit a file that uses tabs. Jscs will report incorrect indentation on every single line even though the indentation looks correct. Also there's the rules disallowTabs and disallowMixedSpacesAndTabs to take on related problems.

Also is the test not passing or is my code not working? I just tried it with and without and it behaved as expected.

kylepaulsen: Oh! I didn't test the code on that ticket. I just assumed my issue was the same... Oops... maybe my issue is different...

kylepaulsen: Thinking about moving this functionality under a different setting. Maybe something like:
```
"validateIndentation": {
    "value": 4,
    "allowTabs": true
}
```

Thoughts?

markelog: So you want to use spaces for indentation but still using tabs? This would be quite unconventional.
687474703a2f2f6576616465666c6f772e636f6d2f77702d636f6e74656e742f75706c6f6164732f323031312f30332f54616273537061636573426f74682e706e67

kylepaulsen: No. Absolutely not. I'm not saying mix tabs and spaces in the same file (or even the same project). I'm talking about working on a different project (maybe one you pulled from github) and that uses tabs.
Again, I have a global .jscsrc file (in my home dir), so my linter settings take over for this project I pulled.

markelog: Sorry for the delay, @mdevils, @hzoo opinions?

AuthorProxy: @kylepaulsen if you work on different project (one that can be pulled from github) it should contains it's own .jscsrc config with `validateIndentation` rules, right?

kylepaulsen: @AuthorProxy , Ideally yes, but in my case, it wasn't my project and they probably don't know jscs exists.

AuthorProxy: @kylepaulsen so it's the same that there only one rule - "no rules at all". In your case you want jscs to allow mixed tabs at document, I can't agree with this, when I open some js document I have hundreds of errors for all rules, not just for whitespace. Every time you should ignore jscs, or fix js document.

kylepaulsen: I'm not sure I understand what you mean, but all's I'm suggesting is that jscs figures out what indentation is intended in the document. Maybe it looks for what comes first: tabs or spaces, and then when it sees that, it assumes the rest of the document should use that kind of indentation. I just want to clarify: I want consistent indentation, but I also want the rule to adjust to whatever kind of indentation seems to be the main indentation of the file.

AuthorProxy: @kylepaulsen So why adding custom .jscs to root folder of project which overrides global config is not your case?

kylepaulsen: @AuthorProxy I could, but three things:
* The project owner might not want to have that file in their project.
* I have to manually do this for every project I fork
* Wouldn't it be nice to just have an option to look for the first indentation in a file and have the linter use that as the validateIndentation setting?

I have decided that this PR and my initial thoughts of how this would work are incorrect. If I ever have time, I'll make a PR for an option to check for the first indentation in a file and then error on inconsistent indentations thereafter - this is what I really wanted in the first place I think.

PR #2070, opened 2016-01-16
Preset: enforce Airbnb rule 7.11 (function signatures)

redoPop:
This PR addresses some disagreements between [`airbnb.json`](https://github.com/jscs-dev/node-jscs/blob/c704e9e0713c5462a61dac309a6e6abd30098545/presets/airbnb.json) and [Airbnb rule 7.11](https://github.com/airbnb/javascript/blob/c2fd8fbbd6cf058244024db3ab791ddba15567a1/README.md#7.11):

```javascript
// bad
const f = function(){};
const g = function (){};
const h = function() {};

// good
const x = function () {};
const y = function a() {};
```

* Spaces before opening round braces in anonymous function expressions should be required, not [disallowed](https://github.com/jscs-dev/node-jscs/blob/c704e9e0713c5462a61dac309a6e6abd30098545/presets/airbnb.json#L9-L11)
	* Shorthand is an exception, per [Airbnb rule 3.5](https://github.com/airbnb/javascript/blob/c2fd8fbbd6cf058244024db3ab791ddba15567a1/README.md#3.5)
* Due to the need for differentiated behavior between named vs. anonymous function expressions, the broader rule `disallowSpacesInFunctionExpression` should _not_ be set

I've updated the preset test file (`airbnb.js`) accordingly.

### DIFF
--- a/presets/airbnb.json
+++ b/presets/airbnb.json
@@ -3,12 +3,6 @@
     "disallowSpacesInNamedFunctionExpression": {
         "beforeOpeningRoundBrace": true
     },
-    "disallowSpacesInFunctionExpression": {
-        "beforeOpeningRoundBrace": true
-    },
-    "disallowSpacesInAnonymousFunctionExpression": {
-        "beforeOpeningRoundBrace": true
-    },
     "disallowSpacesInFunctionDeclaration": {
         "beforeOpeningRoundBrace": true
     },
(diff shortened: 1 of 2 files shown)


presets/airbnb.json
```diff
@@ -3,12 +3,6 @@
     "disallowSpacesInNamedFunctionExpression": {
```
hzoo: Sounds like we should remove this as well right?

redoPop: I think this rule should be retained; named function expressions seem to be handled differently from anonymous ones in Airbnb's style guide:

```javascript
// good
const x = function () {};
const y = function a() {};
```

Where `x` is considered an example of an anonymous function expression (where the space before open paren is required) and `y` is considered a named function expression (where the space before open paren is disallowed).

markelog:
@hzoo LGTM?

hzoo:
Yeah :+1: - Thanks @redoPop!

Issue #2029, opened 2015-12-19
requireCamelCaseOrUpperCaseIdentifiers: allow specific prefixes and suffixes to be ignored


pabigot:
Proposal for rule enhancement:

I've got projects where I want to use `requireCamelCaseOrUpperCaseIdentifiers `but have a higher priority requirement that variables and properties that hold measurements specify the [units of measure](http://unitsofmeasure.org/ucum.html) in a suffix with an underscore separator. For example: `ambientTemperature_cCel` or `centerFrequency_kHz`.

Issue #1900 expresses a similar desire for specific prefixes to be ignored, but the proposed solution may involve wildcards that would not validate the remainder of the identifier.

The proposed solution is to augment the configuration value for `requireCamelCaseOrUpperCaseIdentifiers` to take options `allowedPrefixes` and `allowedSuffixes`.  Each of those is an array of strings or regular expressions identifying text sequences that are permitted as a prefix followed by an underscore (for `allowedPrefixes` or a suffix after an underscore (for `allowedSuffixes`).  If an identifier has one of these prefixes (suffixes) and would pass the rule if the prefix (suffix) was removed, then it is allowed with the prefix (suffix) as well.

Pull request #2028 will have an implementation of this just as soon as I know what issue number this is so I can reference it in the commit message.

ChrisPearce:
This would also be very useful when referencing NPM script config values from within Node JS (e.g. var myProp = process.env.npm_package_config_MY_ENVIRONMENT_VARIABLE).

pabigot:
@markelog @ChrisPearce @tonyganch Pull request #2028 which had my initial attempt at this has gotten complex and involves questions of how to specify the configuration, so I want to move attention back here where we can work out the semantics and interface in a more visible area.  If there are others who should be invited to participate please do so.

As I noted in the original description, the behavior I want is to specify arbitrary prefix and suffix values with separating underscores that are stripped off identifiers before submitting the remainder to the validation of `requireCamelCaseOrUpperCaseIdentifiers`.

In #1900 and a comment in #2028 it was proposed to use `allExcept` for something like this, but I believe that name implies completely excluding a matching identifier from the rule.  As such I don't believe the prefix and suffix exceptions should be incorporated into the `allExcept` option, though I do need that capability as well.

Thus I propose three additional configuration options, all arrays, and all containing string, RegExp, or [minimatch](https://github.com/isaacs/minimatch) string values.  For example

[...]

PR #2071, opened 2016-01-16
requireCamelCaseOrUpperCaseIdentifiers: add exception options

pabigot:
>Some coding guidelines might want optional parameters to start with `opt_` while still checking conformance of the remainder of the name, or have specific identifiers like `var_args` that are exceptions.
>
>Other projects may want to use this rule but require that values that hold measurements carry a suffix specifying the units of measurement; e.g. `ambientTemperature_dCel` or `centerFrequency_kHz`.  The interpretation of units can be misleading when conjoined with the identifier without a separating underscore.
>
>This commit adds options `allExcept`, `allowedPrefixes`, and `allowedSuffixes`.  All three recognize an array of strings, RegExps, or ESTree RegExpLiteral objects that identify strings or string sequences. Identifiers that match strings/patterns in `allExcept` are excluded from validation by the rule.  Identifiers that start (end) with a match from `allowedPrefixes` (`allowedSuffixes`) are checked after the match is removed from the identifier.

Resubmission following comments in the original pull request #2028 and feature issue #2029.  The primary changes are:

* Add [ESTree RegExpLiteral](https://github.com/estree/estree/blob/master/spec.md#regexpliteral) objects so regular expressions can be expressed in JSON config specs;
* Incorporate the prefix/suffix separator character (generally but not always `_`) into the strings and patterns;
* Add `allExcept` to complete the functionality required to close issue #1900.

### DIFF
--- a/lib/rules/require-camelcase-or-uppercase-identifiers.js
+++ b/lib/rules/require-camelcase-or-uppercase-identifiers.js
@@ -115,9 +115,15 @@ module.exports.prototype = {
     check: function(file, errors) {
         file.iterateTokensByType('Identifier', function(token) {
             var value = token.value;
-            if (value.replace(/^_+|_+$/g, '').indexOf('_') === -1 || value.toUpperCase() === value) {
+
+            // Leading and trailing underscores signify visibility/scope and do not affect
+            // validation of the rule.  Remove them to simplify the checks.
+            var isPrivate = (value[0] === '_');
+            value = value.replace(/^_+|_+$/g, '');
+
+            if (value.indexOf('_') === -1 || value.toUpperCase() === value) {
                 if (!this._strict) {return;}
-                if (value[0].toUpperCase() !== value[0] || value[0] === '_') {
+                if (value.length === 0 || value[0].toUpperCase() !== value[0] || isPrivate) {
                     return;
                 }
             }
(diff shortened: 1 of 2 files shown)


lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -79,10 +88,161 @@
...
+            try {
+                // ESTree RegExpLiteral ok if it produces RegExp
+                rv.push(new RegExp(elt.regex.pattern, elt.regex.flags || ''));
+            } catch (e) {
+                // Not a valid RegExpLiteral
+                rv = null;
+            }
+        } else {
+            // Unknown value
+            rv = null;
```
gibson042: An `assert` (and corresponding exception) is probably better than these `rv = null` silent failures.

gibson042: Nevermind, I see that it's just handled elsewhere.

gibson042:
I like this a lot.

markelog:
@pabigot Nicely done! note @gibson042 comment, that is the high praise :)

pabigot:
Thanks; I like it too.

Please let me know if you have additional changes you'd like made, and if you intend to merge the amend commit with the main one or would prefer I do it.

lib/rules/require-camelcase-or-uppercase-identifiers.js
```diff
@@ -79,10 +88,161 @@
  * var snake_case = { SnakeCase: 6 };
  * ```
  *
+ * ##### Valid for `{ allowedPrefix: ["opt_", /pfx\d+_/] }`
```
markelog: I think we need how we could store regexp in json

markelog: Oh, i see, you already did that :)

PR #2103, opened 2016-01-28
Misc: removing esnext since it will be on by default

hzoo:


hzoo:
Will just merge https://github.com/jscs-dev/node-jscs/pull/2104

PR #2108, opened 2016-01-29
Test cst with runtime plugin

hzoo:
This is to support node 0.10


hzoo:
Added in https://github.com/jscs-dev/node-jscs/commit/b492de717ef1faeaac31b49965839015d4608ae3

PR #2114, opened 2016-02-01
Pragma Index

mdevils:
**WIP**

* [x] PragmaIndex utility implementation.
* [ ] Integration into StringChecker.


lib/pragma-index.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+PragmaIndex.prototype._buildIndex = function(firstToken) {
+    var tokens = [];
+    var index = [];
+    var currentPosition = 0;
+    var currentToken = firstToken;
+    var lastBlockState = {'*': true};
+    var tokenState;
+
+    while (currentToken) {
```
markelog: Could you decompose it? And add comments, right now it is veery hard to grasp

hzoo:
I don't think you can switch branches in github so closing (make the new PR to master instead of 3.0) :smile:

PR #2115, opened 2016-02-02
disallow-nested-ternaries: fixes incorrect option type

### LINKED ISSUE #2092: disallowNestedTernaries - invalid option types described on website
The `disallowNestedTernaries` rule described [here](http://jscs.info/rule/disallowNestedTernaries) indicates that values for the rule take a `boolean` or an `integer`.  However, the example and inspection of the source indicates it takes a `boolean` or an `object`, where the `object` has a `maxLevel` integer attribute.

cookch10:
disallow-space-after-object-keys.js: adds missing option type

Fixes #2092 - disallowNestedTernaries - invalid option types described on website

### DIFF
--- a/lib/rules/disallow-nested-ternaries.js
+++ b/lib/rules/disallow-nested-ternaries.js
@@ -1,9 +1,9 @@
 /**
  * Disallows nested ternaries.
  *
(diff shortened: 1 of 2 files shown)


markelog:
Thank you!

PR #2130, opened 2016-02-11
safeContextKeyword: fix destructuring handling

### LINKED ISSUE #2119: Bug with safeContextKeyword and destructuring with multiple-variable declarations
The following should not warn, but currently does:

```js
const a = 1, { smth } = this;
```

#1699 did not take multi-var declarations into account.

I'm looking into fixing this myself, but still wrapping my head around how the tokenizing works and figured I'd open the issue if its a quick fix for someone else.

jimf:
Update `safeContextKeyword` to handle destructuring when multiple
variables are being declared and the destructuring is not the first.

Fixes #2119

### DIFF
--- a/lib/rules/safe-context-keyword.js
+++ b/lib/rules/safe-context-keyword.js
@@ -50,15 +50,14 @@ module.exports.prototype = {
 
         // var that = this
         file.iterateNodesByType('VariableDeclaration', function(node) {
-            var firstToken = node.firstToken;
-
-            // Miss destructing assignment (#1699)
-            if (file.getNextToken(firstToken).value === '{') {
-                return;
-            }
 
             for (var i = 0; i < node.declarations.length; i++) {
                 var decl = node.declarations[i];
+
+                // Miss destructing assignment (#1699, #2119)
+                if (file.getNextToken(file.getPrevToken(decl)).value === '{') {
+                    continue;
+                }
 
                 // decl.init === null in case of "var foo;"
                 if (decl.init &&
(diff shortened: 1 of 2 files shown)


jimf:
There might be a better way to achieve this. I was fumbling around a bit in the debugger to get to this solution, but the `getNextToken(getPrevToken())` business admittedly feels a little clunky.

lib/rules/safe-context-keyword.js
```diff
@@ -50,16 +50,15 @@ module.exports.prototype = {
...
-            // Miss destructing assignment (#1699)
-            if (file.getNextToken(firstToken).value === '{') {
-                return;
-            }
 
             for (var i = 0; i < node.declarations.length; i++) {
                 var decl = node.declarations[i];
 
+                // Miss destructing assignment (#1699, #2119)
+                if (file.getNextToken(file.getPrevToken(decl)).value === '{') {
```
markelog: How about 

```js
if (decl.firstChild.type === 'ObjectPattern') {
```
?

markelog:
> Coverage decreased (-0.002%) to 88.698%

?

You check the coverage by running 
```shell
npm run coverage-html && open coverage.html
```

lib/rules/safe-context-keyword.js
markelog: Can't port that to the `2.x` branch though, so it might have to wait until 3.0 release :/

jimf: +1. I like that, and I've confirmed it works. Would you like me to update my PR, or hold off due to your 3.0 point?

jimf:
Not sure what to say about the coverage. I've only touched `safe-context-keyword.js`, and it has 100% coverage.

image

markelog:
Had issues with rebasing it against `2.x` branch feel free to create one for 2.x

PR #2132, opened 2016-02-12
Misc: fix dynamic test requires

jimf:
When reading in test files from the directory, omit non-js files before attempting to `require` them (relieves annoyances while using any editor(s) that store backup files of anything being worked on within the same directory, e.g., Vim .swp files).


jimf:
Maybe worth explaining as well: despite this test being within a `describe.skip`, the dynamic `it` block generation still gets run.

markelog:
Thank you

PR #2133, opened 2016-02-14
requireNumericLiterals: Prevent erroring when first argument is an Identifier.

rwjblue:
Prior to this fix, the following would trigger an error:

```
let foo = '123';
parseInt(foo, 16);
```

After this change, an error is only triggered when using an actual literal (as intended and indicated by the rule name).

Addresses https://github.com/DockYard/ember-suave/pull/80#issuecomment-183779369.

/cc @hzoo


markelog:
Thanks

hzoo:
Cool, thanks @rwjblue!

PR #2136, opened 2016-02-15
requireSpaceBeforeKeywords: hotfix for predefined list of keywords

### LINKED ISSUE #2135: v2.10.0 - `requireSpaceBeforeKeywords` reports on a keyword not in its list
In my config file, I have:
```
       "requireSpaceBeforeKeywords": [
         "do",
         "for",
         "if",
         "else",
         "switch",
         "case",
         "try",
         "catch",
         "finally",
         "while",
         "with",
         "return"
       ],
```

However, upon upgrading to v2.10.0, I get output like:
```
Missing space before "typeof" keyword at index.js :
    32 |module.exports = function isCallable(value) {
    33 | if (!value) { return false; }
    34 | if (typeof value !== 'function' && typeof value !== 'object') { return false; }
-------------^
    35 | if (hasToStringTag) { return tryFunctionObject(value); }
    36 | if (isES6ClassFn(value)) { return false; }

Missing space before "typeof" keyword at test.js :
    96 |test('Typed Arrays', function (st) {
    97 | forEach(typedArrayNames, function (typedArray) {
    98 |  if (typeof global[typedArray] !== 'undefined') {
--------------^
    99 |   st.ok(isCallable(global[typedArray]), typedArray + ' is callable');
   100 |  }
```

Note that "typeof" is not in my array of things I want a space before, and this was not an error in v2.9.x.

zxqfox:
Fixes #2135

/cc @hzoo @ljharb

### DIFF
--- a/lib/rules/require-space-before-keywords.js
+++ b/lib/rules/require-space-before-keywords.js
@@ -55,11 +55,13 @@ module.exports.prototype = {
             excludedKeywords = keywords.allExcept;
         }
 
-        if (!Array.isArray(keywords)) {
-            keywords = defaultKeywords.filter(function(keyword) {
-                return (excludedKeywords.indexOf(keyword) === -1);
-            });
+        if (keywords === true || isValidObject) {
+            keywords = defaultKeywords;
         }
+
+        keywords = keywords.filter(function(keyword) {
+            return (excludedKeywords.indexOf(keyword) === -1);
+        });
 
         this._keywords = keywords;
     },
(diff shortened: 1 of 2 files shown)


lib/rules/require-space-before-keywords.js
```diff
@@ -55,7 +55,7 @@ module.exports.prototype = {
             excludedKeywords = keywords.allExcept;
         }
 
-        keywords = defaultKeywords.filter(function(keyword) {
+        keywords = (keywords === true ? defaultKeywords : keywords).filter(function(keyword) {
```
ljharb: this will work but @hzoo pointed out that instead, this statement should instead be moved into the preceding "if" block.

ljharb:
Looks like this can be closed with @markelog's 594a4eed93fd51baec4c5f4f581c13ad348efa83?

hzoo:
Yeah

PR #2138, opened 2016-02-16
New rule: require-newline-before-single-statements

### LINKED ISSUE #1829: [[FEAT]] requireNewlineBeforeSingleStatements
Consider these options:
```json
{
    "requireNewlineBeforeBlockStatements": false,
    "requireCurlyBraces": false
}
```
I'd like to enforce this style:
```js
// bad
if (x) doX();
else doY();

// good
if (x)
    doX();
else
    doY();
```

or is this a case of me being too peculiar?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27088117-feat-requirenewlinebeforesinglestatements?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F281640&utm_medium=issues&utm_source=github).
</bountysource-plugin>

schempy:
requireNewLineBeforeSingleStatements: New rule

Enforce newlines for expression statements within if/else 

Fixes #1829 - issue

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -1026,7 +1026,7 @@ Configuration.prototype.registerDefaultRules = function() {
 
     this.registerRule(require('../rules/require-early-return'));
 
-    this.registerRule(require('../rules/require-newline-before-single-statements'));
+    this.registerRule(require('../rules/require-newline-before-single-statements-in-if'));
 };
 
 /**
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -1025,6 +1025,8 @@ Configuration.prototype.registerDefaultRules = function() {
     this.registerRule(require('../rules/require-aligned-multiline-params'));
 
     this.registerRule(require('../rules/require-early-return'));
+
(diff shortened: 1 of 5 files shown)


markelog:
Hm, maybe `requireNewlineBeforeSingleStatementsInIf` name could be more explanatory?

schempy:
Sure, that makes sense. Could also allow options to work with if/else, for etc... Don't want to make it more complicated than it needs. What do you think?

markelog:
> Could also allow options to work with if/else, for etc...

I think we can consider using different rule for that?

schempy:
Sure another rule for those makes sense. I'll change the rule name to be ```requireNewlineBeforeSingleStatementsInIf```

schempy:
I changed the rule name to ```requireNewlineBeforeSingleStatementsInIf``` If things look OK I can squash the commits if you'd like.

PR #2152, opened 2016-03-01
airbnb preset - rule #8.2 - omit the braces

iilei:
> If the function body consists of a single expression, omit the braces and use the implicit return. Otherwise, keep the braces and use a return statement.

[See rule #8.2](https://github.com/airbnb/javascript#8.2)


markelog:
`requireShorthandArrowFunctions` is good but `disallowParenthesesAroundArrowParam ` isn't, since that rule is strict, whereas this -
```js
[1, 2, 3].map((number) => {
  const nextNumber = number + 1;
  return `A string containing the ${nextNumber}.`;
});
```
is allowed.
/cc @hzoo 

Thank you!

PR #2153, opened 2016-03-01
Misc: remove browserified version

### LINKED ISSUE #2150: Browser version
I think you shouldn't publish 4mb of browserified bundle on npm. Nobody will use so big lib.

markelog:
Since we don't run tests on it anyway, plus anyone can create this version if needed
Fixes #2150

/cc @mdevils, @hzoo


hzoo:
:+1: But I think if you ctrl f for `browser` there's probably more

like `jscs-browser.js` in gitignore, jshintignore, deleting from comments

Does that mean we can remove `this.registerRule(require...`?

markelog:
> But I think if you ctrl f for browser there's probably more
like jscs-browser.js in gitignore, jshintignore, deleting from comments

Oh yeah, good catch

> Does that mean we can remove this.registerRule(require...?

It sure does!

PR #2160, opened 2016-03-04
New rule: requireUseStrict to ensure strict mode is used

erikvold:


test/specs/rules/require-use-strict.js
```diff
@@ -0,0 +1,93 @@
+var Checker = require('../../../lib/checker');
+var expect = require('chai').expect;
+
+var funcX = 'function x() {"use strict";\nreturn true;}';
+var funcY = 'function x() {return true;}';
+
+function makeChecker(rules) {
```
markelog: Would be better to put this in `beforeEach`

lib/rules/require-use-strict.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * Requires `'use strict';` statements
+ *
+ * Values:
+ *  - `true` for default behavior (require 'use strict' statements for files)
+ *  - `Object`:
+ *    - `'functions'` boolean - require `'use strict';` within functions.
```
markelog: It seems outdated?

lib/rules/require-use-strict.js
```diff
@@ -0,0 +1,98 @@
...
+ *    - `'files'` boolean - require `'use strict';` at the top of each file
```
markelog: If we only have one option, it seems `object` value is redundant

erikvold:
Thanks I did forget to update the docs, I'll fix that!

erikvold:
@markelog ready for another review

test/specs/rules/require-use-strict.js
markelog: Sounds like it would easier to do it like [this](https://github.com/jscs-dev/node-jscs/blob/fbe9bf4ad5ab3e2645affee07a5fc5ff19cea35f/test/specs/rules/disallow-arrow-functions.js#L7)

markelog:
Updated tests a bit, thanks

PR #2164, opened 2016-03-05
New Rule: (require | disallow)SpacesInsideImportedObjectBraces

### LINKED ISSUE #2151: Require spaces for object destructuring in imports, alignment cases for imports
What about rule for require spaces for import cases?

```
invalid:
import {foo,bar} from 'foobar';

valid:
import { foo,bar } from 'foobar';
```

and(maybe for another rule):
```
invalid:
import { foo,bar } from 'foobar';
import { foo as f,bar } from 'foobar';

valid:
import { foo, bar } from 'foobar';
import { foo as f, bar } from 'foobar';
```
and what about alignment cases?

```
invalid:
import { foo, bar } from 'foobar';
import { a, b, c, d, e } from 'alphabet';

valid:
import { foo, bar }      from 'foobar';
import { a, b, c, d, e } from 'alphabet';

```
but I'm not sure how we can detect alignment for very long import strings:

```
import { foo, bar} from 'foobar';
import { a as SomethingWithVeryLongNameA, b as SomethingWithVeryLongNameB, c as SomethingWithVeryLongNameC } from 'alphabet'
```

fxmaxvl:
Restrictions for spaces inside imported objects

Fixes #2151

### DIFF
--- a/presets/airbnb.json
+++ b/presets/airbnb.json
@@ -76,5 +76,6 @@
     "validateIndentation": 2,
     "maximumLineLength": 100,
     "disallowArrayDestructuringReturn": true,
-    "requireShorthandArrowFunctions": true
+    "requireShorthandArrowFunctions": true,
+    "requireSpacesInsideImportedObjectBraces": true
 }
--- a/grouping.json
+++ b/grouping.json
@@ -164,7 +164,9 @@
     "requireArrayDestructuring",
     "disallowVar",
     "requireSpaceBeforeDestructuredValues",
-    "disallowArrayDestructuringReturn"
+    "disallowArrayDestructuringReturn",
+    "requireSpacesInsideImportedObjectBraces",
+    "disallowSpacesInsideImportedObjectBraces"
   ],
   "Everything else": [
     "requireUseStrict",
--- a/lib/rules/disallow-spaces-inside-imported-object-braces.js
+++ b/lib/rules/disallow-spaces-inside-imported-object-braces.js
@@ -1,0 +1,82 @@
+/**
+ * Disallow space after opening object curly brace and before closing in import statements.
+ *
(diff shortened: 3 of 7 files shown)


fxmaxvl:
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
??

fxmaxvl:
review?

markelog:
It happens with coveralls sometimes, for example, now it all the way down...

Would you like to also update presets? At least `airbnb` is candidate.

fxmaxvl:
ok

presets/airbnb.json
```diff
@@ -76,5 +76,6 @@
     "validateIndentation": 2,
     "maximumLineLength": 100,
     "disallowArrayDestructuringReturn": true,
-    "requireShorthandArrowFunctions": true
+    "requireShorthandArrowFunctions": true,
+    "requireSpacesInsideImportedObjectBraces": true
```
markelog: It seems we need add an some `import` statement to https://github.com/jscs-dev/node-jscs/blob/fbe9bf4ad5ab3e2645affee07a5fc5ff19cea35f/test/data/options/preset/airbnb.js ?

fxmaxvl: but, how can i write  a correct case with import there? had trouble with it

markelog: Adding `import { resolve } from "path";` in the begging ?

fxmaxvl: Oh, tnx) I had doubts about the use of real modules for this case

markelog: That file will not be executed anyway, you can even add a non-existent module, will still work

markelog:
Awesome, thank you

PR #996, opened 2015-02-05
Preset: add "requireSpaceBetweenArguments" rule to all presets

markelog:
Also corrects link to yandex code-style

Fixes #940

PR #2171, opened 2016-03-08
Preset: Add 'requireSpaceBetweenArguments' to google preset

jmerrifield:
It's not explicitly mentioned in [the documentation](http://google.github.io/styleguide/javascriptguide.xml) but the examples consistently use a space between arguments.

#996 added this rule to all other presets, but not the google one.


markelog:
Thank you!

PR #2178, opened 2016-03-16
requireVarDeclFirst: use CST

zxqfox:
/cc @markelog @hzoo

### DIFF
--- a/lib/rules/require-var-decl-first.js
+++ b/lib/rules/require-var-decl-first.js
@@ -117,12 +117,31 @@ function isScopeElement(elem) {
         );
 }
 
+/**
+ * Checks for allowed elements before variable declaration
+ * @param {Object} elem
+ * @returns {Boolean}
+ */
+function isRelevantElement(elem) {
+    // Allow comments and whitespaces.
+    var itIs = elem.isComment || elem.isWhitespace;
+
+    // Variable declaration itself.
+    itIs = itIs || elem.type === 'VariableDeclaration';
+
+    // For '{' in `function() { var a; }`.
+    itIs = itIs || elem.type === 'Punctuator';
+
+    // Skip 'use strict' statements at the beginning.
+    itIs = itIs || elem.type === 'ExpressionStatement' && elem.expression.value === 'use strict';
+
+    return itIs;
+}
+
 function isVarDeclFirst(varDecl) {
-    var elem = varDecl.previousSibling;
-    while (elem && (elem.isComment || elem.isWhitespace ||
-        elem.type === 'VariableDeclaration' ||
-        elem.type === 'Punctuator' ||
-        elem.type === 'ExpressionStatement' && elem.expression.value === 'use strict')) {
+    var elem = varDecl;
+    // All elements should be relevant.
+    while (elem && (elem.isComment || elem.isWhitespace || isRelevantElement(elem))) {
         elem = elem.previousSibling;
     }
 
--- a/lib/rules/require-var-decl-first.js
+++ b/lib/rules/require-var-decl-first.js
@@ -108,21 +108,22 @@
 
 var assert = require('assert');
 
+function isScopeElement(elem) {
+    return elem.type === 'Program' ||
+        elem.type === 'BlockStatement' && (
+            elem.parentElement.type === 'FunctionExpression' ||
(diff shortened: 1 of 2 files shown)


test/specs/rules/require-var-decl-first.js
```diff
@@ -200,10 +201,6 @@ describe.skip('rules/require-var-decl-first', function() {
                     testDeclStatements(checker, '"use strict";\nvar a;', 0);
                 });
 
-                it('should not return errors for single const var declaration at top of program scope', function() {
```
markelog: Why we remove those?

lib/rules/require-var-decl-first.js
```diff
@@ -108,111 +108,24 @@
...
-            // Skip comments that occurred before the previous node
-            continue;
+function isVarDeclFirst(varDecl) {
+    var elem = varDecl.previousSibling;
+    while (elem && (elem.isComment || elem.isWhitespace ||
+        elem.type === 'VariableDeclaration' ||
+        elem.type === 'Punctuator' ||
+        elem.type === 'ExpressionStatement' && elem.expression.value === 'use strict')) {
+        // We should goes up from blocks for var declaration.
+        if (elem.type === 'Punctuator' &&
```
markelog: Can we try a different approach with `continue` statements?

Like -
```js
if (elem.type !== 'Punctuator) {
   elem = elem.previousSibling;
   continue;
}
```

?

lib/rules/require-var-decl-first.js
```diff
@@ -108,111 +108,24 @@
...
-            // Stop processing comments that are within multiple declarators in a single variable declaration
-            break;
-        }
-
-        if (previousNode.range[0] >= comment.range[1]) {
-            // Skip comments that occurred before the previous node
-            continue;
+function isVarDeclFirst(varDecl) {
+    var elem = varDecl.previousSibling;
+    while (elem && (elem.isComment || elem.isWhitespace ||
```
markelog: That condition could be moved to a function right? Something like -
```js
while(somethingOrOther(elem)) {
...
```
?

markelog:
Beautiful, with CST that thing really takes shape!

test/specs/rules/require-var-decl-first.js
zxqfox: We have test for const and let above. And this rule ignores them at all.

lib/rules/require-var-decl-first.js
zxqfox: I'm not sure we even need to allow `var`s inside blocks. I mean, if there is `var` — it should be placed FIRST in scope, while scope for `var` is some kind of function node or the program.

Fixed in the second commit. Please check it out!

lib/rules/require-var-decl-first.js
zxqfox: Please, look at the code. I'll move it out if need, but looks like not needed.

lib/rules/require-var-decl-first.js
```diff
@@ -108,21 +108,22 @@
...
+        elem.type === 'BlockStatement' && (
+            elem.parentElement.type === 'FunctionExpression' ||
+            elem.parentElement.type === 'FunctionDeclaration' ||
+            elem.parentElement.type === 'ArrowFunctionExpression'
+        );
+}
+
 function isVarDeclFirst(varDecl) {
     var elem = varDecl.previousSibling;
     while (elem && (elem.isComment || elem.isWhitespace ||
```
markelog: Much better!

How about we go further with it - 
```js
 // Probably mess up some condition
 function isRelevant /*or some better name*/ (elem) {
    var itIs = false;

    if (!elem) {
        return itIs;
    }

    if (elem.expression.value !== 'use strict') {
        return itIs;
    }

    itIs = elem.isComment || elem.isWhitespace;
    itIs = itIs || elem.type === 'VariableDeclaration';
    itIs = itIs || elem.type === 'Punctuator';
    itIs = itIs || elem.type === 'ExpressionStatement'

    return itIs;
  }

while(IsRelevant(elem)) {
  elem = elem.previousSibling;
}
```

Does it reads better?

zxqfox: Better? ;-)

zxqfox:
Thanks!

PR #2179, opened 2016-03-17
Fix: fixup skipped requirePaddingNewlinesBeforeKeywords test

hzoo:
I thought this was a cst issue since I got an internal error?

But it was just removed in https://github.com/jscs-dev/node-jscs/commit/301232104111027106588f230c371672fb0ace43?

```js
rules/require-padding-newlines-before-keywords true value should not report when returning a function:
     Expected not to have errors, but "internalError: Error running rule requirePaddingNewlinesBeforeKeywords: This is an issue with JSCS and not your codebase.
Error: Expected end of node list but "FunctionExpression" found
    at ElementAssert.assertEnd (/Users/hzhu/dev/node-jscs/node_modules/cst/lib/elements/ElementAssert.js:208:23)
    at ReturnStatement._acceptChildren (/Users/hzhu/dev/node-jscs/node_modules/cst/lib/elements/types/ReturnStatement.js:49:22)
    at ReturnStatement._setChildren (/Users/hzhu/dev/node-jscs/node_modules/cst/lib/elements/Element.js:415:18)
```

@markelog

### DIFF
--- a/test/specs/rules/require-padding-newlines-before-keywords.js
+++ b/test/specs/rules/require-padding-newlines-before-keywords.js
@@ -103,51 +103,50 @@ describe('rules/require-padding-newlines-before-keywords', function() {
 
         it('should not report on first expression in a block', function() {
             expect(checker.checkString(
-                    'function x() { return; }'
-                )).to.have.no.errors();
+                'function x() { return; }'
+            )).to.have.no.errors();
         });
 
         it('should not report in object definitions', function() {
             expect(checker.checkString(
-                    'var obj = {\n' +
-                    '  foo: function x() { return; }\n' +
-                    '};'
-                )).to.have.no.errors();
+                'var obj = {\n' +
+                '  foo: function x() { return; }\n' +
+                '};'
+            )).to.have.no.errors();
         });
 
         it('should not report when parameter inside a function', function() {
             expect(checker.checkString(
-                    'watchlist.on( "watch", function () {} );'
-                )).to.have.no.errors();
+                'watchlist.on( "watch", function () {} );'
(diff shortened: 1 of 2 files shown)


test/specs/rules/require-padding-newlines-before-keywords.js
```diff
@@ -103,51 +103,50 @@ describe('rules/require-padding-newlines-before-keywords', function() {
...
             expect(checker.checkString(
-                    'function x() { var a; return; }'
-                )).to.have.one.validation.error.from('requirePaddingNewlinesBeforeKeywords');
+                'function x() { var a; return; }'
+            )).to.have.one.validation.error.from('requirePaddingNewlinesBeforeKeywords');
         });
 
-        // TODO: fix this in CST
-        it.skip('should not report when returning a function', function() {
+        it('should not report when returning a function', function() {
```
markelog: `only`

hzoo: yeah I saw haha

markelog:
I think we should fix this on the cst side, can you port your changes over there?

hzoo:
I didn't figure out why it was failing on the cst side so I looked at the original file

markelog:
I'm thinking if that happened in this case it [should](https://en.wikipedia.org/wiki/Murphy%27s_law) happen elsewhere too

markelog:
So what do you think? Still wanna do it here, regardless?

hzoo:
Yeah although we can leave an issue open in cst

hzoo:
`return \n function() {}` is invalid since it results in `return undefined` due to the newline, so merging

PR #2181, opened 2016-03-17
Skipped tests2

hzoo:


markelog:
Judging by the travis log, it seems we need to increase the test timeout?

hzoo:
Yeah only for those tests I guess

hzoo:
Oh, I see - that sucks - so that timeout was for my computer

markelog:
> Means that the tests were taking < 3 sec and now > 15 which is pretty bad and something we certainly need to investigate?

Definitely! /cc @mdevils

PR #2184, opened 2016-03-22
Fixup tests

hzoo:


lib/rules/validate-quote-marks.js
```diff
@@ -116,7 +116,9 @@ module.exports.prototype = {
         file.iterateTokensByType('String', function(token) {
             if (
                 ignoreJSX &&
-                token.parentElement.type === 'StringLiteral' &&
+                // TODO: fix in CST?
+                (token.parentElement.type === 'Literal' ||
```
hzoo: Need to figure out how token.parentElement is being output as `Literal`

hzoo:
Interesting it timeouted more than 15sec on node4/5

lib/rules/validate-quote-marks.js
hzoo: Ok I am confused why I thought this was needed before but it works now

Issue #2076, opened 2016-01-19
Unused var, const or let


felipekm:
There's some configuration that handle it?

hzoo:
No, we don't have a rule for that yet - there is http://jscs.info/rule/disallowUnusedParams.

markelog:
It should be in 3.0 though!

markelog:
With autofixing too

felipekm:
Great, thanks!

schempy:
If nobody is working on this I'd like to implement.

Thanks!

PR #2199, opened 2016-04-07
New rule: disallowUnusedVariables

schempy:
disallowUnusedVariables: warn/auto-fix unused variables defined with var, let or const.

Fixes: #2076

### DIFF
--- a/lib/rules/disallow-unused-variables.js
+++ b/lib/rules/disallow-unused-variables.js
@@ -1,0 +1,142 @@
+/**
+ * Disallows unused variables defined with var, let or const.
+ *
+ * Types: `Boolean`
+ *
+ * Values: `true`
+ *
+ * #### Example
+ *
+ * ```js
+ * "disallowUnusedVariables": true
+ * ```
+ *
+ * ##### Valid
+ *
+ * ```js
+ * var x=1;
+ *
+ * function getX() {
+ *     return x;
+ * }
+ *
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var x =1;
+ *
+ * function getX() {
+ *     var x=1;
+ *     return x;
+ * }
+ *
+ * ```
+ */
+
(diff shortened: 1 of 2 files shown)


schempy:
The Travis CI build shows all tests passing. I noticed the following message from the node v0.10.44 build:
```
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

Not sure if that is from coveralls. I did a rebase from the current master branch prior to the pull request.

markelog:
Wow, if everything works, that would be an impressive one!

/cc @mdevils 

> Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}

You probably force-push at some point, which what happens, i.e. totally okay.

lib/rules/disallow-unused-variables.js
```diff
@@ -0,0 +1,142 @@
...
+ * ```
+ *
+ * ##### Invalid
+ *
+ * ```js
+ * var x =1;
+ *
+ * function getX() {
+ *     var x=1;
+ *     return x;
```
markelog: It seems in this case, variable is used?

schempy:
Thanks for catching the invalid example. I updated.

markelog:
@schempy you are the man! Thanks

jdalton:
This is having issues:
```bash
disallowUnusedVariables: Variable `object` is not used at lodash.js :
 14622 |     */
 14623 |    var method = rest(function(path, args) {
 14624 |      return function(object) {
---------------------------------^
 14625 |        return baseInvoke(object, path, args);
 14626 |      };
```

You can see the var is used

markelog:
That was fast! @schempy could you check it out?

mdevils:
Thanks, will fix it soon.

mdevils:
Fixed in `3.0.1`. Please check it out.

jdalton:
It works! Updated dep in Lodash :D

PR #2208, opened 2016-04-13
Misc: Make an exception for esnext/verbose since they are removed

hzoo:
```js
{
"esprima": true,
    "esprimaOptions": true,
    "verbose": true,
    "esnext": true,
}
```

screen shot 2016-04-16 at 12 05 40 pm

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -378,6 +378,30 @@ Configuration.prototype._getOptionsFromConfig = function(config) {
     }, {});
 };
 
+Configuration.prototype._errorOnRemovedOptions = function(config) {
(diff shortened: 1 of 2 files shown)


markelog:
Tests?

lib/string-checker.js
```diff
@@ -158,8 +158,14 @@ StringChecker.prototype = {
             }
         }, this);
 
+        var removed = ['esnext', 'verbose'];
```
hzoo: also esprimaOptions, esprima

hzoo:
Ok that was the wrong place to put that lol

zxqfox:
LGTM

Issue #2221, opened 2016-04-15
CST process out of memory


TheSavior:
On one of my travis builds from the 3.0.2 greenkeeper bump, I have this error:

https://travis-ci.org/TheSavior/find-parent/jobs/123328178
```
> eslint . && jscs .
<--- Last few GCs --->
   71265 ms: Mark-sweep 1394.1 (1458.1) -> 1395.6 (1458.1) MB, 1424.4 / 0 ms [allocation failure] [GC in old space requested].
   72689 ms: Mark-sweep 1395.6 (1458.1) -> 1397.9 (1458.1) MB, 1424.0 / 0 ms [allocation failure] [GC in old space requested].
   74122 ms: Mark-sweep 1397.9 (1458.1) -> 1389.1 (1458.1) MB, 1433.6 / 0 ms [last resort gc].
   75554 ms: Mark-sweep 1389.1 (1458.1) -> 1394.1 (1458.1) MB, 1431.7 / 0 ms [last resort gc].
<--- JS stacktrace --->
==== JS stack trace =========================================
Security context: 0x122c644e3ac1 <JS Object>
    2: _acceptChildren [/home/travis/build/TheSavior/find-parent/node_modules/cst/lib/elements/types/IfStatement.js:61] [pc=0x245f03371086] (this=0x3bed2a019961 <an IfStatement with map 0x1f71cd5cefb9>,children=0x3bed2a019a19 <an ElementAssert with map 0x294acfca3279>)
    3: _setChildren [/home/travis/build/TheSavior/find-parent/node_modules/cst/lib/elements/Element.js:721] [pc=0x245f03377094]...
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
Aborted (core dumped)
```

wulfsolter:
Failing on local machine as well with same symptoms.

wulfsolter:
Manually added `node_modules` to my JSCS ignores in `.jscsrc`

    {
        ...
        excludeFiles: [
            ...
            "node_modules/**",
            ...
        ],
        ...
    }

hzoo:
Hm interesting since we had already excluded node_modules before.

Oh I see https://github.com/jscs-dev/node-jscs/commit/908bd20654723eac98949f994710c9ea902fe671 which changed `node_modules/**` to `*/node_modules/**`. Maybe we need both.. globs are confusing @kirbysayshi

hzoo:
[...]

PR #2222, opened 2016-04-16
Revert "Configuration: exclude nested node_modules by default"

hzoo:
This reverts commit 908bd20654723eac98949f994710c9ea902fe671.

Ref https://github.com/jscs-dev/node-jscs/issues/2221

Either we revert back to the old behavior and only exclude the top node_modules by default (which is what eslint is doing) or we also exclude nested ones by default.

### DIFF
--- a/lib/config/configuration.js
+++ b/lib/config/configuration.js
@@ -117,7 +117,7 @@ function Configuration() {
      * @protected
      * @type {Array}
      */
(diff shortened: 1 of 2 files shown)


hzoo:
@kirbysayshi Ok I think we are going to revert to previous behavior in 2.x and users will have to specify nested node_modules themselves since that isn't as common of a use case.

PR #2239, opened 2016-05-06
validateIndentation: Add 'consistent' mode.

kylepaulsen:
Consistent mode tries to guess the intended indentation in the file and
sets the indentation settings accordingly.

Hello, I finally decided to take the time to do this. This just adds a 'consistent' option to validateIndentation. So now validateIndentation can be a number, "\t" or "consistent". This is helpful for people who work on a wide variety of projects with their own indentation rules where the projects don't necessarily have a jscs file. It is also helpful for people who have a global jscsrc file so they can enjoy consistent indentation along with all of their other favorite style rules. This is especially helpful for editors that use that global jscsrc file on any source file (a new unsaved file or otherwise).

It's important to note that when using consistent mode, jscs has to make a guess at what the intended indentation should be. So there might be cases where it will guess wrong. So the trade off is more flexibility when switching between projects with different indentation rules and incorrect indentation warnings (probably a rare occurrence).

Here's how it works:
After a lot of thought, I decided that the most reliable place to sample indentation is the first Whitespace node directly inside a BlockStatement node from the AST. Additionally, the Whitespace node must start with a newline character (\n or \r\n). If it can't find a place to sample the indentation, it will try once more, but this time the Whitespace node doesn't have to be in a BlockStatement. However, if it has to search a second time, it will only take note of the type of indentation being used (so it will only warn if the code mixes spaces and tabs).

I think this would be a good option for jscs to have. It would certainly save me a lot of frustration. Let me know if I need to change anything about the PR. Thanks!

### DIFF
--- a/lib/rules/validate-indentation.js
+++ b/lib/rules/validate-indentation.js
@@ -195,21 +195,30 @@ function isFunctionExpression(node) {
 /**
  * Returns the first non-comment indentation in the first block or undefined.
  *
- * @param {Node} node
- * @param {String} parentType
+ * @param {String[]} programLines - The js file's lines
+ * @param {Node} node - esprima node
+ * @param {String} parentType - esprima parent node type
  * @param {Boolean} findAnywhere - Allow sampling whitespace outside of block nodes.
  * @returns {String|undefined}
  */
-function guessIntendedIndentation(node, parentType, findAnywhere) {
+function guessIntendedIndentation(programLines, node, parentType, findAnywhere) {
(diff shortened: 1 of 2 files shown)


kylepaulsen:
I think Travis is complaining about things unrelated to my PR. Can someone trigger a rebuild?

markelog:
Really sorry, we do not accept new features at this time :/.

See https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.jsnhpxsjz

Issue #2273, opened 2016-06-15
not possible to ignore '.js' files with fileExtensions


graingert:
https://github.com/graingert/jscs-error
With only .es2015.js files included:
```
{
    "preset": "airbnb",
    "fileExtensions": ['.es2015.js']
}
```
JSCS still complains if .js files have errors:

```
$ npm install
$ ./node_modules/.bin/jscs src
Unexpected character '#' (3:4) at src/old.js :
     1 |sdlkfjhskldjads
     2 |asdf
     3 |asdf#
------------^
     4 |asdf
     5 |asdf


1 code style error found.
```

markelog:
Well in that case, i.e. `*.es2015.js`, file extension would be `.js`, like 
`require("path").extname('test.a.b')` would be `.b`. For example if you filename would be `v1.0.js`, extension should still be `.js`.

You can mitigate this by not linting the full folder, but using glob like `jscs **/*.es2015.js`

graingert:
This is supported by eslint

graingert:
Also jscs doesn't support globs it has to be expanded by the shell

markelog:
You can use globs in `.eslintignore` yes, if that is what you mean by "supported", but you can't use it with `--ext` flag.

You can use glob patterns with jscs from the command line

graingert:
still fails when using ".es" https://github.com/graingert/jscs-error/tree/a7a0778de32084d9b8b1dd75bfb4ca0f92d2356a

graingert:
@markelog also you can use it with the ext flag: "./node_modules/.bin/eslint --ext .es2015.js src", works just fine.

graingert:
@markelog can you reopen this please?
Not sure what you mean by:

> Well in that case, i.e. `*.es2015.js`, file extension would be .js, like `require("path").extname('test.a.b')` would be `.b`. For example if you filename would be `v1.0.js`, extension should still be `.js`.
>
> You can mitigate this by not linting the full folder, but using glob like jscs **/*.es2015.js

Because in the case of `*.es2015.js`, the file extension would be `.es2015.js` and `.js`.
```require('lodash').endsWith('test.a.b', '.a.b')``` would be true. Still when you had `v1.0.js` the extension would be `.js`. But you could also only test: `0.js` files.

PR #2276, opened 2016-06-20
file extension fixes. Fixes #2275 #2273

### LINKED ISSUE #2275: not possible to ignore '.js' files with fileExtensions #2273
https://github.com/graingert/jscs-error
With only .es files included:
```
{
    "preset": "airbnb",
    "fileExtensions": ['.es']
}
```
JSCS still complains if .js files have errors:

```
$ npm install
$ ./node_modules/.bin/jscs src
Unexpected character '#' (3:4) at src/old.js :
     1 |sdlkfjhskldjads
     2 |asdf
     3 |asdf#
------------^
     4 |asdf
     5 |asdf


1 code style error found.
```

graingert:
Fixes #2275 #2273


markelog:
Decided to fix this a bit differently, would appreciate a second look though - https://github.com/jscs-dev/node-jscs/commit/0d027a7b4a95c7c349f57ffd37f10e497d66a5d4

graingert:
@markelog this doesn't fix #2273

markelog:
It's not supposed to

graingert:
Can you reopen #2273 so I can recreate the PR to fix it?

markelog:
I don't know man, everywhere i looked, like https://en.wikipedia.org/wiki/File_format#Filename_extension,  https://en.wikipedia.org/wiki/Filename_extension https://docs.python.org/2/library/os.path.html#os.path.splitext,
`path.extname` of node.js, 
http://ruby-doc.org/core-2.2.0/File.html#method-c-extname,
https://golang.org/pkg/path/filepath/#Ext
and others... give me the same result

markelog:
You can probably workaround it with something like this though - 
```js
"excludeFiles": ["**/!(*.es2015.js)"]
```

graingert:
@markelog nope, that seems to exlude all files.

markelog:
It should ignore all files which ends with `.es2015.js`

graingert:
@markelog I know. But it seems to exclude all files

markelog:
Oh shoot, typo there, it *shouldn't* ignore files which ends with  `.es2015.js`

graingert:
Sorry I doubled the mistake and thought that's what you said originally. It still ignores all files

markelog:
Weird, works for me

markelog:
Maybe you could illustrate it with test repo?

graingert:
Yeah will do

PR #2284, opened 2016-06-28
Add facebook preset to jscs

AlexJuarez:
Here is the facebook jscs linting config.

### DIFF
--- a/presets/facebook.json
+++ b/presets/facebook.json
@@ -1,0 +1,101 @@
+{
+  "requireCurlyBraces": [
+    "if",
+    "else",
+    "for",
+    "while",
+    "do",
+    "try",
+    "catch"
+  ],
+  "requireOperatorBeforeLineBreak": true,
+  "requireCamelCaseOrUpperCaseIdentifiers": {
+    "ignoreProperties": true
+  },
+  "maximumLineLength": {
+    "value": 80,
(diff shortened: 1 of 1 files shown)


markelog:
Pretty sure facebook doesn't have public javascript code style, regardless, this is not implemented right and we are not adding new features at this time.

See https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2, http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint

bgw:
I'm not encouraging this commit, but for completeness,

> facebook doesn't have public javascript code style

isn't entirely true. [We publish an eslint config](https://github.com/facebook/fbjs/blob/master/scripts/eslint/.eslintrc.js), but we don't necessarily advocate that others use it. That configuration is used by many open source FB projects, like react.

markelog:
I meant formal document for it, like 
https://google.github.io/styleguide/javascriptguide.xml
https://github.com/airbnb/javascript

since linters are not yet at the point when everything could be defined through config.

Also i don't see it used, like: 
https://github.com/reactjs/react-router/blob/master/.eslintrc
https://github.com/facebook/react/blob/master/.eslintrc.js (redefines it quite a lot)
https://github.com/reactjs/redux/blob/master/.eslintrc

are very different

PR #2292, opened 2016-09-02
Disallowing multiple spaces.

tomayac:
Disallow multiple indentation characters (tabs or spaces) between identifiers, keywords, and any other token.

Would be great if you could still merge this trivial PR, my team are using ```jscs``` for certain projects.


markelog:
No, sorry, you still can do this with custom config or overwriting this rule on the project level

PR #2293, opened 2016-11-16
require `cst` v0.4.6 or higher because v0.4.5 was published to npm wihouth a `lib` dir

rubensayshi:
big problem, easy fix, hope this can get sneaked past the EOL ;-)


markelog:
Just reinstall jscs

